PR #2, opened 2023-10-24
Add null/trace pool and provider implementations

ldorau:

PR #3, opened 2023-10-24
Add memoryPool and memoryProvider tests

ldorau:
Requires:
- [x] #2

### DIFF
--- a/test/memoryPoolAPI.cpp
+++ b/test/memoryPoolAPI.cpp
@@ -1,0 +1,306 @@
+// Copyright (C) 2023 Intel Corporation
+// Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// This file contains tests for UMF pool API
+
+#include "pool.hpp"
+#include "provider.hpp"
+#include "test_helpers.h"
+
+#include "memoryPool.hpp"
+#include "umf/memory_provider.h"
+
+#include <array>
+#include <string>
+#include <thread>
+#include <unordered_map>
+
+using umf_test::test;
+
+TEST_F(test, memoryPoolTrace) {
+    static std::unordered_map<std::string, size_t> poolCalls;
+    static std::unordered_map<std::string, size_t> providerCalls;
+    auto tracePool = [](const char *name) { poolCalls[name]++; };
+    auto traceProvider = [](const char *name) { providerCalls[name]++; };
+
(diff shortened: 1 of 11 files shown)


test/memoryProviderAPI.cpp
```diff
@@ -0,0 +1,91 @@
...
+#include <string>
+#include <unordered_map>
+
+using umf_test::test;
+
+TEST_F(test, memoryProviderTrace) {
+    static std::unordered_map<std::string, size_t> calls;
+    auto trace = [](const char *name) { calls[name]++; };
+
+    auto nullProvider = umf_test::wrapProviderUnique(nullProviderCreate());
```
bratpiorka, line 19: I'm not convinced about using wrapProviderUnique function instead of using regular UMF API (umfProviderCreate) here. I think it complicates the code and we also don't test here umfProviderCreate functions straightly.

test/memoryPoolAPI.cpp
```diff
@@ -0,0 +1,307 @@
...
+
+using umf_test::test;
+
+TEST_F(test, memoryPoolTrace) {
+    static std::unordered_map<std::string, size_t> poolCalls;
+    static std::unordered_map<std::string, size_t> providerCalls;
+    auto tracePool = [](const char *name) { poolCalls[name]++; };
+    auto traceProvider = [](const char *name) { providerCalls[name]++; };
+
+    auto nullProvider = umf_test::wrapProviderUnique(nullProviderCreate());
```
bratpiorka, line 27: I'm not convinced about using wrapProviderUnique function instead of using regular UMF API (umfProviderCreate, umfPoolCreate) here. I think it complicates the code and we also don't test here umfProviderCreate and umfPoolCreate functions straightly.

test/memoryProviderAPI.cpp
igchor: wrapProvider unique does not call any create function underneath, it only wraps the handle into unique_ptr - you still need to call umfProviderCreate (or nullProviderCreate as in this case) so we don't hide anything

test/memoryPoolAPI.cpp
igchor: wrapProvider unique does not call any create function underneath, it only wraps the handle into unique_ptr - you still need to call umfProviderCreate (or nullProviderCreate as in this case) so we don't hide anything

ldorau:
Rebased and updated.
@bratpiorka @igchor please review

bratpiorka (approved):

igchor (approved):

PR #6, opened 2023-10-31
Refactor umfPoolCreate to accept only one provider

vinser52:
Pool is a combination of memory provider and pool manager. The umfPoolCreate function s is refactored to accept only a single memory provider as we agreed.

### DIFF
--- a/include/umf/memory_pool.h
+++ b/include/umf/memory_pool.h
@@ -24,16 +24,14 @@ struct umf_memory_pool_ops_t;
 ///
 /// \brief Creates new memory pool.
 /// \param ops instance of umf_memory_pool_ops_t
-/// \param providers array of memory providers that will be used for coarse-grain allocations.
-///        Should contain at least one memory provider.
-/// \param numProvider number of elements in the providers array
+/// \param provider memory provider that will be used for coarse-grain allocations.
 /// \param params pointer to pool-specific parameters
 /// \param hPool [out] handle to the newly created memory pool
 /// \return UMF_RESULT_SUCCESS on success or appropriate error code on failure.
 ///
 enum umf_result_t umfPoolCreate(const struct umf_memory_pool_ops_t *ops,
-                                umf_memory_provider_handle_t *providers,
-                                size_t numProviders, void *params,
+                                umf_memory_provider_handle_t provider,
+                                void *params,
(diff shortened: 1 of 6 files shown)


vinser52:
Should we update the UR repo as well?

igchor:
> Should we update the UR repo as well?

For UR, we actually use some C++ wrappers so it might be enough to just update those - I'll take a look at this once DisjointPool is merged.

include/umf/memory_pool.h
```diff
@@ -137,15 +135,12 @@ umf_memory_pool_handle_t umfPoolByPtr(const void *ptr);
...
-/// \param numProviders [in] number of memory providers to return
-/// \param numProvidersRet pointer to the actual number of memory providers
+/// \param hProvider [out] memory providers handle.
 /// \return UMF_RESULT_SUCCESS on success or appropriate error code on failure.
+///         UMF_RESULT_ERROR_INVALID_ARGUMENT if hProvider is NULL
 enum umf_result_t
-umfPoolGetMemoryProviders(umf_memory_pool_handle_t hPool, size_t numProviders,
-                          umf_memory_provider_handle_t *hProviders,
-                          size_t *numProvidersRet);
+umfPoolGetMemoryProviders(umf_memory_pool_handle_t hPool,
```
igchor, line 142: Should we change the name to `umfPoolGetMemoryProviders`? (without the `s`)

vinser52:
> > Should we update the UR repo as well?
> 
> For UR, we actually use some C++ wrappers so it might be enough to just update those - I'll take a look at this once DisjointPool is merged.

If disjointPool merged before this PR, I will update this PR to fix DisjointPool if needed.

igchor:
Btw, @ldorau from which UR commit have you taken the umf code? E.g. I don't see the memory tracker (as in this version: https://github.com/oneapi-src/unified-runtime/tree/88c7f8f4e3db8a22880140ea14098c6572c4e2de/source/common/unified_malloc_framework/src) - will this be added later?

include/umf/memory_pool.h
vinser52: Done

igchor (approved):

bratpiorka (approved):

PR #11, opened 2023-11-02
Add disjointPool test

ldorau:
Requires:
- [x] https://github.com/oneapi-src/unified-memory-framework/pull/21

### DIFF
--- a/test/disjoint_pool.cpp
+++ b/test/disjoint_pool.cpp
@@ -1,0 +1,82 @@
+// Copyright (C) 2023 Intel Corporation
+// Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "memoryPool.hpp"
+#include "pool.hpp"
+#include "provider_null.h"
+#include "provider_trace.h"
+#include "provider.hpp"
+#include "pool_disjoint_impl.hpp"
+
+static usm::DisjointPool::Config poolConfig() {
+    usm::DisjointPool::Config config{};
+    config.SlabMinSize = 4096;
+    config.MaxPoolableSize = 4096;
+    config.Capacity = 4;
+    config.MinBucketSize = 64;
+    return config;
+}
+
+static auto makePool() {
+    auto [ret, provider] =
+        umf::memoryProviderMakeUnique<umf_test::provider_malloc>();
+    EXPECT_EQ(ret, UMF_RESULT_SUCCESS);
+    auto [retp, pool] = umf::poolMakeUnique<usm::DisjointPool>(
+        {std::move(provider)}, poolConfig());
+    EXPECT_EQ(retp, UMF_RESULT_SUCCESS);
+    return std::move(pool);
+}
+
+using umf_test::test;
+
+TEST_F(test, freeErrorPropagation) {
+    static enum umf_result_t freeReturn = UMF_RESULT_SUCCESS;
+    struct memory_provider : public umf_test::provider_base {
+        enum umf_result_t alloc(size_t size, size_t, void **ptr) noexcept {
+            *ptr = malloc(size);
+            return UMF_RESULT_SUCCESS;
+        }
+        enum umf_result_t free(void *ptr,
+                               [[maybe_unused]] size_t size) noexcept {
+            ::free(ptr);
+            return freeReturn;
+        }
+    };
+
+    auto [ret, providerUnique] =
+        umf::memoryProviderMakeUnique<memory_provider>();
+    ASSERT_EQ(ret, UMF_RESULT_SUCCESS);
+
+    auto config = poolConfig();
+    config.MaxPoolableSize =
+        0; // force all allocations to go to memory provider
+
+    auto [retp, pool] = umf::poolMakeUnique<usm::DisjointPool>(
+        {std::move(providerUnique)}, config);
+    EXPECT_EQ(retp, UMF_RESULT_SUCCESS);
+
+    static constexpr size_t size = 1024;
+    void *ptr = umfPoolMalloc(pool.get(), size);
+
+    freeReturn = UMF_RESULT_ERROR_MEMORY_PROVIDER_SPECIFIC;
+    auto freeRet = umfPoolFree(pool.get(), ptr);
+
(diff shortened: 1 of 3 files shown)


test/disjoint_pool.cpp
```diff
@@ -0,0 +1,82 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "memoryPool.hpp"
+#include "pool.hpp"
+#include "provider/provider_null.h"
+#include "provider/provider_trace.h"
+#include "provider.hpp"
+#include "pool_disjoint.hpp"
+
+static usm::DisjointPool::Config poolConfig() {
```
igchor, line 12: It might be better to use the C API you added for configuration. Then, you would use regular C API `umfPoolCreate` instead of the `poolMakeUnique` below.

ldorau: Why is it better to use `umfPoolCreate` than `poolMakeUnique` here?

igchor: I think this test should test the public API and `poolMakeUnique` will not be public (at least there is no plan for that right now). The C++ helpers (`poolMakeUnique`, `providerMakeUnique`, etc.) were created mostly to simplify writing tests where we need to have a custom (mock) pool/provider and to simplify integration with Unified Runtime adapters.

ldorau:
This PR requires: https://github.com/oneapi-src/unified-memory-framework/pull/21

test/disjoint_pool.cpp
ldorau: @igchor Could you take a look at the top-last commit?

test/disjoint_pool.cpp
```diff
@@ -0,0 +1,107 @@
...
+        umfPoolDestroy(pool);
+        umfMemoryProviderDestroy(provider_handle);
+    };
+
+    enum umf_result_t retp;
+    umf_memory_pool_handle_t pool = NULL;
+    retp = umfPoolCreate(&UMF_DISJOINT_POOL_OPS, provider_handle, &params, &pool);
+    EXPECT_EQ(retp, UMF_RESULT_SUCCESS);
+
+    return std::move(umf::pool_unique_handle_t(pool, std::move(poolDestructor)));
```
igchor, line 50: nit: the first std::move is not needed

test/disjoint_pool.cpp
igchor: Looks good. Can we remove the `poolConfig` function now?

test/disjoint_pool.cpp
```diff
@@ -0,0 +1,107 @@
...
+static usm::DisjointPool::Config poolConfig() {
+    usm::DisjointPool::Config config{};
+    config.SlabMinSize = 4096;
+    config.MaxPoolableSize = 4096;
+    config.Capacity = 4;
+    config.MinBucketSize = 64;
+    return config;
+}
+
+struct umf_disjoint_pool_params params = {
```
igchor, line 22: I would suggest making a function that returns `umf_disjoint_pool_params` rather than a global variable. `freeErrorPropagation` test modifies some members of the `params` and if we add more tests in the future, the behavior might be unexpected.

test/disjoint_pool.cpp
ldorau: Removed

test/disjoint_pool.cpp
ldorau: @igchor poolConfig() is still used in `umf_test::makePoolWithOOMProvider<usm::DisjointPool>()`
https://github.com/ldorau/unified-memory-framework/blob/5f80bced56adf7a8d8f41a16bd4ad1a092dcc260/test/disjoint_pool.cpp#L90

ldorau:
> Looks good. Can we remove the poolConfig function now?

@igchor poolConfig() is still used in `umf_test::makePoolWithOOMProvider<usm::DisjointPool>()`
https://github.com/ldorau/unified-memory-framework/blob/5f80bced56adf7a8d8f41a16bd4ad1a092dcc260/test/disjoint_pool.cpp#L90

test/disjoint_pool.cpp
ldorau: Done

test/disjoint_pool.cpp
igchor: Would it be possible to rewrite `makePoolWithOOMProvider` so that we would use the C API as well? I would like to get rid of the `DisjointPoolConfig` entirely.

I see that `makePoolWithOOMProvider` is also used here: https://github.com/ldorau/unified-memory-framework/blob/5f80bced56adf7a8d8f41a16bd4ad1a092dcc260/test/memoryPoolAPI.cpp#L143 but I think we can remove it - it was only to 'test the test'.

igchor: Actually, I'm working on one PR where I want to remove the DisjointPoolConfig entirely so I can rewrite the `makePoolWithOOMProvider` there. You can leave it for now.

ldorau: OK, I leave it for now.

test/disjoint_pool.cpp
```diff
@@ -0,0 +1,112 @@
...
+    retp = umfPoolCreate(&UMF_DISJOINT_POOL_OPS, provider_handle, &params, &pool);
+    EXPECT_EQ(retp, UMF_RESULT_SUCCESS);
+
+    static constexpr size_t size = 1024;
+    void *ptr = umfPoolMalloc(pool, size);
+
+    freeReturn = UMF_RESULT_ERROR_MEMORY_PROVIDER_SPECIFIC;
+    auto freeRet = umfPoolFree(pool, ptr);
+
+    EXPECT_EQ(freeRet, freeReturn);
```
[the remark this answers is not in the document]
igchor: It seems like a bug in pool_disjoint.cpp: https://github.com/oneapi-src/unified-memory-framework/blob/main/src/pool/disjoint/pool_disjoint.cpp#L78

We always return UMF_RESULT_SUCCESS and we should just return the result of `pool_data->disjoint_pool->free(ptr)`.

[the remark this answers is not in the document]
ldorau: That's it! :-D Thanks a lot, @igchor!

igchor (approved):

PR #12, opened 2023-11-02
Rename disjoint pool files

ldorau:
Rename files in src/pool/disjoint/ :
pool_disjoint_heap_manager.cpp -> pool_disjoint.cpp 
pool_disjoint.cpp -> pool_disjoint_impl.cpp
pool_disjoint.hpp -> pool_disjoint_impl.hpp


bratpiorka (approved):

src/pool/disjoint/pool_disjoint.cpp
```diff
@@ -2,998 +2,115 @@
...
-    size_t getChunkSize() const;
-    size_t getNumChunks() const { return Chunks.size(); }
-
-    bool hasAvail();
-
-    Bucket &getBucket();
-    const Bucket &getBucket() const;
-
-    void freeChunk(void *Ptr);
+struct disjoint_memory_pool {
```
vinser52, line 16: I believe we still need a namespace. I agree that we should not use "usm" namespace name. Let's use "umf" instead.

vinser52 (changes requested):

src/pool/disjoint/pool_disjoint.cpp
ldorau: Can I do this in a separate PR? I do not want make changes in the PR renaming files.

PR #18, opened 2023-11-03
Add memory tracker and critnib

ldorau:


bratpiorka:
Please make the tracker provider similar to other providers. Also, please think about proper directory structure.

src/memory_pool_tracking.c
```diff
@@ -0,0 +1,93 @@
...
+    ret = ops->initialize(pool->provider, params, &pool->pool_priv);
+    if (ret != UMF_RESULT_SUCCESS) {
+        goto err_pool_init;
+    }
+
+    *hPool = pool;
+    return UMF_RESULT_SUCCESS;
+
+err_pool_init:
+err_providers_init:
```
igchor, line 57: Shouldn't this label be after `umfMemoryProviderDestroy`?

src/memory_pool_tracking.c
```diff
@@ -0,0 +1,93 @@
...
+
+    enum umf_result_t ret = UMF_RESULT_SUCCESS;
+    umf_memory_pool_handle_t pool = malloc(sizeof(struct umf_memory_pool_t));
+    if (!pool) {
+        return UMF_RESULT_ERROR_OUT_OF_HOST_MEMORY;
+    }
+
+    assert(ops->version == UMF_VERSION_CURRENT);
+
+    pool->provider = calloc(1, sizeof(umf_memory_provider_handle_t));
```
igchor, line 35: Hm, I don't think we need this calloc anymore, this was needed when we had an array of providers, but now just calling `umfTrackingMemoryProviderCreate` on `pool->provider` should be enough.

igchor:
> Please make the tracker provider similar to other providers.

What do you mean by that?

src/memory_pool_tracking.c
ldorau: Done

src/memory_pool_tracking.c
ldorau: Right, done.

bratpiorka:
> > Please make the tracker provider similar to other providers.
> 
> What do you mean by that?

@igchorI was discussing this with Luke online. I was thinking of keeping the tracking provider implementation in separate .c and .h files that use the same style as other providers. The latest PR version does that.

CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ include(helpers)
 # Build Options
 option(UMF_DEVELOPER_MODE "enable developer checks, treats warnings as errors" OFF)
 option(UMF_BUILD_SHARED_LIBRARY "Build UMF as shared library" OFF)
+option(UMF_ENABLE_POOL_TRACKING "Build UMF with pool tracking" OFF)
```
vinser52, line 19: We should also move the APIs (based on tracking capabilities) mentioned below to `memory_pool.h` and add corresponding tests to make sure tracking works as expected. Also, I think we need to turn on this functionality by default.
```
///
/// \brief Frees the memory space pointed by ptr if it belongs to UMF pool, does nothing otherwise.
/// \param ptr pointer to the allocated memory
///         Whether any status other than UMF_RESULT_SUCCESS can be returned
///         depends on the memory provider used by the pool.
///
enum umf_result_t umfFree(void *ptr);

///
/// \brief Retrieve memory pool associated with a given ptr. Only memory allocated
///        with the usage of a memory provider is being tracked.
/// \param ptr pointer to memory belonging to a memory pool
/// \return Handle to a memory pool that contains ptr or NULL if pointer does not belong to any UMF pool.
umf_memory_pool_handle_t umfPoolByPtr(const void *ptr);
```

ldorau: The API mentioned above already are in `memory_pool.h`.
Some tests already are in `test/memoryPool.hpp`.

ldorau: Turned on by default.

vinser52 (approved):

bratpiorka (approved):

igchor (approved):

PR #22, opened 2023-11-07
Remove duplication of DisjointPoolConfig members

KFilipek:
This PR removes duplications of members. 
Instead of, the DisjointPoolConfig will extend pool params.

### DIFF
--- a/src/pool/disjoint/pool_disjoint_impl.hpp
+++ b/src/pool/disjoint/pool_disjoint_impl.hpp
@@ -10,13 +10,14 @@
 #include <string>
 
 #include "umf.h"
+#include "pool/pool_disjoint.h"
 
 namespace usm {
 
 inline constexpr size_t MIN_BUCKET_DEFAULT_SIZE = 8;
 
 // Configuration for specific USM allocator instance
-class DisjointPoolConfig {
+class DisjointPoolConfig : public umf_disjoint_pool_params {
   public:
     DisjointPoolConfig();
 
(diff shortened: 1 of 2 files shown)


src/pool/disjoint/pool_disjoint_impl.hpp
```diff
@@ -32,26 +33,6 @@ class DisjointPoolConfig {
...
-
-    // Allocations up to this limit will be subject to chunking/pooling
-    size_t MaxPoolableSize = 0;
-
-    // When pooling, each bucket will hold a max of 4 unfreed slabs
-    size_t Capacity = 0;
-
-    // Holds the minimum bucket size valid for allocation of a memory type.
-    // This value must be a power of 2.
-    size_t MinBucketSize = MIN_BUCKET_DEFAULT_SIZE;
```
igchor, line 47: This is not related to this PR specifically but I just realized that we don't have any way to set the default values for params in the C API. @ldorau perhaps we should add some function to the C API, like 'umf_pool_disjoint_default_params()`?

Btw, I see that we do specify default values in https://github.com/oneapi-src/unified-memory-framework/blob/main/include/pool/pool_disjoint.h but I think this is a bug. It will not compile in C. It would be good to have at least one test for disjoint pool in C rather than in C++, so we could catch this

igchor (approved):
LGTM, but I expect the `DisjointPoolConfig` to be replaced by `umf_disjoint_pool_params` in future anyway. We just need to figure out how to pass shared limits.

src/pool/disjoint/pool_disjoint_impl.hpp
igchor: Ref: https://github.com/oneapi-src/unified-memory-framework/pull/28

PR #27, opened 2023-11-08
Add UMF_BUILD_TESTS option (ON by default)

ldorau:
**Fix required by oneCCL.**

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,9 +15,10 @@ list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 include(helpers)
 
 # Build Options
-option(UMF_DEVELOPER_MODE "enable developer checks, treats warnings as errors" OFF)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -71,7 +71,10 @@ install(
     EXPORT ${PROJECT_NAME}-targets)
 
 add_subdirectory(src)
-add_subdirectory(test)
+# do not include tests inside oneCCL
+if(NOT CCL_MAJOR_VERSION)
```
igchor, line 75: I think it would be better to have a generic "UMF_BUILD_TESTS" option that oneCCL (and anyone else) can switch

bratpiorka: agree - please set this option to true by default and I will disable it in oneCCL

ldorau: Done

ldorau:
@bratpiorka @igchor Done

bratpiorka (approved):

PR #31, opened 2023-11-13
add windows build and test to gha

bratpiorka:


.github/workflows/basic.yml
```diff
@@ -43,3 +43,41 @@ jobs:
         run: |
           make -j$(nproc)
           ctest --output-on-failure
+
+  windows-build:
+    name: Build - Windows
+    runs-on: windows-2019
```
igchor, line 49: I think it should be: `runs-on: ${{matrix.os}}`.

igchor:
Some of the tests are failing and it seems it's because of critnib. Looks like InterlockedOr64Acquire is not working as expected and/or there is some overflow. I will try to fix it tomorrow.

.github/workflows/basic.yml
```diff
@@ -43,3 +43,40 @@ jobs:
...
+      run: python3 -m pip install clang-format==15.0.7
+
+    - name: Configure CMake
+      run: >
+        cmake
+        -B${{github.workspace}}/build
+        -DCMAKE_C_COMPILER=${{matrix.compiler.c}}
+        -DCMAKE_CXX_COMPILER=${{matrix.compiler.cxx}}
+        -DCMAKE_POLICY_DEFAULT_CMP0094=NEW
+        -DUMF_ENABLE_POOL_TRACKING=ON
```
igchor, line 73: Should we have  `pool_tracking: ['-DUMF_ENABLE_POOL_TRACKING=ON', '-DUMF_ENABLE_POOL_TRACKING=OFF']` similarly to linux?

igchor (approved):

.github/workflows/basic.yml
bratpiorka: Right now would keep the list of build targets minimal. I think we can come back to this topic when we create a decent plan for our CI

PR #37, opened 2023-11-15
Optimize bucket selection for allocation sizes

DamianDuy:
This PR changes linear search for buckets to calculating the bucket index based on calculating the power of two (or halfway power of two) larger than Size parameter by using bitwise operations.


src/pool/disjoint/pool_disjoint_impl.cpp
```diff
@@ -290,8 +295,12 @@ class DisjointPool::AllocImpl {
         // Generate buckets sized such as: 64, 96, 128, 192, ..., CutOff.
         // Powers of 2 and the value halfway between the powers of 2.
         auto Size1 = params.MinBucketSize;
+        // MinBucketSize cannot be larger than CutOff.
+        Size1 = std::min(Size1, CutOff);
         // Buckets sized smaller than the bucket default size- 8 aren't needed.
         Size1 = std::max(Size1, UMF_DISJOINT_POOL_MIN_BUCKET_DEFAULT_SIZE);
+        // Calculate the exponent for MinBucketSize used for finding buckets.
+        MinBucketSizeExp = (size_t)log2(Size1);
```
vinser52, line 303: Can we use `getLeftmostSetBitPos()` here instead of `log2`.

src/utils/utils_math.cpp
```diff
@@ -0,0 +1,24 @@
...
+// The position of the bit is counted from 0
+// e.g. for 01000011110 the position equals 9.
+size_t getLeftmostSetBitPos(size_t num) {
+    // From C++20 countl_zero could be used for that.
+    size_t position = 0;
+    while (num > 0) {
+        num >>= 1;
+        position++;
+    }
+    position--;
```
vinser52, line 22: nit: It is better to use prefix form of increment/decrement operations in that case.

src/utils/utils_math.cpp
```diff
@@ -0,0 +1,24 @@
...
+ * Copyright (C) 2023 Intel Corporation
+ *
+ * Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ *
+ */
+
+#include "utils_math.h"
+
+// Retrieves the position of the leftmost set bit.
```
pbalcer, line 12: I think this can be replaced with `__builtin_clz`.

src/pool/disjoint/pool_disjoint_impl.cpp
```diff
@@ -815,16 +825,36 @@ void *DisjointPool::AllocImpl::allocate(size_t Size, size_t Alignment,
     return nullptr;
 }
 
-Bucket &DisjointPool::AllocImpl::findBucket(size_t Size) {
+std::size_t DisjointPool::AllocImpl::sizeToIdx(size_t Size) {
```
pbalcer, line 828: is there any speed-up after doing this?

src/utils/utils_math.cpp
vinser52: As I remember correctly, TBB library has a similar approach. They implemented log2 for integers via this builtin if it is available.

src/utils/utils_math.cpp
```diff
@@ -0,0 +1,24 @@
+/*
```
igchor, line 1: Is there a reason for this file to be in C++? I think we should default to C if possible - we might have use cases where we don't want any dependency on C++ (runtime or compiler)

ldorau (changes requested):
Please change `src/utils/utils_math.cpp` to C instead of C++

src/pool/disjoint/pool_disjoint_impl.cpp
DamianDuy: Compared to the linear search there is a noticeable speed up. From what I've seen in microbenchmark it's faster in all cases , the larger the allocation size, the bigger the difference, but even for allocation sizes like 64 or 128 (when MinBucketSize equals 8), it's approx. 4 times faster.

src/pool/disjoint/pool_disjoint_impl.cpp
DamianDuy: Done.

src/utils/utils_math.cpp
DamianDuy: Replaced this function with builtin.

src/utils/utils_math.cpp
DamianDuy: Done.

src/pool/disjoint/pool_disjoint_impl.cpp
```diff
@@ -280,6 +282,9 @@ class DisjointPool::AllocImpl {
     // Configuration for this instance
     DisjointPoolConfig params;
 
+    // Used in alghoritm for finding buckets
```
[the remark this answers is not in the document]
DamianDuy: Done.

src/utils/utils_math.cpp
```diff
@@ -0,0 +1,24 @@
...
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ *
+ */
+
+#include "utils_math.h"
+
+// Retrieves the position of the leftmost set bit.
+// The position of the bit is counted from 0
+// e.g. for 01000011110 the position equals 9.
+size_t getLeftmostSetBitPos(size_t num) {
```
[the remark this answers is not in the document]
DamianDuy: I replaced this function with a builtin and the value for `num == 0` is undefined so I added assert not zero for this value.

src/utils/utils_math.cpp
DamianDuy: Done.

src/utils/utils_math.cpp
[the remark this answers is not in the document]
DamianDuy: Replaced this function with builtin.

src/utils/utils_posix_math.c
```diff
@@ -0,0 +1,20 @@
...
+#include "utils_math.h"
+
+// Retrieves the position of the leftmost set bit.
+// The position of the bit is counted from 0
+// e.g. for 01000011110 the position equals 9.
+size_t getLeftmostSetBitPos(size_t num) {
+    assert(num != 0 &&
+           "Finding leftmost set bit when number equals zero is undefined");
+    // From C++20 countl_zero could be used for that.
+    return 31 - __builtin_clz(num);
```
PatKamin, line 19: Are you sure that `size_t` will always be 32 bit? Also, see that there are clzl and clzll variants of this builtin, too - pick one suitable for size_t type.

DamianDuy: You're right. I changed it to use `sizeof` operator.

PatKamin (approved):

ldorau (approved):
LGTM

pbalcer (approved):

src/utils/utils_posix_math.c
```diff
@@ -0,0 +1,21 @@
...
+#include "utils_math.h"
+#include <limits.h>
+
+// Retrieves the position of the leftmost set bit.
+// The position of the bit is counted from 0
+// e.g. for 01000011110 the position equals 9.
+size_t getLeftmostSetBitPos(size_t num) {
+    assert(num != 0 &&
+           "Finding leftmost set bit when number equals zero is undefined");
+    // From C++20 countl_zero could be used for that.
```
vinser52, line 19: The comment is misleading because UMF is a C library

vinser52 (approved):

src/utils/utils_posix_math.c
DamianDuy: Right. I've deleted it.

PR #49, opened 2023-11-28
Add spell check workflow in CI

PatKamin:


bratpiorka (approved):

igchor (approved):

.github/workflows/.spellcheck-conf.toml
```diff
@@ -0,0 +1,2 @@
+[default]
+extend-ignore-words-re = ["ASSERTne", "ASSER", "Tne"]
```
igchor, line 2: isn't `ASSERTne` enoguh?

PatKamin: No, in fact `ASSERTne` is not needed here. That's because the Action splits this into two words from a dictionary: `ASSERT` and `The`. I've removed `ASSERTne` and added a comment.

PR #53, opened 2023-12-01
Initial memspace and memory_target APIs

igchor:
This is an initial draft of APIs for memspace and memory_target (without any actual implementation).

A few things to discuss:
1. Should memory_target_ops be a public API? Exposing it, would allow us to implement support for some targets outside of this repo. E.g. we could implement support for UR directly where it's needed.
2. Where should the logic for creating pool/provider from a memspace reside? In this PR I proposed to make it a part of memory_target implementation. This won't work if a memspace consists of memory_targets of different_types (but we haven't explored this too much yet anyway)
3. Do we want to have type-specific queries (.e.g node_id for numa memory_target)?


include/umf/memspaces/memspace_numa.h
```diff
@@ -0,0 +1,28 @@
...
+#include <umf/memspace.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+enum umf_result_t
+umfMemspaceCreateFromNumaId(size_t id,
+                            umf_memspace_handle_t *memspace /* out */);
+
```
igchor, line 23: We should also add something to create a memspace from multiple NUMA nodes. One option is to pass an array of node ids but I think a better choice would be to have something like umfMemspaceCreateFromNumaMask() which would be more consistent with os_provider.

vinser52: Should we have just a single API to create memspace from a single node and multiple nodes? I mean that API that accepts NUMA-mask can be also used for a case with a single NUMA node. So we can remove thee the above function.

igchor: Yes, that's also an option, but in C it's a bit cumbersome to set this mask so I felt that it might be useful to have a simpler API for this case. However, we could probably just implement some helpers for for managing those masks.

igchor:
Implementation for NUMA is here: https://github.com/igchor/unified-memory-framework/pull/1

include/umf/memspace.h
```diff
@@ -0,0 +1,43 @@
...
+enum umf_result_t umfPoolCreateFromMemspace(umf_memspace_handle_t memspace,
+                                            umf_memspace_policy_handle_t policy,
+                                            umf_memory_pool_handle_t *pool);
+enum umf_result_t
+umfMemoryProviderCreateFromMemspace(umf_memspace_handle_t memspace,
+                                    umf_memspace_policy_handle_t policy,
+                                    umf_memory_provider_handle_t *provider);
+
+// TODO: iteration, filtering, sorting API
+// can we just define memspace as an array of targets?
```
bratpiorka, line 37: // can we just define memspace as an array of targets?
We can do that only if we're sure our memspaces won't contain any private data. This could be some common properties and data related to filtering or ordering - consider an example where the ordering would be "dynamic" so it would be applied before each allocation request (like order by current free space). If we support dynamic filtering or ordering, it has to be stored somewhere in the memspace priv data.

include/umf/memspaces/memspace_numa.h
bratpiorka: If we want to make this user friendly we should add an array and mask versions. Note that the array version should also support a single NUMA node case so we don't need to provide a separate function for a single NUMA node. Please also note, that if we have a version for a single node then we have to also add functions to concatenate two memspaces and I don't think we want to introduce such API right now.

include/umf/memspace.h
igchor: Do we have a use case for dynamic filtering or ordering? I don't really see a benefit of doing this. We would need to apply the filtering/ordering any time user calls iteration API anyway. So why not just apply the filtering/ordering immediately and have an array of always sorted/filtered memory_targets. With that, user can just do:

```
for (size_t i = 0; i < memspaceHandle->size; i++) {
    memspaceHandle->memoryTarget[i]->doSomething;
}
```

src/memory_target_ops.h
```diff
@@ -0,0 +1,48 @@
...
+
+struct umf_memory_target_ops_t {
+    /// Version of the ops structure.
+    /// Should be initialized using UMF_VERSION_CURRENT
+    uint32_t version;
+
+    enum umf_result_t (*initialize)(void *params, void **mem_target);
+    void (*finalize)(void *mem_target);
+
+    enum umf_result_t (*pool_create_from_memspace)(
```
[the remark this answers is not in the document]
igchor: This is an implementation detail for now.

include/umf/memspace.h
```diff
@@ -0,0 +1,43 @@
...
+
+// TODO: consider 3rd function that would only return pool ops (as opposed
+// to actual pool instance with a provider). This could be useful for UR,
+// where we could ask UMF for a best pool (e.g. based on policy) but supply
+// a custom provider (L0/CUDA, etc.).
+enum umf_result_t umfPoolCreateFromMemspace(umf_memspace_handle_t memspace,
+                                            umf_memspace_policy_handle_t policy,
+                                            umf_memory_pool_handle_t *pool);
+enum umf_result_t
+umfMemoryProviderCreateFromMemspace(umf_memspace_handle_t memspace,
```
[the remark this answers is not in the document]
igchor: I think we should expose it. Consider a case where a user wants to use memspace API for discovery but also use a custom memory pool (implemented in his application, which UMF knows nothing about).

include/umf/memspaces/memspace_numa.h
igchor: Ok, I've exposed the version with array only for now. I can also create a PR to make os provider take array instead of mask as well. 

Once we finish the implementation for NUMA, we can add some example/test that will show how to use this with bitmasks. I think we can expose something like this:

```c++
void umfBitmaskToArray(unsigned long *nodemask, unsigned long maxnode, size_t *ids);
```

igchor:
Depends on: https://github.com/oneapi-src/unified-memory-framework/pull/69

We can also not implement `umfPoolCreateFromMemspace` in this PR if you think we need to discuss the approach in #69 more.

include/umf/memory_pool.h
```diff
@@ -141,6 +141,20 @@ umf_memory_pool_handle_t umfPoolByPtr(const void *ptr);
...
+///
+/// \brief Creates a new memory pool that owns the provided memory provider. Upon the
+///        destruction of the memory pool, the associated memory provider will also be destroyed.
+/// \param ops instance of umf_memory_pool_ops_t
+/// \param provider memory provider that will be used for coarse-grain allocations.
+/// \param params pointer to pool-specific parameters
+/// \param hPool [out] handle to the newly created memory pool
+/// \return UMF_RESULT_SUCCESS on success or appropriate error code on failure.
+///
+enum umf_result_t umfPoolCreateOwning(const struct umf_memory_pool_ops_t *ops,
```
vinser52, line 153: We have a separate PR #69 for that right? Just to duplicate my comment from PR #69 let's keep this function private unless we have a strong demand to make it public.

include/umf/memspace.h
vinser52: I agree with Igor that the dynamic allocation policy (if it is allowed) should be handled by the memory provider because the provider is responsible for actual memory allocation from the memory target. Memory provider is instantiated and configured based on input parameters of `umfMemoryProviderCreateFromMemspace` function.
But anyway we need to decide what is encapsulated by each parameter. Form my perspective memspace can encapsulate information about memory location (e.g. memory targets) and policy can define how memory location(s) are used (e.g. local first, interleaving, interleaving size, etc.).

Any thoughts/comments/objections?

src/memory_pool_internal.h
```diff
@@ -24,6 +26,9 @@ typedef struct umf_memory_pool_t {
 
     // Memory provider used by the pool.
     umf_memory_provider_handle_t provider;
+
+    // Tells whether memory provider is owned by the pool.
+    bool own_provider;
```
vinser52, line 31: an alternative approach could be a pointer to the "destructor function" instead of just storing a `bool` flag.
But it is a minor comment and we can keep it as is at least now.

src/memory_target_ops.h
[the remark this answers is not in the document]
vinser52: I agree that architecture discussion is needed, but my understanding is the following. Starting with this PR and until we haven't decided to expose it as a public API we have been gaining implementation experience. So let's see how it will glue together (Memspace API on top of low-level pool creation API).

src/memory_target_ops.h
```diff
@@ -0,0 +1,48 @@
...
+        umf_memspace_handle_t memspace, void **mem_targets, size_t num_targets,
```
vinser52, line 33: If memspace is a set of memory targets why do we need to pass `mem_targets` array and its size (`num_targets`) as additional parameters?

src/memory_target_ops.h
```diff
@@ -0,0 +1,48 @@
...
+        umf_memspace_policy_handle_t policy, umf_memory_pool_handle_t *pool);
+
+    enum umf_result_t (*memory_provider_create_from_memspace)(
```
vinser52, line 36: Interesting design question: Who (which entity) should be responsible for choosing a pool manager?
If a user calls `umfMemoryProviderCreateFromMemspace` then the user will explicitly specify the pool manager when the `umfPoolCreate` function is called.
I am mainly interested in the case when a user calls 'umfPoolCreateFromMemspace'. There possible candidates:

- Memspace implementation chooses the heap manager
- UMF implementation chooses the heap manager based on the memspace or memory provider instance.

Another consideration to keep in mind is that some heap managers are provided as static libraries (not part of libumf).

Let's think about that and discuss it in the next sync.

src/memory_target_ops.h
igchor: This is because `mem_targets` are pointers to private memory target data (to the structure that was initialized by memory_target_ops::initialize) whereas memspace->nodes points to the internal `struct umf_memory_target_t`. See here how they are translated: https://github.com/oneapi-src/unified-memory-framework/blob/ad096207c6ff8a9e7fa098c39682e1de7e56f4a7/src/memspace.c#L58

This is not strictly necessary for now but it will be necessary when we decide to expose memory_target_ops as public interface in the future (we don't want to leak `struct umf_memory_target_t` definition).

src/memory_target_ops.h
igchor: We have 2 entities that impact which pool/provider should be used: memspace (i.e collection of memory_targets) and policy.

In C++, we could probably make it a 'cusomization point' and let user/developer just define this function for each combination of policy/memory targets he's interested in. In C, I think we have to put this logic either in memory_target_ops or in policy somehow. I've chosen to put it in memory_target_ops since it's easier for now (we already have the ops abstraction). However, I actually belive that policy would be a better place (since it would allow us to mix tagets).

We could implement policy using `policy_ops` structure (similarly as memory_target, pool, provider) and have those `pool_from_memspace` and `provider_from_memspace` be part of that. We can then have a default policy that would be a fallback and also allow users or memory_target developers to implement their own policies that would have some custom logic for choosing the right pool/provider. This should work pretty well with tiering for example: we would have a 'tiered' policy and the core logic would be just in `pool_from_memspace` in that policy.

include/umf/memspace.h
igchor: Agree. However, I think we can have some 'dynamic' metrics as part of the memspace: total available memory in the system, etc. that could be queried. But I don't think making the memspace dynamically sorted based on this information is a good idea. Especially since those metrics can changed right after the memspace has been sorted and we'll have inconsistent state.

src/memory_target_ops.h
igchor: @vinser52 @bratpiorka I made the pool_create_from_memspace return NOT_SUPPORTED for now. This function is not strictly needed for now.

src/memory_targets/memory_target_numa.c
```diff
@@ -0,0 +1,131 @@
...
+        .maxnode = 0, // TODO: numa_max_node/GetNumaHighestNodeNumber
+        .numa_mode = UMF_NUMA_MODE_DEFAULT,
+        .numa_flags = UMF_NUMA_FLAGS_STRICT, // TODO: determine default behavior
+
+        // Logging
+        .traces = 0, // TODO: parse env variable for log level?
+};
+
+static unsigned long
+numa_targets_get_nodemask(struct numa_memory_target_t **targets,
```
bratpiorka, line 60: I think we should return here nodemask_t from libnuma. Please note that this mask could exceed unsigned long capacity. Also as libnuma already implements some functions that handle these types of mask I think we should use it, so please consider calling numa_bitmask_setbit() here.

src/memspace.c
```diff
@@ -0,0 +1,109 @@
...
+    if (privs == NULL) {
+        return UMF_RESULT_ERROR_OUT_OF_HOST_MEMORY;
+    }
+
+    for (size_t i = 0; i < memspace->size; i++) {
+        privs[i] = memspace->nodes[i]->priv;
+    }
+    *pPrivs = privs;
+
+    *pPrivs = privs;
```
bratpiorka, line 50: this line is doubled

bratpiorka (changes requested):

include/umf/memspace.h
```diff
@@ -0,0 +1,58 @@
...
+typedef struct umf_memspace_t *umf_memspace_handle_t;
+
+///
+/// \brief Creates new memory pool from memspace and policy.
+/// \param hMemspace handle to memspace
+/// \param hPolicy handle to policy
+/// \param hPool [out] handle to the newly created memory pool
+/// \return UMF_RESULT_SUCCESS on success or appropriate error code on failure.
+///
+enum umf_result_t
```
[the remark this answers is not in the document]
igchor: Done.

src/memory_target.c
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+#include <assert.h>
+#include <stdlib.h>
+
+#include "memory_target.h"
+#include "memory_target_ops.h"
+
+enum umf_result_t
+umfMemoryTargetCreate(const struct umf_memory_target_ops_t *ops, void *params,
```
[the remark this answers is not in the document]
igchor: Done.

src/memory_target.h
```diff
@@ -0,0 +1,38 @@
...
+#ifndef UMF_MEMORY_TARGET_H
+#define UMF_MEMORY_TARGET_H 1
+
+#include <umf/base.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+struct umf_memory_target_ops_t;
```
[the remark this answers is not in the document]
igchor: Done.

src/memspace.c
igchor: Done. Rebase conflict.

src/memory_targets/memory_target_numa.c
igchor: Done.

igchor: But tbh, this will need to be changed to something cross-platform anyway. We might use hwloc instead of numa to make it work on windows as well (same is true for os provider).

bratpiorka (approved):

bratpiorka:
@vinser52 I think we could go with this code - I know it still needs some changes in the future, but I think it is a good basis for introducing the memspace API

include/umf/memspace.h
```diff
@@ -0,0 +1,57 @@
...
+umf_result_t
+umfMemoryProviderCreateFromMemspace(umf_memspace_handle_t hMemspace,
+                                    umf_memspace_policy_handle_t hPolicy,
+                                    umf_memory_provider_handle_t *hProvider);
+
+///
+/// \brief Destroys memspace
+/// \param hMemspace handle to memspace
+///
+void umfMemspaceDestroy(umf_memspace_handle_t hMemspace);
```
vinser52, line 51: Where is the corresponding create function?

Another question: if we consider only predefined memspaces as a first step do we need Create/Destroy functions for memspaces in a public API?

igchor: Right now we only have `umfMemspaceCreateFromNumaArray` that is defined in memspace_numa.h

We could move that function to memspace.h for now but in the long term, I think those create functions should be in separate headers (since we might want to have the memspaces in separate libs).

igchor: > Another question: if we consider only predefined memspaces as a first step do we need Create/Destroy functions for memspaces in a public API?

Ok, good point. @bratpiorka for openMP, we won't need `umfMemspaceCreateFromNumaArray` right? If that's the case I can make it internal for now.

igchor: Done. I've made those functions private for now.

igchor: This PR does not expose the predefined memspaces but this will be added in next PR (it requires integrating with hwloc).

bratpiorka:
@igchor please rebase
@vinser52 can we merge the PRs? I think we could implement the next steps in separate PRs.

igchor:
> @igchor please rebase @vinser52 can we merge this PR? I think we could implement the next steps in separate PRs.

Rebased.

igchor:
@vinser52 please take a look. We'd like to unblock next step which is implementing HOST_MEMSPACE

src/memory_target_ops.h
```diff
@@ -0,0 +1,46 @@
...
+
+    umf_result_t (*initialize)(void *params, void **memoryTarget);
+    void (*finalize)(void *memoryTarget);
+
+    umf_result_t (*pool_create_from_memspace)(
+        umf_memspace_handle_t memspace, void **memoryTargets, size_t numTargets,
+        umf_memspace_policy_handle_t policy, umf_memory_pool_handle_t *pool);
+
+    umf_result_t (*memory_provider_create_from_memspace)(
+        umf_memspace_handle_t memspace, void **memoryTargets, size_t numTargets,
```
vinser52, line 37: what if `memspace` consists of multiple `memory_targets`? How we will choose  which `memory_target` should we ask to create a pool or provider from `memspace`?

src/memspaces/memspace_numa.c
```diff
@@ -0,0 +1,62 @@
...
+ */
+
+#include <stdlib.h>
+
+#include "../memory_targets/memory_target_numa.h"
+#include "../memspace_internal.h"
+#include "memspace_numa.h"
+
+enum umf_result_t
+umfMemspaceCreateFromNumaArray(size_t *nodeIds, size_t numIds,
```
vinser52, line 17: why not just create a memspace from the array of memory targets?

src/memory_target_ops.h
igchor: That's why we only support memory targets of the same type for now. See here: https://github.com/oneapi-src/unified-memory-framework/blob/9c4e3dacc99ec0a303f35357eed0c18acd41ad61/src/memspace.c#L68

src/memspaces/memspace_numa.c
igchor: This is just a wrapper that does exactly that underneath. It will be used for implementing HOST_MEMSPACE.

Also, this function was initially public (and we don't expose memory_targets in the public API for now) You suggested to hide this function for now which I did. However, we do plan to expose this function in future (possibly before we expose memory_targets).

src/memory_target_ops.h
vinser52: I understand that today we support only memory targets of the same type (same ops structure). My question is mostly about the architecture/design, not the current implementation. if we are creating pool/provider from memspace perhaps the current design is wrong and memspace should be responsible for pool/provider creation? In the future I can imagine memspace which consists of GPU memory target and numa memory target - in that case I expect that we will choose GPU memory provider (L0 or cuda) and create USM shared pool.

What do you think?

vinser52 (approved):

PR #56, opened 2023-12-05
Add TBB pool manager with tests

ldorau:


include/umf/pools/pool_tbb.h
```diff
@@ -0,0 +1,30 @@
...
+#endif
+
+#include <umf/memory_pool.h>
+#include <umf/memory_provider.h>
+
+struct umf_TBB_pool_config {
+    int TODO; // TODO
+};
+
+extern struct umf_memory_pool_ops_t UMF_TBB_POOL_OPS;
```
vinser52, line 24: Should we rename it to the "scalable pool" from "TBB pool"?

src/CMakeLists.txt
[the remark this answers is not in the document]
ldorau: OK, thanks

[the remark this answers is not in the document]
bratpiorka: @vinser52 this is not relevant to this PR, but should we also store the Disjoint Pool in a static library?

src/pool/pool_tbb.c
```diff
@@ -0,0 +1,200 @@
...
+                                       .fixed_pool = false,
+                                       .keep_all_memory = false,
+                                       .reserved = 0};
+
+    pthread_once(&tbb_is_initialized, load_tbb_symbols);
+
+    struct tbb_memory_pool *pool_data = malloc(sizeof(struct tbb_memory_pool));
+    if (!pool_data) {
+        fprintf(stderr, "cannot allocate memory for metadata\n");
+        abort();
```
igchor, line 138: Just return an error

src/pool/pool_tbb.c
```diff
@@ -0,0 +1,200 @@
...
+    *(void **)&tbb_ops.pool_msize =
+        dlsym(tbb_handle, "_ZN3rml10pool_msizeEPNS_10MemoryPoolEPv");
+
+    if (!tbb_ops.pool_malloc || !tbb_ops.pool_realloc ||
+        !tbb_ops.pool_aligned_malloc || !tbb_ops.pool_free ||
+        !tbb_ops.pool_create_v1 || !tbb_ops.pool_destroy ||
+        !tbb_ops.pool_identify) {
+        fprintf(stderr, "Could not find symbols in %s.\n", so_name);
+        dlclose(tbb_handle);
+        abort();
```
igchor, line 89: Return an error.

src/pool/pool_tbb.c
```diff
@@ -0,0 +1,200 @@
...
+static struct tbb_callbacks g_tbb_ops;
+static pthread_once_t tbb_is_initialized = PTHREAD_ONCE_INIT;
+
+void load_tbb_symbols(void) {
+    fprintf(stdout, "Initializing TBB symbols\n");
+    const char so_name[] = "libtbbmalloc.so.2";
+    void *tbb_handle = dlopen(so_name, RTLD_LAZY);
+    if (!tbb_handle) {
+        fprintf(stderr, "%s not found.\n", so_name);
+        abort();
```
igchor, line 60: Return an error.

src/pool/pool_tbb.c
```diff
@@ -0,0 +1,200 @@
...
+    }
+
+    return resPtr;
+}
+
+void tbb_raw_free_wrapper(intptr_t pool_id, void *ptr, size_t bytes) {
+    struct tbb_memory_pool *pool = (struct tbb_memory_pool *)pool_id;
+    enum umf_result_t ret =
+        umfMemoryProviderFree(pool->mem_provider, ptr, bytes);
+    if (ret != UMF_RESULT_SUCCESS) {
```
igchor, line 112: Not really related to this PR, but wouldn't it be better to just implement a 'logging' memory provider (and perhaps a pool as well) that we could wrap around the actual provider (similarly to how we use tracking memory provider)? This way we wouldn't have to add those logs everywhere.

src/pool/pool_tbb.c
```diff
@@ -0,0 +1,200 @@
...
+        abort();
+    }
+
+    g_tbb_ops = tbb_ops;
+}
+
+void *tbb_raw_alloc_wrapper(intptr_t pool_id, size_t *raw_bytes) {
+    void *resPtr;
+    struct tbb_memory_pool *pool = (struct tbb_memory_pool *)pool_id;
+    enum umf_result_t ret =
```
igchor, line 98: Probably we should store the error in TLS so we can return it (potentially after some translation) in get_last_allocation error.

src/pool/pool_tbb.c
```diff
@@ -0,0 +1,200 @@
...
+    if (ret != UMF_RESULT_SUCCESS) {
+        return NULL;
+    }
+
+    return resPtr;
+}
+
+void tbb_raw_free_wrapper(intptr_t pool_id, void *ptr, size_t bytes) {
+    struct tbb_memory_pool *pool = (struct tbb_memory_pool *)pool_id;
+    enum umf_result_t ret =
```
igchor, line 110: Probably we should store the error in TLS and return it from tbb_free if necessary. Or perhaps we could just store it in the tbb_memory_pool struct?

src/CMakeLists.txt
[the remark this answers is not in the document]
vinser52: It is a good question. The only reason to make a static library for Disjoint Pool is C++ dependency. If Disjoint Pool does not require C++ dependency I would say it can be embedded in the UMF library.

src/pool/pool_tbb.c
ldorau: Done

src/pool/pool_tbb.c
ldorau: Cannot - this function has to return `void`, because it is an argument in `pthread_once()`

src/pool/pool_tbb.c
ldorau: Cannot - this function has to return `void`, because it is an argument in `pthread_once()`

src/CMakeLists.txt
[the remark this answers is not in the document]
ldorau: > Just a reminder, as we agreed TBB pool manager should be a separate static library. It should not be part of UMF by default.

Done.

src/pool/pool_tbb.c
ldorau: `get_last_allocation_error()` retrieves `umf_result_t` representing the error of the last failed **allocation operation** in this thread (**malloc, calloc, realloc, aligned_malloc**) - not free.

src/pool/pool_tbb.c
ldorau: Done

ldorau:
Tests added

ldorau:
Ready for review

src/pool/pool_tbb.c
igchor: In that case, I think you should just store the result in some global variable and check it in tbb_pool_initialize. We don't ever want to crash when creating a memory pool/provider.

src/pool/pool_tbb.c
igchor: Same as above.

src/pool/pool_tbb.c
igchor: That's true, but we can return the result directly from umfPoolFree (`tbb_free`). Right now we always return SUCCESS from what I see.

README.md
```diff
@@ -39,6 +39,17 @@ The `UMF_BUILD_LIBUMF_POOL_JEMALLOC` option has to be turned `ON` to build this
 2) Required packages:
 - libjemalloc-dev
 
+### libumf_pool_tbb (Linux-only)
```
igchor, line 42: I think we should name it 'pool_scalable' as Sergey suggested. That's how we were referencing it on the slides, etc.

src/pool/pool_tbb.c
```diff
@@ -0,0 +1,238 @@
...
+static void *tbb_calloc(void *pool, size_t num, size_t size) {
+    assert(pool);
+    size_t csize = num * size;
+    void *ptr = tbb_malloc(pool, csize);
+    if (ptr == NULL) {
+        // TLS_last_allocation_error is set by tbb_malloc()
+        return NULL;
+    }
+
+    memset(ptr, 0, csize); // TODO: device memory is not accessible by host
```
igchor, line 180: I think this TODO is not needed here. I don't think we can use tbb pool with USM anyway (in TBB, metadata is stored alongside the data).

src/pool/pool_tbb.c
```diff
@@ -0,0 +1,238 @@
...
+static umf_result_t tbb_free(void *pool, void *ptr) {
+    struct tbb_memory_pool *pool_data = (struct tbb_memory_pool *)pool;
+    g_tbb_ops.pool_free(pool_data->tbb_pool, ptr);
+    return UMF_RESULT_SUCCESS;
+}
+
+static size_t tbb_malloc_usable_size(void *pool, void *ptr) {
+    (void)pool; // not used
+    (void)ptr;  // not used
+    return 0;   // unsupported
```
igchor, line 220: I think there is pool_msize function in tbb. Can't we use it?

igchor (changes requested):

src/pool/pool_tbb.c
ldorau: Done

src/pool/pool_tbb.c
ldorau: Done

README.md
ldorau: Done

include/umf/pools/pool_tbb.h
ldorau: Done

src/pool/pool_tbb.c
ldorau: Right. Done.

src/pool/pool_tbb.c
ldorau: Right. Done.

src/pool/pool_tbb.c
ldorau: I think there already is a task for that in our Jira, isn't there @bratpiorka ?

src/pool/pool_tbb.c
ldorau: In `tbb_free()` yes, we can, but a return value of `pool_free()` is never used or checked in the TBB repo ... Done anyway.

src/pool/pool_scalable.c
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    struct tbb_memory_pool *pool_data = malloc(sizeof(struct tbb_memory_pool));
+    if (!pool_data) {
+        fprintf(stderr, "cannot allocate memory for metadata\n");
+        return UMF_RESULT_ERROR_OUT_OF_HOST_MEMORY;
+    }
+
+    pool_data->mem_provider = provider;
+    g_tbb_ops.pool_create_v1((intptr_t)pool_data, &policy,
```
igchor, line 151: this functions returns a status, shouldn't we check it?

igchor (approved):

igchor (approved):

PR #57, opened 2023-12-05
WIP: pool memory resource that can be used with an upstream memory resource

igchor:


vinser52:
Some general question:
On the last WG meeting we agreed that we do not need to implement memory resources on top of UMF (at least not now), correct? Instead we are discussing to have memory resource implementation on top of SYCL.
So why do we need this PR?

igchor:
@vinser52 that's correct, I just want to have something I can use for benchmarking/testing once we have the SYCL provider and different pool implementations (jemalloc). This PR just exposes some simple abstractions to build a pool memory_resource. I don't plan to have it merged for now. Would you prefer if I just kept it on a separate branch and not have the PR open?

vinser52:
Thank you for your clarifications.

> Would you prefer if I just kept it on a separate branch and not have the PR open?

Keeping it in a separate fork is OK. Regarding the branch instead of the fork do you mean creating a separate branch in this repo? Regarding PR, what is your intention by opening the PR? If you need a review than it totally makes sense and we can keep this PR and review/discuss the code.

igchor:
My intention was just to have this visible so I can reference this PR when discussing the SYCL provider, show some examples, etc.

bratpiorka:
outdated

PR #58, opened 2023-12-07
Add script for generating API html docs

PatKamin:
Add a script that generates HTML documentation for the UMF API.

Fix and extend doxygen comments to fully cover UMF API.


include/umf/memory_pool_ops.h
```diff
@@ -41,13 +41,72 @@ struct umf_memory_pool_ops_t {
     /// \param pool pool to finalize
     void (*finalize)(void *pool);
 
-    /// Refer to memory_pool.h for description of those functions
+    /// \brief Allocates \p size bytes of uninitialized storage of \p pool
```
vinser52, line 44: Could you please also align comments with the corresponding functions from the `memory_pool.h` file

include/umf/memory_pool_ops.h
```diff
@@ -41,13 +41,72 @@ struct umf_memory_pool_ops_t {
...
+    /// \param pool memory pool handle
+    /// \param size number of bytes to allocate
+    /// \return Pointer to the allocated memory
+    ///
     void *(*malloc)(void *pool, size_t size);
+
+    /// \brief Allocates memory of \p pool for an array of \p num elements
+    ///        of \p size bytes each and initializes all bytes in the allocated storage
+    ///        to zero
+    /// \param pool memory pool handle
```
igchor, line 54: I think we should drop 'handle' here and perhaps have just "pointer to the memory pool" to avoid confusion. This parameter is NOT the same as pool parameter in the umfPool* functions.

PatKamin:
> Could you please also align comments with the corresponding functions from the `memory_pool.h` file 

Done

PatKamin:
> Could you please also align comments with the corresponding functions from the `memory_provider.h` file 

Done

PatKamin:
> I think we should drop 'handle' here and perhaps have just "pointer to the memory pool" to avoid confusion. This parameter is NOT the same as pool parameter in the umfPool* functions. 

Done

PatKamin:
I've added to this PR a commit with the GitHub Pages deployment. You can see the rendered version of docs on my [fork](https://patkamin.github.io/unified-memory-framework/) now.

bratpiorka:
Why is the theme different from what uses UR here? https://oneapi-src.github.io/unified-runtime/ 
Could we use the same one?

include/umf/memory_provider_ops.h
```diff
@@ -29,25 +31,113 @@ struct umf_memory_provider_ops_t {
...
     /// \brief Finalizes memory provider.
     /// \param provider provider to finalize
+    ///
     void (*finalize)(void *provider);
 
-    /// Refer to memory_provider.h for description of those functions
+    ///
+    /// \brief Allocates \p size bytes of uninitialized storage from memory \p provider
+    ///        with the specified \p alignment
+    /// \param provider pointer to the memory pool
```
igchor, line 46: 'provider', not 'pool'

PatKamin: Done

PatKamin:
> Why is the theme different from what uses UR here? https://oneapi-src.github.io/unified-runtime/ 
Could we use the same one? 

I liked this one beteter :) I've changed it to the same theme as in UR, though.

igchor (approved):

include/umf/memory_pool_ops.h
```diff
@@ -17,37 +17,108 @@
...
+    /// @brief Initializes memory pool.
+    /// @param providers array of memory providers that will be used for coarse-grain allocations.
     ///        Should contain at least one memory provider.
-    /// \param numProvider number of elements in the providers array
-    /// \param params pool-specific params
-    /// \param pool [out] returns pointer to the pool
-    /// \return UMF_RESULT_SUCCESS on success or appropriate error code on failure.
+    /// @param numProvider number of elements in the providers array
+    /// @param params pool-specific params
+    /// @param pool [out] returns pointer to the pool
```
vinser52, line 36: maybe we need to describe better that returned `pool` pointer (which is not equal to the pool handle) points to the pool-specific private data that is used as an input parameter for subsequent `malloc`, `free`, etc. callbacks of that pool.

@pbalcer, @igchor, @bratpiorka what do you think?

bratpiorka: I agree, but remember that we have plans to improve all the documentation later anyway - this PR focues on generation. BTW, what I don't like about the oneAPI documentation is that it often focuses mainly on parameters rather than helping the user understand what the function does. So in addition to this kind of API documentation, we should also publish an overview (what we have in our HLD) and a user guide.

vinser52: > we should also publish an overview (what we have in our HLD) and a user guide.

But in a separate PR, right?

bratpiorka: right

PatKamin (commented):
> maybe `of \p hPool` change to `from \p hPool`?
And in all places below. 

Done

include/umf/memory_pool_ops.h
```diff
@@ -17,37 +17,108 @@
...
-    /// \brief Finalizes memory pool
-    /// \param pool pool to finalize
+    /// @brief Finalizes memory pool
+    /// @param pool pool to finalize
+    ///
     void (*finalize)(void *pool);
 
-    /// Refer to memory_pool.h for description of those functions
+    ///
+    /// @brief Allocates \p size bytes of uninitialized storage of \p pool
```
vinser52, line 49: Since you fixed comments in the `memory_pool.h` (`of \p pool` -> `from \p pool`). We should fix this file in the same way.

PatKamin: I've omitted this one place, thanks.

PR #64, opened 2023-12-11
Add malloc compliance tests

ldorau:
They will be useful to test pool managers.

See https://github.com/oneapi-src/unified-memory-framework/pull/65 for an example.

### DIFF
--- a/test/common/test_helpers.c
+++ b/test/common/test_helpers.c
@@ -4,6 +4,9 @@
 // This file contains tests for UMF pool API
 
 #include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
 
 #include "pool_null.h"
 #include "pool_trace.h"
@@ -11,6 +14,23 @@
 #include "provider_trace.h"
 #include "umf/memory_pool.h"
 #include "umf/memory_provider.h"
+
+#include "test_helpers.h"
+
+// Check if the memory is filled with the given character
+int is_filled_with_char(void *ptr, size_t size, char c) {
+    char *mem = (char *)ptr;
+    return (*mem == c) && memcmp(mem, mem + 1, size - 1) == 0;
+}
+
+// Check if two memory regions has the same content
+int is_same_content(void *first, void *second, size_t size) {
+    return memcmp(first, second, size) == 0;
+}
+
+int is_aligned(void *ptr, size_t alignment) {
+    return ((uintptr_t)ptr & (alignment - 1)) == 0;
+}
 
 umf_memory_provider_handle_t nullProviderCreate(void) {
     umf_memory_provider_handle_t hProvider;
--- a/test/common/test_helpers.h
+++ b/test/common/test_helpers.h
@@ -14,6 +14,12 @@
 extern "C" {
 #endif
 
+int is_filled_with_char(void *ptr, size_t size, char c);
+
+int is_same_content(void *first, void *second, size_t size);
+
+int is_aligned(void *ptr, size_t alignment);
+
 umf_memory_provider_handle_t nullProviderCreate(void);
 umf_memory_provider_handle_t
 traceProviderCreate(umf_memory_provider_handle_t hUpstreamProvider,
--- a/test/malloc_compliance_tests.cpp
+++ b/test/malloc_compliance_tests.cpp
@@ -1,0 +1,151 @@
+// Copyright (C) 2023 Intel Corporation
+// Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+// Basic tests
+// ISO/IEC 9899:201x 7.22.3 compliance
+
+#include <errno.h>
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include "malloc_compliance_tests.hpp"
+#include "test_helpers.h"
+#include "umf/memory_pool.h"
+
+#include "base.hpp"
+using umf_test::test;
+
+//------------------------------------------------------------------------
+// Configurable defs
+//------------------------------------------------------------------------
+
+#define MAX_ALLOC_SIZE (1024 * 1024) // 1 MB
+#define ITERATIONS 100
+#define SRAND_INIT_VALUE 0
+
+//------------------------------------------------------------------------
+// Helpers
+//------------------------------------------------------------------------
+
+static inline size_t rand_alloc_size(int max) {
+    srand(SRAND_INIT_VALUE);
+    return rand() % max;
+}
+
+static inline void free_memory(umf_memory_pool_handle_t hPool,
+                               void *ptr[ITERATIONS]) {
+    for (int i = 0; i < ITERATIONS; i++) {
+        umfPoolFree(hPool, ptr[i]);
+    }
+}
+
+//------------------------------------------------------------------------
+// Tests
+// -----------------------------------------------------------------------
+
+// ISO/IEC 9899:TC3 7.20.3.3
+void malloc_compliance_test(umf_memory_pool_handle_t hPool) {
+    void *alloc_ptr[ITERATIONS];
+    size_t alloc_ptr_size[ITERATIONS];
+
+    // Each allocation shall yield a pointer to an object disjoint from
+    // any other object.
+    for (int i = 0; i < ITERATIONS; i++) {
+        alloc_ptr_size[i] = rand_alloc_size(MAX_ALLOC_SIZE);
+        alloc_ptr[i] = umfPoolMalloc(hPool, alloc_ptr_size[i]);
+
+#ifndef _WIN32 // TODO: explain why it fails on some Windows systems
+        // Supported under USE_WEAK_ALIGNMENT_RULES macro
+        // For compatibility, on 64-bit systems malloc should align to 16 bytes
+        size_t alignment =
+            (sizeof(intptr_t) > 4 && alloc_ptr_size[i] > 8) ? 16 : 8;
+        ASSERT_NE(is_aligned(alloc_ptr[i], alignment), 0)
+            << "Malloc should return pointer that is suitably aligned so that "
+               "it may be assigned to a pointer to any type of object";
+#endif
+
+        // Fill memory with a pattern ((id of the allocation) % 256)
+        ASSERT_NE(alloc_ptr[i], nullptr)
+            << "malloc returned NULL, couldn't allocate much memory";
+        memset(alloc_ptr[i], i % 0xFF, alloc_ptr_size[i]);
+    }
+    for (int i = 0; i < ITERATIONS; i++) {
+        ASSERT_NE(
+            is_filled_with_char(alloc_ptr[i], alloc_ptr_size[i], i % 0xFF), 0)
+            << "Object returned by malloc is not disjoined from others";
+        memset(alloc_ptr[i], 1, alloc_ptr_size[i]);
+    }
+    free_memory(hPool, alloc_ptr);
+}
+
+// ISO/IEC 9899:TC3 7.20.3.1
+void calloc_compliance_test(umf_memory_pool_handle_t hPool) {
+    void *alloc_ptr[ITERATIONS];
+    size_t alloc_size;
+
+    // Checking that the memory returned by calloc is zero filled
+    for (int i = 0; i < ITERATIONS; i++) {
+        alloc_size = rand_alloc_size(MAX_ALLOC_SIZE);
+        alloc_ptr[i] = umfPoolCalloc(hPool, i + 1, alloc_size);
+
+        ASSERT_NE(alloc_ptr[i], nullptr)
+            << "calloc returned NULL, couldn't allocate much memory";
+        ASSERT_NE(is_filled_with_char(alloc_ptr[i], alloc_size, 0), 0)
+            << "Memory returned by calloc was not zeroed";
+    }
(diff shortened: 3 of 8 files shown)


bratpiorka:
We should somehow standardize all the tests - both in terms of the style of their creation and the way they are built, run and collect results. Could we use gtest to run these tests? Please also use TEST_P, where we could pass different implementations of the pools.

igchor:
> Please also use TEST_P, where we could pass different implementations of the pools.

You can use some of the fixtures we have defined in test/memoryPool.hpp

test/malloc_compliance_tests.c
```diff
@@ -0,0 +1,175 @@
...
+#define MB (1024 * KB)
+
+#define MAX_ALLOC_SIZE (1 * MB)
+#define ITERATIONS 100
+
+//------------------------------------------------------------------------
+// Helpers
+//------------------------------------------------------------------------
+
+#define ASSERT(expr, msg)                                                      \
```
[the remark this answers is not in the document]
ldorau: Check now, please.

test/malloc_compliance_tests_glibc.cpp
```diff
@@ -0,0 +1,80 @@
...
+    /* .get_last_allocation_error = */ NULL,
+};
+
+static umf_memory_pool_t UMF_GLIBC_POOL = {
+    /* .pool_priv = */ POOL_PRIV,
+    /* .ops = */ UMF_GLIBC_POOL_OPS,
+    /* .provider = */ NULL,
+};
+
+TEST_F(test, glibc_malloc_compliance_test) {
```
igchor, line 66: Why not `umfPoolTest `? With `umfPoolTest` and TEST_P instead of TEST_F it would be possible to run those testcases for multiple pool types, without having to manually instantiate each case for each pool. See here as an example: https://github.com/oneapi-src/unified-memory-framework/blob/main/test/disjoint_pool.cpp#L163

ldorau: Why is the current solution wrong? Would the one with `umfPoolTest` and `TEST_P` be simpler than this?
```c
TEST_F(test, glibc_malloc_compliance_test) {
    malloc_compliance_test(&UMF_GLIBC_POOL);
}

TEST_F(test, glibc_calloc_compliance_test) {
    calloc_compliance_test(&UMF_GLIBC_POOL);
}

TEST_F(test, glibc_realloc_compliance_test) {
    realloc_compliance_test(&UMF_GLIBC_POOL);
}

TEST_F(test, glibc_free_compliance_test) {
    free_compliance_test(&UMF_GLIBC_POOL);
}
```

test/malloc_compliance_tests.cpp
```diff
@@ -0,0 +1,156 @@
...
+#define MB (1024 * KB)
+
+#define MAX_ALLOC_SIZE (1 * MB)
+#define ITERATIONS 100
+
+//------------------------------------------------------------------------
+// Helpers
+//------------------------------------------------------------------------
+
+#define ASSERT_MSG(expr, msg)                                                  \
```
pbalcer, line 35: in gtest I believe you can do something like `ASSERT(whatever) << "my error message"`.

test/malloc_compliance_tests_glibc.cpp
pbalcer: `TEST_P` gives you the ability to parametrize your tests, so instead of hardcoding umf_glibc_pool, you could use a collection of all the pool implementations as defined by the `umfPoolTest` fixture. So you'd only have one malloc test, not one for each possible implementation.

test/malloc_compliance_tests.cpp
```diff
@@ -0,0 +1,156 @@
...
+    memset(ret, 1, 10);
+    umfPoolFree(hPool, ret);
+
+    // The contents of the new object shall be the same
+    // as that of the old object prior to deallocation
+    void *realloc_ptr[ITERATIONS];
+    size_t alloc_size;
+    for (int i = 0; i < ITERATIONS; i++) {
+        // Generate allocation size
+        alloc_size = rand_alloc_size(MAX_ALLOC_SIZE);
```
pbalcer, line 123: I don't like adding randomness in tests. Can this be just a list of sizes?

test/malloc_compliance_tests.cpp
```diff
@@ -0,0 +1,156 @@
...
+            "Malloc should return pointer that is suitably aligned so that "
+            "it may be assigned to a pointer to any type of object ");
+
+        // Fill memory with zeros
+        ASSERT_MSG(alloc_ptr[i],
+                   "malloc returned NULL, couldn't allocate much memory");
+        memset(alloc_ptr[i], 0, alloc_ptr_size[i]);
+    }
+    for (int i = 0; i < ITERATIONS; i++) {
+        ASSERT_MSG(is_zero_mem(alloc_ptr[i], alloc_ptr_size[i]),
```
pbalcer, line 82: is this a reliable way to check this? wouldn't it be better to fill the buffer with a different pattern for each allocation (e.g., the id of the allocation, and then cap the iters at 256)?

test/malloc_compliance_tests_glibc.cpp
```diff
@@ -0,0 +1,80 @@
...
+}
+
+static enum umf_result_t glibc_free(void *pool, void *ptr) {
+    (void)pool; // unused
+    assert(pool == POOL_PRIV);
+    free(ptr);
+    return UMF_RESULT_SUCCESS;
+}
+
+static const umf_memory_pool_ops_t UMF_GLIBC_POOL_OPS = {
```
bratpiorka, line 47: Please extract this pool to separate files similar to "pool_null" that we have right now. Also please note that glibc - based poll is already created here: https://github.com/oneapi-src/unified-memory-framework/blob/main/test/common/pool.hpp#L112C8. @pbalcer @igchor we should decide to go with C++ approach already used in disjoint pool tests or with this one.  I like this one because of simplicity.. but if you disagree please add more comments to disjoint pool tests because I don't understand this code

test/malloc_compliance_tests_glibc.cpp
bratpiorka: @ldorau please run these tests on the disjoint pool also and use TEST_P in this case. I think then we could compare both approaches and choose one.

bratpiorka (changes requested):

test/malloc_compliance_tests_glibc.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+#include "../src/memory_pool_internal.h"
+#include "malloc_compliance_tests.hpp"
+#include <umf/memory_pool_ops.h>
+
+// casting to (uintptr_t) because of:
+// warning C4312: 'type cast': conversion from 'unsigned int' to 'void *' of greater size
+#define POOL_PRIV ((void *)(uintptr_t)0xDEADBEAF)
+
+static void *glibc_malloc(void *pool, size_t size) {
```
igchor, line 22: FYI, we already have implementation of a pool based on malloc: https://github.com/oneapi-src/unified-memory-framework/blob/main/test/common/pool.hpp#L112

And here it is used to instantiate the umfPoolTest: https://github.com/oneapi-src/unified-memory-framework/blob/55f63546d558394be3b09493e472f95652e2c722/test/memoryPoolAPI.cpp#L119

igchor: So, as I suggested in another comment, I think you should just implement the tests from malloc_compliance_tests.*pp as `umfPoolTests` using TEST_P and it would already automatically run them on a malloc-based pool and on a disjoint_pool. You won't need this file at all.

test/malloc_compliance_tests_glibc.cpp
igchor: The `C++` approach is not that complicated. The idea is just to have a class that implements all the functions that are defined in 'ops' structure and that's it. You can then use a helper function to instantiate that pool/provider, like this `umf::poolMakeUnique<umf_test::malloc_pool>(...)`.

The benefit of this is that you avoid a lot of boilerplate and you can easily share the implementation. For example, see this test which creates a custom pool type that has some extra logic in `initialize` method but all other methods have a default behavior (thanks to inheritance): https://github.com/oneapi-src/unified-memory-framework/blob/main/test/memoryPoolAPI.cpp#L160

test/malloc_compliance_tests_glibc.cpp
ldorau: Done

test/malloc_compliance_tests_glibc.cpp
ldorau: Done

test/malloc_compliance_tests_glibc.cpp
ldorau: Thanks @igchor a lot!
Done,

ldorau:
Thanks @igchor a lot for your help and the PR: https://github.com/oneapi-src/unified-memory-framework/pull/72

test/malloc_compliance_tests.cpp
ldorau: Indeed. Done.

test/malloc_compliance_tests.cpp
ldorau: This list would have to have at least 100 (ITERATIONS) elements ... but ITERATIONS could be increased to 1000.
I have added `srand(SRAND_INIT_VALUE)` with constant `SRAND_INIT_VALUE` so it is almost a list of sizes now ;-)

test/malloc_compliance_tests.cpp
ldorau: Done

ldorau:
@bratpiorka @pbalcer please review again

test/malloc_compliance_tests.cpp
```diff
@@ -0,0 +1,149 @@
...
+    size_t alloc_ptr_size[ITERATIONS];
+
+    // Each allocation shall yield a pointer to an object disjoint from
+    // any other object.
+    for (int i = 0; i < ITERATIONS; i++) {
+        alloc_ptr_size[i] = rand_alloc_size(MAX_ALLOC_SIZE);
+        alloc_ptr[i] = umfPoolMalloc(hPool, alloc_ptr_size[i]);
+
+        // Supported under USE_WEAK_ALIGNMENT_RULES macro
+        // For compatibility, on 64-bit systems malloc should align to 16 bytes
```
ldorau, line 61: It fails on Windows !?!
https://github.com/oneapi-src/unified-memory-framework/actions/runs/7220175319/job/19672583680?pr=64
https://github.com/oneapi-src/unified-memory-framework/actions/runs/7220509785/job/19673783556

test/memoryPoolAPI.cpp
```diff
@@ -124,10 +124,11 @@ INSTANTIATE_TEST_SUITE_P(
     mallocPoolTest, umfPoolTest,
     ::testing::Values(poolCreateExtParams{&MALLOC_POOL_OPS, nullptr,
                                           &UMF_NULL_PROVIDER_OPS, nullptr},
+                      poolCreateExtParams{&MALLOC_POOL_OPS, nullptr,
```
igchor, line 127: I don't think this is needed - the MALLOC_POOL is implemented in a way such that it ignores the provider param and just calls malloc/free.

igchor (approved):

test/memoryPoolAPI.cpp
ldorau: Done. Removed.

test/common/test_helpers.c
```diff
@@ -12,6 +15,23 @@
 #include "umf/memory_pool.h"
 #include "umf/memory_provider.h"
 
+#include "test_helpers.h"
+
+// Check if the memory is zero filled
```
bratpiorka, line 20: Please fix the comment

test/common/test_helpers.h
```diff
@@ -14,6 +14,12 @@
 extern "C" {
 #endif
 
+int is_filled_with_char(void *ptr, size_t size, char c);
```
bratpiorka, line 17: could you use CamelCase style to match naming convention already used in this file? consider "bufferIsFilledWithChar" etc. Also we could keep the declaration of this group of functions together, without an extra empty line between them

test/memoryPool.hpp
```diff
@@ -15,6 +15,8 @@
 #ifndef UMF_TEST_MEMORY_POOL_OPS_HPP
 #define UMF_TEST_MEMORY_POOL_OPS_HPP
 
+#include "malloc_compliance_tests.hpp"
```
bratpiorka, line 18: Could you move all includes to #ifndef guards?

bratpiorka (changes requested):

test/memoryPool.hpp
ldorau: Done

test/common/test_helpers.h
ldorau: But it is a *.c file, so it should not use CamelCase, so I would prefer to change the rest to not use the CamelCase in a separate PR.

test/common/test_helpers.c
ldorau: Done

test/common/test_helpers.h
bratpiorka: ok lest unify this in the separate PR

bratpiorka (approved):

pbalcer (approved):

test/common/test_helpers.h
ldorau: Done: https://github.com/oneapi-src/unified-memory-framework/pull/76

PR #67, opened 2023-12-11
Unify names of structures and typedefs with UR

ldorau:


igchor:
Why are you suggesting to distinguish between structure names and other types? I actually prefer our current approach, it is also consistent with UR: https://github.com/oneapi-src/unified-runtime/blob/main/include/ur_api.h

ldorau:
> consistent with UR

OK, so let's make it consistent with UR - check the new version, please.

ldorau:
> Why are you suggesting to distinguish between structure names and other types? I actually prefer our current approach, it is also consistent with UR: https://github.com/oneapi-src/unified-runtime/blob/main/include/ur_api.h

OK, so let's make it consistent with UR - check the new version, please.

bratpiorka:
Hmm, where is the new version? Our approach to naming should be consistent with UR, unless we have strong arguments against it.

ldorau:
> Hmm, where is the new version? Our approach to naming should be consistent with UR, unless we have strong arguments against it.

Just here. It is like in UR now:
https://github.com/oneapi-src/unified-runtime/blob/main/include/ur_api.h
names of structures end with the `_t_` suffix and names of types end with the `_t` suffix, for example:
```c
///////////////////////////////////////////////////////////////////////////////
/// @brief Handle of platform's device object
typedef struct ur_device_handle_t_ *ur_device_handle_t;

///////////////////////////////////////////////////////////////////////////////
/// @brief Handle of context object
typedef struct ur_context_handle_t_ *ur_context_handle_t;

///////////////////////////////////////////////////////////////////////////////
/// @brief Handle of event object
typedef struct ur_event_handle_t_ *ur_event_handle_t;
```

bratpiorka:
but in the file you mention I see that they not always use "_":

typedef struct ur_device_partition_properties_t {
    ur_structure_type_t stype;                         ///< [in] type of this structure, must be
...
} ur_device_partition_properties_t;

So they use this extra "_" postfix only on the definition of handles. So it is ok to do this change in UMF:

typedef struct umf_memory_pool_handle_t_ *umf_memory_pool_handle_t;

But not this:

struct umf_memory_pool_ops_t_;

Note also that in our handles example the usual structure name "umf_memory_pool" does not contain the world "handle". Therefore, we do not need to add an extra "_". So there are two conventions: the first is to use the extra word "handle", and the second is to use "_". When I think about it, I like our approach better, because the structure itself cannot be a handle, only a pointer can work that way.

bratpiorka:
BTW, I think we should consider to always use typedef so that we don't have to add an extra enum/struct keyword when we use UMF entities in our code

ldorau:
> BTW, I think we should consider to always use typedef so that we don't have to add an extra enum/struct keyword when we use UMF entities in our code

@bratpiorka Done

bratpiorka (approved):

include/umf/memory_pool.h
```diff
@@ -19,8 +20,6 @@ extern "C" {
 
 typedef struct umf_memory_pool_t *umf_memory_pool_handle_t;
 
-struct umf_memory_pool_ops_t;
```
vinser52, line 22: why you replaced forward declaration of the `struct umf_memory_pool_ops_t;` with the `#include <umf/memory_pool_ops.h>`

vinser52 (approved):

include/umf/memory_pool.h
ldorau: By mistake. Reverted. Thanks.

PR #69, opened 2023-12-12
Add the umfPoolCreateEx internal function

kswiecicki:
... that creates a pool owning the provided memory provider.

### DIFF
--- a/src/memory_pool.c
+++ b/src/memory_pool.c
@@ -15,6 +15,37 @@
 #include <assert.h>
 #include <stdlib.h>
 
+umf_result_t umfPoolCreateEx(const umf_memory_pool_ops_t *pool_ops,
+                             void *pool_params,
+                             const umf_memory_provider_ops_t *provider_ops,
+                             void *provider_params,
+                             umf_memory_pool_handle_t *hPool) {
+    if (!pool_ops || !provider_ops || !hPool) {
+        return UMF_RESULT_ERROR_INVALID_ARGUMENT;
+    }
+
+    umf_memory_provider_handle_t provider = NULL;
+    umf_result_t ret =
+        umfMemoryProviderCreate(provider_ops, provider_params, &provider);
+    if (ret != UMF_RESULT_SUCCESS) {
+        return ret;
+    }
+    assert(provider != NULL);
+
+    umf_memory_pool_handle_t pool = NULL;
+    ret = umfPoolCreate(pool_ops, provider, pool_params, &pool);
+    if (ret != UMF_RESULT_SUCCESS) {
+        umfMemoryProviderDestroy(provider);
+        return ret;
+    }
+    assert(pool != NULL);
+
+    pool->own_provider = true;
+    *hPool = pool;
+
+    return UMF_RESULT_SUCCESS;
+}
+
 void *umfPoolMalloc(umf_memory_pool_handle_t hPool, size_t size) {
     return hPool->ops.malloc(hPool->pool_priv, size);
 }
--- a/src/memory_pool_default.c
+++ b/src/memory_pool_default.c
@@ -30,6 +30,7 @@ umf_result_t umfPoolCreate(const umf_memory_pool_ops_t *ops,
     assert(ops->version == UMF_VERSION_CURRENT);
 
     pool->provider = provider;
+    pool->own_provider = false;
 
     pool->ops = *ops;
     ret = ops->initialize(pool->provider, params, &pool->pool_priv);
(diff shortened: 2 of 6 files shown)


igchor:
FYI, this (or something similar) is needed if we want to implement `umfCreateMemoryPoolFromMemspace`.

vinser52:
I understand that this function is needed to implement `umfCreateMemoryPoolFromMemspace ` function, but it should be a public interface?

bratpiorka:
I would prefer a function that combines pool and provider params so the provider would be both created and destroyed with the pool. Something like:

umfPoolCreateEx(const struct umf_memory_pool_ops_t *pool_ops, void *pool_params, 
				const struct umf_memory_provider_ops_t *provider_ops, void *provider_params, 
				umf_memory_pool_handle_t *hPool);
				
with this API we don't expose the handle to the provider to the user at all so it is obvious who is managing and controlling the live of the provider.

bratpiorka:
Yes, I agree that we shouldn't expose this in public API if this is not needed. But also I think that we should follow the RAII concept - if the pool is an owner of the provider, the handle of provider should be hidden so no one could destroy the provider in the background

vinser52:
Will you add the unit test?

vinser52 (approved):

src/memory_pool.c
```diff
@@ -15,6 +15,30 @@
 #include <assert.h>
 #include <stdlib.h>
 
+enum umf_result_t
```
bratpiorka, line 18: Note that with recent Łukasz D. changes there is no need to use "enum" and "struct" keywords when you reference umf_result_t etc. Please remove them.

src/memory_pool.c
```diff
@@ -15,6 +15,30 @@
...
+umfPoolCreateEx(const struct umf_memory_pool_ops_t *pool_ops, void *pool_params,
+                const struct umf_memory_provider_ops_t *provider_ops,
+                void *provider_params, umf_memory_pool_handle_t *hPool) {
+    umf_memory_provider_handle_t provider = NULL;
```
bratpiorka, line 22: Check for pointers != null

bratpiorka (changes requested):

src/memory_pool.c
kswiecicki: I've added a null-check for the output handle pointer and asserts after provider/pool creation.
From what I've seen in `umfPoolCreate` it null-checks only the input provider handle, whereas `umfMemoryProviderCreate` doesn't check any parameters. Should we validate `...provider/pool_ops_t` pointers and input/output handle pointers in those functions?

kswiecicki:
> Will you add the unit test?

I've added test that checks whether `poolCreateEx` creates a valid pool internally and some basic successful pool creation, param fail tests.
@vinser52 @bratpiorka what do you think about adding AddressSanitizer or Valgrind for UMF testing? I could use it to check whether `poolCreateEx` appropriately destroyed the internal memory provider.

igchor:
> > Will you add the unit test?
> 
> I've added test that checks whether `poolCreateEx` creates a valid pool internally and some basic successful pool creation, param fail tests. @vinser52 @bratpiorka what do you think about adding AddressSanitizer or Valgrind for UMF testing? I could use it to check whether `poolCreateEx` appropriately destroyed the internal memory provider.

I think you could try enabling sanitizer, it should be pretty easy (just need to add flag to the compilation).

bratpiorka:
> > > Will you add the unit test?
> > 
> > 
> > I've added test that checks whether `poolCreateEx` creates a valid pool internally and some basic successful pool creation, param fail tests. @vinser52 @bratpiorka what do you think about adding AddressSanitizer or Valgrind for UMF testing? I could use it to check whether `poolCreateEx` appropriately destroyed the internal memory provider.
> 
> I think you could try enabling sanitizer, it should be pretty easy (just need to add flag to the compilation). And you can even reuse logic from UR: https://github.com/oneapi-src/unified-runtime/blob/4424195c7d5d990901f7343fa9e7ac500b03e38d/cmake/helpers.cmake#L45

We had a plans to add various sanitizers - look for them in our JIRA (UMFW-130)

src/memory_pool.c
bratpiorka: Please validate all input pointers. About other functions - we should add similar checks there, probably in separate PRs. Please note that we'll have to do it anyway when we enable coverity :)

bratpiorka (changes requested):

src/memory_pool.c
kswiecicki: Done, I've omitted the params, since pool/provider implementations not necessarily need a parameter for its initialization.

kswiecicki:
> > > > Will you add the unit test?
> > > 
> > > 
> > > I've added test that checks whether `poolCreateEx` creates a valid pool internally and some basic successful pool creation, param fail tests. @vinser52 @bratpiorka what do you think about adding AddressSanitizer or Valgrind for UMF testing? I could use it to check whether `poolCreateEx` appropriately destroyed the internal memory provider.
> > 
> > 
> > I think you could try enabling sanitizer, it should be pretty easy (just need to add flag to the compilation). And you can even reuse logic from UR: https://github.com/oneapi-src/unified-runtime/blob/4424195c7d5d990901f7343fa9e7ac500b03e38d/cmake/helpers.cmake#L45
> 
> We had plans to add various sanitizers - look for them in our JIRA (UMFW-130). If you want to take care of these please do it in separate PR.

I've assigned myself for this task.

bratpiorka (approved):

PR #72, opened 2023-12-14
Simplify test fixtures

igchor:
Expose ops for all pools/providers implemented in tests and change fixtures to accepts ops and params, instead of factory functions.

Also, remove no longer needed pool/providerMakeUnique functions from umf_helpers.hpp.


ldorau (approved):
LGTM !

PR #73, opened 2023-12-15
Extract Disjoint Pool into static lib

PatKamin:


CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ include(helpers)
 
 # Build Options
 option(UMF_BUILD_SHARED_LIBRARY "Build UMF as shared library" OFF)
+option(UMF_BUILD_LIBUMF_POOL_DISJOINT "Build the libumf_pool_disjoint static library" OFF)
```
bratpiorka, line 19: please update the README file

bratpiorka (changes requested):

src/pool/CMakeLists.txt
vinser52, line 1: As I understand, with this PR C++ is only required to build DisjointPool. The core UMF library is in pure C, right?
If so, could you please update the CMake file accordingly and change the build requirements in the Readme file?

igchor (approved):

src/pool/CMakeLists.txt
PatKamin: Not yet. There is a file `provider_tracking_windows.cpp` which is not a part of Disjoint Pool nor tests. For now we still have to require C++ for UMF build as this is one of the core UMF files (on Windows). Shall this file be rewritten in C in the future?

bratpiorka: @PatKamin yes, we should rewrite it if needed. I will add a task for this

src/pool/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
+# Copyright (C) 2023 Intel Corporation
+# Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# libumf_pool_disjoint
+if(UMF_BUILD_LIBUMF_POOL_DISJOINT)
+    add_umf_library(disjoint_pool STATIC
+        pool_disjoint.cpp)
+
+    add_library(${PROJECT_NAME}::disjoint_pool ALIAS disjoint_pool)
```
bratpiorka, line 10: why don't you use add_umf_library? Or better - update this function and use it also for pool_jemalloc

bratpiorka (changes requested):

src/pool/CMakeLists.txt
PatKamin: I don't use this custom wrapper of `add_library` as on aliased targets `target_include_directories` cannot be called. Beside that, note that `add_library(name ALIAS target)` is not the same function as ie. `add_library(name STATIC files)`. The former just creates an alias (another name) for an already existing CMake target whereas the latter creates a brand new CMake target.

I don't want to include aliasing into the `add_umf_library` function as well as we don't need all of our targets to have nice aliases with a namespace prefix, only those that are meant to be public and exposed to users when including the UMF repo with ie. FetchContent(). These I think should be limited to targets including headers and libraries containing the UMF API.

I believe that `pool_jemalloc` should be aliased, too. However, I save it for the more general UMF installation refactoring in another PR.

PatKamin:
Rebased

src/pool/CMakeLists.txt
bratpiorka: ok thanks for the clarification. Regarding  aliasing pool_jemalloc - I will add a task about that

bratpiorka (approved):

KFilipek (approved):
LGTM

src/pool/CMakeLists.txt
bratpiorka: @vinser52 if you think everything is ok could you please close this conversation?

.github/workflows/basic.yml
```diff
@@ -45,6 +45,7 @@ jobs:
           -DUMF_FORMAT_CODE_STYLE=OFF
           -DUMF_DEVELOPER_MODE=ON
           -DUMF_BUILD_LIBUMF_POOL_JEMALLOC=ON
+          -DUMF_BUILD_LIBUMF_POOL_DISJOINT=ON
           ${{matrix.pool_tracking}}
 
       - name: Build UMF
```
ldorau, line 51: Could you add (as the last one) the new step for testing `make install`?
`- name: Install UMF`

ldorau (changes requested):
Will you add the `uninstall` target later?

ldorau:
@bratpiorka  BTW: later there should be added a *.pc (pkgconfig) file for each library.

.github/workflows/basic.yml
bratpiorka: I think this should be a part of separate PR where we add some tests for the installation. There is a tasks for that in our JIRA [UMFW-176](https://jira.devtools.intel.com/browse/UMFW-176)

PatKamin:
> Will you add the `uninstall` target later?

This might be added in another PR.

.github/workflows/basic.yml
ldorau: OK

ldorau (approved):

ldorau:
@PatKamin Rebase please

PatKamin:
> @PatKamin Rebase please

Rebased

PatKamin:
> @bratpiorka BTW: later there should be added (and installed) a *.pc (pkgconfig) file for each library.

I'm not sure if we need them if we have CMake config file? I guess that serves the same purpose?

PR #74, opened 2023-12-18
Fix os_alloc() of OS memory provider to return aligned address

ldorau:
Fix `os_alloc()` of OS memory provider. Make it return aligned address. Replace `os_mmap()` with `os_mmap_aligned()` that returns aligned address.

### DIFF
--- a/src/provider/provider_os_memory_linux.c
+++ b/src/provider/provider_os_memory_linux.c
@@ -117,22 +117,48 @@ long os_get_mempolicy(int *mode, unsigned long *nodemask, unsigned long maxnode,
                          MPOL_F_NODE | MPOL_F_ADDR);
 }
 
-int os_mmap(void *addr, size_t length, int prot, int flags, int fd, long offset,
-            void **out_addr) {
-    if (out_addr == NULL) {
-        assert(0);
-        return -1;
+int os_mmap_aligned(void *hint_addr, size_t length, size_t alignment, int prot,
+                    int flags, int fd, long offset, void **out_addr) {
+    assert(out_addr);
+
+    size_t extended_length = length;
+    size_t page_size = os_get_page_size();
+
+    if (alignment > page_size) {
+        // We have to increase length by alignment to be able to "cut out"
+        // the correctly aligned part of the memory from the mapped region
+        // by unmapping the rest: unaligned beginning and unaligned end
+        // of this region.
+        extended_length += alignment;
     }
 
     // MAP_ANONYMOUS - the mapping is not backed by any file
-    void *mmap_addr =
-        mmap(addr, length, prot, MAP_ANONYMOUS | flags, fd, offset);
-    if (mmap_addr == MAP_FAILED) {
+    void *ptr = mmap(hint_addr, extended_length, prot, MAP_ANONYMOUS | flags,
+                     fd, offset);
+    if (ptr == MAP_FAILED) {
         return -1;
     }
 
-    *out_addr = mmap_addr;
+    if (alignment > page_size) {
+        uintptr_t addr = (uintptr_t)ptr;
+        uintptr_t aligned_addr = (addr + alignment) & ~(alignment - 1);
 
+        size_t head_len = aligned_addr - addr;
+        if (head_len > 0) {
+            munmap(ptr, head_len);
+        }
+
+        uintptr_t tail = aligned_addr + length;
+        size_t tail_len = (addr + extended_length) - (aligned_addr + length);
+        if (tail_len > 0) {
+            munmap((void *)tail, tail_len);
+        }
+
+        *out_addr = (void *)aligned_addr;
+        return 0;
+    }
+
+    *out_addr = ptr;
     return 0;
 }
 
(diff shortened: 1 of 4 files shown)


src/provider/provider_os_memory_linux.c
```diff
@@ -117,22 +117,39 @@ long os_get_mempolicy(int *mode, unsigned long *nodemask, unsigned long maxnode,
...
-            void **out_addr) {
-    if (out_addr == NULL) {
-        assert(0);
-        return -1;
-    }
+int os_mmap_aligned(void *hint_addr, size_t length, size_t alignment, int prot,
+                    int flags, int fd, long offset, void **out_addr) {
+    assert(out_addr);
+
+    size_t extended_length = length + alignment;
```
vinser52, line 124: if the passed `alignment` is less than the page size the `mmap` guarantees that the allocated range is page aligned. But according to the current logic `head_len` will be `0`, but `tail_len` will be equal to `alignment` and `munmap` will be called.
Should we optimize this case or it's not worth it?

ldorau: Fixed.
Done.

pbalcer (approved):

src/provider/provider_os_memory_linux.c
```diff
@@ -117,22 +117,44 @@ long os_get_mempolicy(int *mode, unsigned long *nodemask, unsigned long maxnode,
...
     }
 
-    *out_addr = mmap_addr;
+    if (alignment > page_size) {
+        uintptr_t addr = (uintptr_t)ptr;
+        uintptr_t aligned_addr = (addr + alignment) & ~(alignment - 1);
+
+        size_t head_len = aligned_addr - addr;
+        if (head_len > 0) {
+            munmap(ptr, head_len);
```
bratpiorka, line 144: shouldn't we do this when head_len > page_size? Please check for the unmap error

src/provider/provider_os_memory_linux.c
```diff
@@ -117,22 +117,44 @@ long os_get_mempolicy(int *mode, unsigned long *nodemask, unsigned long maxnode,
...
+        }
+
+        uintptr_t tail = aligned_addr + length;
+        size_t tail_len = (addr + extended_length) - (aligned_addr + length);
+        if (tail_len > 0) {
+            munmap((void *)tail, tail_len);
```
bratpiorka, line 150: why do we have this code? Will mmap allocate more than needed? If so, can't we just provide right value for "extended_lenght" ?

src/provider/provider_os_memory_linux.c
ldorau: Is it OK now?

test/provider_os_memory.cpp
```diff
@@ -91,7 +91,7 @@ TEST_F(test, provider_os_memory_create_MBIND_FAILED) {
     ASSERT_EQ(os_memory_provider, nullptr);
 }
 
-TEST_F(test, provider_os_memory_alloc_free) {
+TEST_F(test, provider_os_memory_alloc_free_4k_alignment_0_SUCCESS) {
```
bratpiorka, line 94: I think test success is the default case, so we don't need to add this information. However, if we decide to do so, we should rename all the tests in our repo.

bratpiorka (changes requested):

test/provider_os_memory.cpp
ldorau: Done

src/provider/provider_os_memory_linux.c
vinser52: I think the main issue is that we do not know the initial alignment we received from mmap. So we allocate a little bit more than needed to be able to align manually.

vinser52 (approved):

bratpiorka (approved):

src/provider/provider_os_memory_linux.c
```diff
@@ -117,22 +117,48 @@ long os_get_mempolicy(int *mode, unsigned long *nodemask, unsigned long maxnode,
...
     }
 
-    *out_addr = mmap_addr;
+    if (alignment > page_size) {
+        uintptr_t addr = (uintptr_t)ptr;
+        uintptr_t aligned_addr = (addr + alignment) & ~(alignment - 1);
+
+        size_t head_len = aligned_addr - addr;
+        if (head_len > 0) {
+            munmap(ptr, head_len);
```
igchor, line 148: Just a question, from what I understand unmap unmaps whole pages (i.e. even if len is == 1, it will unmap the whole 4k page).   Will this code not unmap the memory that we actually want to use?

ldorau: This code will not unmap the memory that we actually want to use for sure. `ptr` and `head_len` are always correct.
The only problem is that the `tail` address can be not page-aligned and therefore the second `munmap()` call can fail.
I have submitted a fix for this issue: https://github.com/oneapi-src/unified-memory-framework/pull/84
Thanks @igchor !

PR #75, opened 2023-12-18
Add test for jemalloc pool manager

ldorau:
Requires:
- [x] #64
- [x] #74


pbalcer (approved):

test/memoryPoolAPI.cpp
vinser52, line 1: As I understand we will execute the general pool manager tests we have so far for all other pool managers.
Should we test that jemalloc-specific API is implemented in the right way? I mean the case with split/merge arena extents.

KFilipek (approved):

test/memoryPoolAPI.cpp
ldorau: > As I understand we will execute the general pool manager tests we have so far for all other pool managers. 
>
Yes.

> Should we test that jemalloc-specific API is implemented in the right way? I mean the case with split/merge arena extents.
>
I think we should not test directly the internal jemalloc hooks. They are called many times (also the `arena_extent_split/merge` hooks) when we call the public pool's API, so they are already tested indirectly.

igchor (approved):

test/memoryPoolAPI.cpp
ldorau: If we implement splitting/merging extents, we can inverse the return value. Now it is not implemented, so we have to return a failure according to the reality ;-)

ldorau: > So, in other words: if we inverse the return value of the `arena_extent_split/merge` functions the existing tests will catch that?

The existing tests already call those extents, so they can fail if you inverse the return value without implementing them.

vinser52 (approved):

Issue #63, opened 2023-12-11
Disjoint pool with OS memory provider floods screen with error messages


ldorau:
Disjoint pool with OS memory provider floods screen with huge amount of the following error messages:
```
DisjointPool: error from memory provider: 2
Native error msg: memory deallocation failed: Invalid argument, native error code: 1005
```
because each `memoryProviderFree()` calls `umfMemoryProviderFree(os_memory_provider, ptr, 0);`  that calls `munmap(ptr, 0);` and generates such error, because `size` equals `0`.

Steps to reproduce - just run benchmark:
```
[build]$ ./benchmark/ubench
```

ldorau:
@bratpiorka

bratpiorka:
@ldorau could you add instructions on how to reproduce this problem?

bratpiorka:
@ldorau could you add instructions on how to reproduce this problem?

ldorau:
> @ldorau could you add instructions on how to reproduce this problem?

Just run:
```
[build]$ ./benchmark/ubench
```

PR #80, opened 2023-12-19
Fix: print errors to stderr instead of stdout

ldorau:
Ref: #63


bratpiorka:
OK but please note that in the future we'll switch to a logging framework based on UR code

bratpiorka (approved):

ldorau:
> OK but please note that in the future we'll switch to a logging framework based on UR code

Of course

PR #84, opened 2023-12-20
Make sure pointers are correctly aligned in os_mmap_aligned()

ldorau:
Ref: https://github.com/oneapi-src/unified-memory-framework/pull/74#discussion_r1432442666

### DIFF
--- a/src/provider/provider_os_memory_linux.c
+++ b/src/provider/provider_os_memory_linux.c
@@ -117,12 +117,18 @@ long os_get_mempolicy(int *mode, unsigned long *nodemask, unsigned long maxnode,
                          MPOL_F_NODE | MPOL_F_ADDR);
 }
 
-int os_mmap_aligned(void *hint_addr, size_t length, size_t alignment, int prot,
-                    int flags, int fd, long offset, void **out_addr) {
+static inline void assert_is_page_aligned(uintptr_t ptr, size_t page_size) {
+    assert((ptr & (page_size - 1)) == 0);
+    (void)ptr;       // unused in Release build
+    (void)page_size; // unused in Release build
+}
+
+int os_mmap_aligned(void *hint_addr, size_t length, size_t alignment,
+                    size_t page_size, int prot, int flags, int fd, long offset,
+                    void **out_addr) {
     assert(out_addr);
 
     size_t extended_length = length;
-    size_t page_size = os_get_page_size();
 
     if (alignment > page_size) {
         // We have to increase length by alignment to be able to "cut out"
@@ -141,15 +147,30 @@ int os_mmap_aligned(void *hint_addr, size_t length, size_t alignment, int prot,
 
     if (alignment > page_size) {
         uintptr_t addr = (uintptr_t)ptr;
-        uintptr_t aligned_addr = (addr + alignment) & ~(alignment - 1);
+        uintptr_t aligned_addr = addr;
+        uintptr_t rest_of_div = aligned_addr % alignment;
+
+        if (rest_of_div) {
+            aligned_addr += alignment - rest_of_div;
+        }
+
+        assert_is_page_aligned(aligned_addr, page_size);
 
         size_t head_len = aligned_addr - addr;
         if (head_len > 0) {
             munmap(ptr, head_len);
         }
 
+        // tail address has to page-aligned
         uintptr_t tail = aligned_addr + length;
-        size_t tail_len = (addr + extended_length) - (aligned_addr + length);
+        if (tail & (page_size - 1)) {
+            tail = (tail + page_size) & ~(page_size - 1);
+        }
+
+        assert_is_page_aligned(tail, page_size);
+        assert(tail >= aligned_addr + length);
+
+        size_t tail_len = (addr + extended_length) - tail;
         if (tail_len > 0) {
             munmap((void *)tail, tail_len);
         }
--- a/test/provider_os_memory.cpp
+++ b/test/provider_os_memory.cpp
@@ -12,6 +12,9 @@ using umf_test::test;
 
 #define SIZE_4K (4096)
 
+#define ASSERT_IS_ALIGNED(ptr, alignment)                                      \
+    ASSERT_EQ(((uintptr_t)ptr % alignment), 0)
+
 static umf_os_memory_provider_params_t UMF_OS_MEMORY_PROVIDER_PARAMS_TEST = {
     /* .protection = */ UMF_PROTECTION_READ | UMF_PROTECTION_WRITE,
     /* .visibility = */ UMF_VISIBILITY_PRIVATE,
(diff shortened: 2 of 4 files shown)


test/provider_os_memory.cpp
```diff
@@ -147,6 +147,34 @@ TEST_F(test, provider_os_memory_alloc_free_2k_alignment_2k) {
...
+                                         &os_memory_provider);
+    ASSERT_EQ(umf_result, UMF_RESULT_SUCCESS);
+    ASSERT_NE(os_memory_provider, nullptr);
+
+    void *ptr = nullptr;
+    size_t size = SIZE_4K + 4;
+    size_t alignment = 4 * SIZE_4K;
+    umf_result =
+        umfMemoryProviderAlloc(os_memory_provider, size, alignment, &ptr);
+    ASSERT_EQ(umf_result, UMF_RESULT_SUCCESS);
```
bratpiorka, line 167: please check here if the ptr is aligned correctly

ldorau: Done

ldorau:
@igchor

bratpiorka (approved):

src/provider/provider_os_memory_linux.c
```diff
@@ -142,14 +149,25 @@ int os_mmap_aligned(void *hint_addr, size_t length, size_t alignment, int prot,
     if (alignment > page_size) {
         uintptr_t addr = (uintptr_t)ptr;
         uintptr_t aligned_addr = (addr + alignment) & ~(alignment - 1);
+        assert_is_page_aligned(aligned_addr, page_size);
 
         size_t head_len = aligned_addr - addr;
         if (head_len > 0) {
+            // head_len has to be page-aligned, because aligned_addr has to be page-aligned
+            assert_is_page_aligned(head_len, page_size);
```
vinser52, line 157: Why do we just crash? Assert indicates a mistake in the code logic. Is this condition only possible by mistake? what if the page size 4kb and the requested alignment 6kb?

src/provider/provider_os_memory_linux.c
```diff
@@ -125,6 +131,7 @@ int os_mmap_aligned(void *hint_addr, size_t length, size_t alignment, int prot,
     size_t page_size = os_get_page_size();
 
     if (alignment > page_size) {
+        assert_is_page_aligned(alignment, page_size);
```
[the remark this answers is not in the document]
igchor: I think we should also update documentation of `umfMemoryProviderAlloc`. 

We should either require `alignment` to always be a multiple of what `umfMemoryProviderGetMinPageSize()` returns (and return INVALID_ARGUMENT otherwise) or change the meaning of `alignment` param to `minAlignment` (and for example allow providers to return 8k aligned memory when 6k alignment is requested). @vinser52 any preferences?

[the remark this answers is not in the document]
ldorau: IMHO the first option is better:
"We should require `alignment` to always be a multiple of what `umfMemoryProviderGetMinPageSize()` returns (and return `INVALID_ARGUMENT` otherwise) "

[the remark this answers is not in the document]
vinser52: I prefer the first one since it is more clear. And it is consistent with mmap behavior.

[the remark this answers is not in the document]
ldorau: I think that the `alignment` should be a multiple of the minimum page size (`umfMemoryProviderGetMinPageSize()`) **or a divider of that**. If it is a divider, the address will be aligned to the minimum page size (for example: alignment 1024 is also a correct alignment, because an address aligned to 4096 is also aligned to 1024).

ldorau:
Requires: [ ] #91

src/provider/provider_os_memory_linux.c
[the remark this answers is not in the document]
ldorau: `alignment` is verified in `os_alloc()` now, 
see: 
- https://github.com/oneapi-src/unified-memory-framework/pull/91
- https://github.com/oneapi-src/unified-memory-framework/pull/90

Done.

src/provider/provider_os_memory_linux.c
ldorau: Code changed.
Done.

ldorau:
Rebased

src/provider/provider_os_memory_linux.c
ldorau: BTW tests added in a separate PR: https://github.com/oneapi-src/unified-memory-framework/pull/93

src/provider/provider_os_memory_linux.c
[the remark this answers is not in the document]
igchor: > I think that the alignment should be a multiple of the minimum page size (umfMemoryProviderGetMinPageSize()) or a divider of that. If it is a divider, the address will be aligned to the minimum page size (for example: alignment 1024 is also a correct alignment, because an address aligned to 4096 is also aligned to 1024).

Ok, but that's a bit inconsistent IMO. Why is 1024 a correct alignment and 6k is not? I could argue that since an address aligned to 8k (a multiple of page size) is also aligned to 6k it should be OK as well. A generic code that wants to wrap our allocate would have to have a logic like this (or just always round up to page size): 

```
assert(is_pow2(alignment);
if (alignment < `umfMemoryProviderGetMinPageSize()`) return umfMemoryProviderAlloc(..., alignment);
else return umfMemoryProviderAlloc(..., round_up_to_page(alignment));
```

I would prefer either option no 1 (both 1k and 6k is disallowed) or option 2 (both are fine).

src/provider/provider_os_memory_linux.c
igchor: I still only see asserts and not the INVALID_ARGUMENT error code.

src/provider/provider_os_memory_linux.c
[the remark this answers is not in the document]
ldorau: - 6k is not a power of 2 and you have `assert(is_pow2(alignment);` here, so it would assert in case of 6k :-)
- all dividers of 4k (for example 1024) are powers of 2

src/provider/provider_os_memory_linux.c
ldorau: Assert indicates a mistake in the code logic. Those conditions are possible only by mistake.

ldorau: I meant those asserts verify if `os_alloc()` passed a correct alignment. A user cannot trigger them changing the input alignment. All required checks are done in `os_alloc()` and `os_alloc()` passes only a correct alignment to this function. So IMHO those asserts are correct.

igchor: Ah yes, sorry, I missed the PR that changed the checks in os_alloc.

src/provider/provider_os_memory_linux.c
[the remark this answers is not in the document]
igchor: Right, the assert in that example would fail ;d

Right, being a power of 2 and 'being a divider or a multiple of the page size' is always equivalent so from that perspective I think the logic you proposed makes sense.

igchor (approved):

PR #86, opened 2023-12-20
Fix the UMF_BUILD_SHARED_LIBRARY=ON case

ldorau:


benchmark/CMakeLists.txt
```diff
@@ -12,4 +12,5 @@ target_include_directories(ubench PRIVATE ${UMF_CMAKE_SOURCE_DIR}/include/)
 target_link_libraries(ubench
 	unified_memory_framework
 	numa
```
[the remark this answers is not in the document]
bratpiorka: @vinser it is interesting - can you elaborate more on your statement?

test/CMakeLists.txt
```diff
@@ -58,16 +58,20 @@ endfunction()
 
 add_subdirectory(common)
 
+if (UMF_BUILD_SHARED_LIBRARY)
+    set(LIBS_UMF_SHARED_LIBRARY numa)
```
bratpiorka, line 62: I don't understand this logic - how the fact that umf is a shared lib is connected with adding numa to this variable. Event if this fixes the problem this should not be done this way.

bratpiorka (changes requested):

benchmark/CMakeLists.txt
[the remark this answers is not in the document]
vinser52: I meant the following:

- if we build UMF as a static library then all libumf dependencies should be linked with the target binary that links with UMF, right?
- if we build UMF as a dynamic library then libumf is linked with libnuma.

My question is does ubench depend on libnuma or we are linking libnuma because libumf has a dependency on it?

test/CMakeLists.txt
```diff
@@ -58,16 +58,20 @@ endfunction()
...
     set(LIBS_JEMALLOC pool_jemalloc jemalloc numa)
 endif()
 
 add_umf_test(NAME base SRCS base.cpp)
-add_umf_test(NAME memoryPool SRCS memoryPoolAPI.cpp malloc_compliance_tests.cpp LIBS ${LIBS_JEMALLOC})
-add_umf_test(NAME memoryProvider SRCS memoryProviderAPI.cpp)
-add_umf_test(NAME disjointPool SRCS disjoint_pool.cpp malloc_compliance_tests.cpp)
-add_umf_test(NAME c_api_disjoint_pool SRCS c_api/disjoint_pool.c)
-add_umf_test(NAME memory_pool_internal SRCS memory_pool_internal.cpp)
+add_umf_test(NAME memoryPool SRCS memoryPoolAPI.cpp malloc_compliance_tests.cpp LIBS ${LIBS_UMF_SHARED_LIBRARY} ${LIBS_JEMALLOC})
```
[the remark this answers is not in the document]
ldorau: `${LIBS_UMF_SHARED_LIBRARY}` is just `numa` now as you can see.
When `libumf` is a static library, `libnuma` is linked with it statically.
When `libumf` is a dynamic library, `libnuma` has to be linked dynamicly with all its targets (tests in this case)

test/CMakeLists.txt
ldorau: `${LIBS_UMF_SHARED_LIBRARY}` is just `numa` now as you can see.
When `libumf` is a static library, `libnuma` is linked with it statically.
When `libumf` is a dynamic library, `libnuma` has to be linked dynamicly with all its targets (tests in this case).

So how should it be done?

vinser52: imagine UMF depends on libA, libB, libC.

If UMF is built as a static library, all dependencies of libumf should be linked with targets that uses libumf. E.g. if ubench uses UMF then ubench should be linked with libA, libB, libC
if UMF is built as a dynamic library, all dependencies are linked directly with libumf.so. E.g. ubench should only links with libumf.

Regarding how it should be implemented with CMake, I am not sure yet.

benchmark/CMakeLists.txt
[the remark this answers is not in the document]
ldorau: Yes, it should.

test/CMakeLists.txt
ldorau: @vinser52 It is seems to be the opposite: libumf is a static library by default now and tests do not need to be linked with libnuma. When you change libumf to a dynamic library the tests need to be linked with libnuma ...

vinser52: > When you change libumf to a dynamic library the tests need to be linked with libnuma

Why? Who does actually need the libnuma dependency? libumf uses libnuma, right?

ldorau: You are right. I have not cleaned up the previous build and everything got mixed ...
It is fixed now. Please review again.

benchmark/CMakeLists.txt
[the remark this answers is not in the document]
ldorau: We are linking with libnuma because libumf has a dependency on it.
Fixed.
Done.

src/CMakeLists.txt
```diff
@@ -42,6 +42,7 @@ endif()
 if(UMF_BUILD_SHARED_LIBRARY)
     add_umf_library(unified_memory_framework SHARED ${UMF_SOURCES})
     target_compile_definitions(unified_memory_framework PUBLIC UMF_SHARED_LIBRARY)
+    target_link_libraries(unified_memory_framework numa)
```
vinser52, line 45: What if libbumf will have more dependencies?

I think instead of hardcoding particular dependencies in two places we need to link with `LIBS_UMF_STATIC`.
And `LIBS_UMF_STATIC` is a bad name in that case - I would rename it to `LIBS_UMF_DEPENDENCIES` or something like that.

So we should handle the single variable that capture UMF dependencies and

- if UMF is static lib link ubench directly with that dependencies
- if UMF is dynamic lib link UMF itself with the dependencies

igchor: I still don't quite understand why we need to link those dependencies to tests. Do we use any libnuma APIs directly in tests? Otherwise, we should not need to link with it.

benchmark/CMakeLists.txt
```diff
@@ -11,5 +11,5 @@ add_dependencies(ubench unified_memory_framework)
 target_include_directories(ubench PRIVATE ${UMF_CMAKE_SOURCE_DIR}/include/)
 target_link_libraries(ubench
 	unified_memory_framework
-	numa
-	pthread)
+	${LIBS_UMF_STATIC}
+	m)
```
igchor, line 15: `m`?

test/CMakeLists.txt
igchor: Neither tests nor benchmarks should link with UMF dependencies (regardless of whether UMF is built as static or shared library) unless they use the libraries themselves. See: https://github.com/oneapi-src/unified-memory-framework/pull/89

src/CMakeLists.txt
igchor: See: https://github.com/oneapi-src/unified-memory-framework/pull/89

igchor:
Ok, it seems we haven't linked UMF with the required libraries at all. We just linked tests/benchmarks, here's the fix for this: https://github.com/oneapi-src/unified-memory-framework/pull/89 We don't need to link neither tests nor benchmarks with libnuma.

src/CMakeLists.txt
vinser52: Oh, good point. I did not know that static library can somehow contain information about its dependencies. This post explain that: https://stackoverflow.com/questions/43176681/do-statically-linked-libraries-require-linking-with-their-dependencies

Thank you, Igor!

ldorau:
Thanks, @igchor !

ldorau:
Fixed by https://github.com/oneapi-src/unified-memory-framework/pull/89
Thanks, @igchor !

PR #88, opened 2023-12-20
IPC API

vinser52:
This PR introduces IPC API to UMF. The main use cases so far are from the GPU domain (L0, CUDA have such capabilities). The API is pretty similar to what L0 and CUDA have, but it also could be applied for shared memory or file-backed memory.

This PR is big enough, but still it does not contain everything we need.

The remaining tasks/items are the following:
- Implement more efficient caching for IPC handles. This PR implements caching of IPC handles on the producer side (on getting IPC handles). Based on `critnib` data structure (in fact just a map, not a real cache with eviction policy).
- Need to implement proper cache with eviction policies (e.g. LRU) so that we can control the size of the cache.
- Implement cache on the consumer side (when we open IPC handles), in this PR we always call memory provider.
- Evaluate `uthash` data structure that supports arbitrary key type and is more suitable for caching on a consumer side where the key is remote PID + address
- Create documentation.
- Create more tests.
- Extend UMF IPC API to allow retrieving native IPC handle from UMF IPC handle.

### DIFF
--- a/src/provider/provider_level_zero.c
+++ b/src/provider/provider_level_zero.c
@@ -8,6 +8,7 @@
 #include <assert.h>
 #include <stdbool.h>
 #include <stddef.h>
+#include <string.h>
 
 // Level Zero API
 #include <ze_api.h>
@@ -17,6 +18,7 @@
 #include <umf/providers/provider_level_zero.h>
 
 #include "base_alloc_global.h"
+#include "utils_assert.h"
 #include "utils_common.h"
 #include "utils_concurrency.h"
 #include "utils_load_library.h"
@@ -41,6 +43,13 @@ typedef struct ze_ops_t {
                                     const ze_host_mem_alloc_desc_t *, size_t,
                                     size_t, ze_device_handle_t, void *);
     ze_result_t (*zeMemFree)(ze_context_handle_t, void *);
+    ze_result_t (*zeMemGetIpcHandle)(ze_context_handle_t, const void *,
+                                     ze_ipc_mem_handle_t *);
+    ze_result_t (*zeMemPutIpcHandle)(ze_context_handle_t, ze_ipc_mem_handle_t);
+    ze_result_t (*zeMemOpenIpcHandle)(ze_context_handle_t, ze_device_handle_t,
+                                      ze_ipc_mem_handle_t,
+                                      ze_ipc_memory_flags_t, void **);
+    ze_result_t (*zeMemCloseIpcHandle)(ze_context_handle_t, void *);
 } ze_ops_t;
 
 static ze_ops_t g_ze_ops;
@@ -57,9 +66,21 @@ static void init_ze_global_state(void) {
     *(void **)&g_ze_ops.zeMemAllocShared =
         util_get_symbol_addr(0, "zeMemAllocShared");
     *(void **)&g_ze_ops.zeMemFree = util_get_symbol_addr(0, "zeMemFree");
+    *(void **)&g_ze_ops.zeMemGetIpcHandle =
+        util_get_symbol_addr(0, "zeMemGetIpcHandle");
+    *(void **)&g_ze_ops.zeMemPutIpcHandle =
+        util_get_symbol_addr(0, "zeMemPutIpcHandle");
+    *(void **)&g_ze_ops.zeMemOpenIpcHandle =
+        util_get_symbol_addr(0, "zeMemOpenIpcHandle");
+    *(void **)&g_ze_ops.zeMemCloseIpcHandle =
+        util_get_symbol_addr(0, "zeMemCloseIpcHandle");
 
     if (!g_ze_ops.zeMemAllocHost || !g_ze_ops.zeMemAllocDevice ||
-        !g_ze_ops.zeMemAllocShared || !g_ze_ops.zeMemFree) {
+        !g_ze_ops.zeMemAllocShared || !g_ze_ops.zeMemFree ||
+        !g_ze_ops.zeMemGetIpcHandle || !g_ze_ops.zeMemOpenIpcHandle ||
+        !g_ze_ops.zeMemCloseIpcHandle) {
+        // g_ze_ops.zeMemPutIpcHandle can be NULL because it was introduced
+        // starting from Level Zero 1.6
         LOG_ERR("Required Level Zero symbols not found.");
         Init_ze_global_state_failed = true;
     }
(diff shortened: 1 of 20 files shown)


src/provider/provider_tracking.c
```diff
@@ -216,6 +260,116 @@ static const char *trackingName(void *provider) {
...
+                //    But this case should be rare enough.
+                // 2. critnib failed to allocate memory internally. We need
+                //    to cleanup and return corresponding error.
+                free(cache_value);
+                ret = umfMemoryProviderPutIPCHandle(p->hUpstream, ipcData);
+                if (ret != UMF_RESULT_SUCCESS) {
+                    return ret;
+                }
+                if (insRes == ENOMEM) {
+                    return UMF_RESULT_ERROR_OUT_OF_HOST_MEMORY;
```
igchor, line 321: Can't we just skip adding to the cache in that case?

src/provider/provider_tracking.c
```diff
@@ -216,6 +260,116 @@ static const char *trackingName(void *provider) {
...
+            cache_value->size = ipcDataSize;
+            memcpy(cache_value->ipcData, ipcData, ipcDataSize);
+
+            int insRes = critnib_insert(p->ipcCache, (uintptr_t)ptr,
+                                        (void *)cache_value, 0 /*update*/);
+            if (insRes == 0) {
+                cached = 1;
+            } else {
+                // critnib_insert might fail in 2 cases:
+                // 1. Another thread created cache entry. So we need to
```
igchor, line 307: Why can't we just return the `ipcData` we have even if another thread inserted into the cache? We will then have two different IPC handles to the same memory but is it a problem (for correctness or performance)?

src/provider/provider_tracking.c
```diff
@@ -216,6 +260,116 @@ static const char *trackingName(void *provider) {
...
+
+    return ret;
+}
+
+static enum umf_result_t trackingPutIpcHandle(void *provider, void *ipcData) {
+    (void)provider;
+    (void)ipcData;
+    // We just keep ipcData in the provider->ipcCache.
+    // The actual Put is called inside trackingFree
+    return UMF_RESULT_SUCCESS;
```
igchor, line 335: If we never remove from the cache, perhaps it would make sense to have some limit on the size of the cache?

src/provider/provider_tracking.c
```diff
@@ -216,6 +260,116 @@ static const char *trackingName(void *provider) {
...
+}
+
+static enum umf_result_t trackingOpenIpcHandle(void *provider, void *ipcData,
+                                               void **ptr) {
+    umf_tracking_memory_provider_t *p =
+        (umf_tracking_memory_provider_t *)provider;
+    enum umf_result_t ret = UMF_RESULT_SUCCESS;
+
+    // TODO: This is hack to get size of memory pointed by IPC handle.
```
igchor, line 344: I think this is fine, you can remove the 'TODO' mark. I would just suggest to move this logic and the comment into a function (`getIPCHandleSizeFromData` or something like this).

test/ipcAPI.cpp
```diff
@@ -0,0 +1,336 @@
...
+struct umfIpcTest : umf_test::test {
+    umfIpcTest() : pool(nullptr, nullptr) {}
+    void SetUp() override {
+        test::SetUp();
+        this->pool = makePool();
+    }
+
+    void TearDown() override { test::TearDown(); }
+
+    umf::pool_unique_handle_t makePool() {
```
igchor, line 126: You can use this: https://github.com/oneapi-src/unified-memory-framework/blob/main/test/memoryPool.hpp#L23

Perhaps we should move that function to some place where it's more visible.

test/ipcAPI.cpp
```diff
@@ -0,0 +1,336 @@
...
+
+    ret = umfPutIPCHandle(ipcHandleHalf);
+    EXPECT_EQ(ret, UMF_RESULT_SUCCESS);
+
+    ret = umfPoolFree(pool.get(), ptr);
+    EXPECT_EQ(ret, UMF_RESULT_SUCCESS);
+
+    EXPECT_EQ(stat.getCount, 1);
+    EXPECT_EQ(stat.putCount, stat.getCount);
+    // EXPECT_EQ(stat.openCount, 1);
```
igchor, line 206: Why is this commented out?

test/ipcAPI.cpp
```diff
@@ -0,0 +1,336 @@
...
+    EXPECT_EQ(stat.closeCount, stat.openCount);
+}
+
+TEST_F(umfIpcTest, ConcurrentGetPutHandles) {
```
igchor, line 210: nit: it would be good to have a multithreaded tests with a provider that does not use any locks internally. With provider_mock_ipc you introduced extra synchronization which can hide some potential issues with critnib I think.

src/provider/provider_tracking.c
vinser52: Current implementation of put function does nothing today. The implementation rely on ipcCache and we close ipc handle when free method is called.
I considered the same idea like you suggested but it requires put method to be thread-safe and check whether ipcData is in the cache or not.

src/provider/provider_tracking.c
vinser52: No with the current logic. See my previous comment about put method. Current logic asumes that we add to the ipc cache once and all clients reuse it.

src/provider/provider_tracking.c
vinser52: Yeah, I considered it. We need real cache implementation with LRU or LFU policy instead of just map. We can do it later if we will see that real use cases require this

test/ipcAPI.cpp
vinser52: Because cache for open IPC handle is not implemented yet.

test/ipcAPI.cpp
vinser52: Agree. I will think about that.

src/provider/provider_tracking.c
igchor: Oh, right, because we might have a leak otherwise.

src/provider/provider_tracking.c
vinser52: Done

src/provider/provider_tracking.c
```diff
@@ -429,6 +466,132 @@ static const char *trackingName(void *provider) {
...
+static umf_result_t trackingCloseIpcHandle(void *provider, void *ptr) {
+    umf_tracking_memory_provider_t *p =
+        (umf_tracking_memory_provider_t *)provider;
+
+    // umfMemoryTrackerRemove should be called before umfMemoryProviderFree
+    // to avoid a race condition. If the order would be different, other thread
+    // could allocate the memory at address `ptr` before a call to umfMemoryTrackerRemove
+    // resulting in inconsistent state.
+    if (ptr) {
+        // TODO: below method requires size but does not use it internally
```
bratpiorka, line 583: please either keep it w/o TODOs or remove it :) btw we could always add it later. What about CUDA? Do they use it?

test/ipcAPI.cpp
```diff
@@ -0,0 +1,342 @@
+// Copyright (C) 2023 Intel Corporation
+// Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// This file contains tests for UMF pool API
+
+//#include "memoryPool.hpp"
```
bratpiorka, line 6: please remove if not needed

test/ipcAPI.cpp
bratpiorka: Could you add a TODO before this line?

src/provider/provider_tracking.c
```diff
@@ -314,6 +342,14 @@ static umf_result_t trackingFree(void *hProvider, void *ptr, size_t size) {
         }
     }
 
+    void *value = critnib_remove(p->ipcCache, (uintptr_t)ptr);
+    if (value) {
+        ipc_cache_value_t *cache_value = (ipc_cache_value_t *)value;
+        ret = umfMemoryProviderPutIPCHandle(p->hUpstream, cache_value->ipcData);
+        assert(ret == UMF_RESULT_SUCCESS);
```
igchor, line 349: This should be a regular check imo, not an assert.

src/provider/provider_tracking.c
```diff
@@ -429,6 +466,132 @@ static const char *trackingName(void *provider) {
...
+            cached = 1;
+        } else {
+            ret =
+                umfMemoryProviderGetIPCHandle(p->hUpstream, ptr, size, ipcData);
+            if (ret != UMF_RESULT_SUCCESS) {
+                return ret;
+            }
+
+            ret = umfMemoryProviderGetIPCHandleSize(p->hUpstream, &ipcDataSize);
+            assert(ret == UMF_RESULT_SUCCESS);
```
igchor, line 496: Why not a regular check?

src/provider/provider_tracking.c
```diff
@@ -429,6 +466,132 @@ static const char *trackingName(void *provider) {
...
+
+            size_t value_size = sizeof(ipc_cache_value_t) + ipcDataSize;
+            ipc_cache_value_t *cache_value =
```
igchor, line 499: Add check for NULL ptr

src/provider/provider_tracking.c
```diff
@@ -429,6 +466,132 @@ static const char *trackingName(void *provider) {
...
+static umf_result_t trackingGetIpcHandle(void *provider, const void *ptr,
+                                         size_t size, void *ipcData) {
+    umf_tracking_memory_provider_t *p =
+        (umf_tracking_memory_provider_t *)provider;
+    umf_result_t ret = UMF_RESULT_SUCCESS;
+    size_t ipcDataSize = 0;
+    int cached = 0;
+    do {
+        void *value = critnib_get(p->ipcCache, (uintptr_t)ptr);
+        if (value) { //cache hit
```
igchor, line 484: Should we validate the size of the allocation stored in the cache here?

test/ipcAPI.cpp
```diff
@@ -0,0 +1,342 @@
...
+        auto ret =
+            umfMemoryProviderCreate(&IPC_MOCK_PROVIDER_OPS, &stat, &hProvider);
+        EXPECT_EQ(ret, UMF_RESULT_SUCCESS);
+
+        ret = umfPoolCreate(umfProxyPoolOps(), hProvider, nullptr,
+                            UMF_POOL_CREATE_FLAG_OWN_PROVIDER, &hPool);
+        EXPECT_EQ(ret, UMF_RESULT_SUCCESS);
+
+        // capture provider and destroy it after the pool is destroyed
+        auto poolDestructor = [](umf_memory_pool_handle_t pool) {
```
igchor, line 143: you don;t need this `poolDestructor` anymore, just pass umfPoolDestroy as a second arg to `pool_unique_handle_t`

test/ipcAPI.cpp
```diff
@@ -0,0 +1,342 @@
...
+    for (size_t i = 0; i < NUM_POINTERS; ++i) {
+        void *ptr = umfPoolMalloc(pool.get(), ALLOC_SIZE);
+        EXPECT_NE(ptr, nullptr);
+        ptrs.push_back(ptr);
+    }
+
+    std::array<std::vector<umf_ipc_handle_t>, NTHREADS> ipcHandles;
+
+    auto getHandlesFn = [&ipcHandles, &ptrs](size_t tid) {
+        // TODO: better to wait on the barrier here so that every thread
```
igchor, line 224: We have a custom implementation of that in: https://github.com/oneapi-src/unified-memory-framework/blob/main/benchmark/multithread.hpp#L66 You can move it to a common location and use it here.

src/provider/provider_level_zero.c
```diff
@@ -254,6 +270,101 @@ static umf_result_t ze_memory_provider_allocation_split(void *provider,
...
+                                                      void *ipcData) {
+    ze_result_t ze_result;
+    struct ze_memory_provider_t *ze_provider =
+        (struct ze_memory_provider_t *)provider;
+    struct ipc_data_t *ipc_data = (struct ipc_data_t *)ipcData;
+
+    if (g_ze_ops.zeMemPutIpcHandle == NULL) {
+        // g_ze_ops.zeMemPutIpcHandle can be NULL because it was introduced
+        // starting from Level Zero 1.6
+        // TODO: Should we return UMF_RESULT_SUCCESS or UMF_RESULT_ERROR_NOT_SUPPORTED?
```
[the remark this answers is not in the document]
bratpiorka: ok I agree - could you remove the TODO and update the comment with something like "this handle can be released and according to the specification L0 < 1.6 we are done here"

test/ipcAPI.cpp
igchor: Yeah, and I can do that in a separate PR

src/provider/provider_tracking.c
bratpiorka: ok so I suggest to remove size param

src/provider/provider_tracking.c
```diff
@@ -429,6 +466,132 @@ static const char *trackingName(void *provider) {
...
+            ipc_cache_value_t *cache_value = (ipc_cache_value_t *)value;
+            memcpy(ipcData, cache_value->ipcData, cache_value->size);
+            cached = 1;
+        } else {
+            ret =
+                umfMemoryProviderGetIPCHandle(p->hUpstream, ptr, size, ipcData);
```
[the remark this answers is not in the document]
vinser52: > I tried but I don't get what each 'size' means :( can you change the names of these variables so they become more meaningful? or add some comments?

I have renamed `ipc_cache_value_t::size` to `ipc_cache_value_t::ipcDataSize`. And also I added comments to `ipc_cache_value_t` data structure and `umf_ipc_data_t` data structure. Hope it will make things clear.

> Also, as I saw Level Zero doesn't use this arg, right?

Correct. But the same story with `ze_memory_provider_free` function - there `bytes` parameter is not used in case of L0, but in general case it might be used by other implementation that is why we have the size parameter in these two interfaces of memory provider.

src/provider/provider_tracking.c
vinser52: I did some code cleanup, please review.

bratpiorka (approved):

bratpiorka:
@vinser52 overall LGTM
@lukaszstolarczuk I will take care of creating issues when this PR is merged

vinser52:
> @lukaszstolarczuk I will take care of creating issues when this PR is merged

@bratpiorka I am going to create issues today (right now).

igchor (approved):

vinser52:
drd and helgrind CI jobs should be fixed after PR #411 is merged

PR #89, opened 2023-12-20
Fix linking

igchor:
Some libraries that were dependencies of UMF were not link to UMF but rather directly to tests and benchmark. Fix this by extending add_umf_library and passing required libs there.


vinser52 (approved):

ldorau (approved):

ldorau:
Thanks, @igchor !

PR #91, opened 2023-12-21
Fix checking alignment in os_alloc()

ldorau:
The alignment has to be a multiple or a divider of the minimum page size.

Remove the `UMF_OS_RESULT_ERROR_WRONG_ALIGNMENT` error.

Ref: #90

### DIFF
--- a/src/provider/provider_os_memory.c
+++ b/src/provider/provider_os_memory.c
@@ -48,8 +48,6 @@ static __TLS os_last_native_error_t TLS_last_native_error;
 
 // helper values used only in the Native_error_str array
 #define _UMF_OS_RESULT_SUCCESS (UMF_OS_RESULT_SUCCESS - UMF_OS_RESULT_SUCCESS)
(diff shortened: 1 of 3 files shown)


src/provider/provider_os_memory.c
```diff
@@ -237,6 +233,8 @@ void os_finalize(void *provider) {
     free(os_provider);
 }
 
+umf_result_t os_get_min_page_size(void *provider, void *ptr, size_t *page_size);
```
vinser52, line 236: not related to this PR, but why we are marking only the `os_alloc` function as `static`?

nit: Instead of forward declaration we can just move `os_get_min_page_size` function definition here.

vinser52 (approved):

src/provider/provider_os_memory.c
ldorau: > not related to this PR, but why we are marking only the `os_alloc` function as `static`?
> 
Done: https://github.com/oneapi-src/unified-memory-framework/pull/92

PR #93, opened 2023-12-21
Add tests for alignment in OS memory provider

ldorau:
Add tests for alignment in OS memory provider and refactor tests to use gtest's `TEST_P()`.

### DIFF
--- a/test/provider_os_memory.cpp
+++ b/test/provider_os_memory.cpp
@@ -6,14 +6,21 @@
 
 #include "provider_os_memory_internal.h"
 #include "umf/providers/provider_os_memory.h"
+#include "umf_helpers.hpp"
 #include <umf/memory_provider.h>
 
 using umf_test::test;
 
-#define SIZE_4K (4096)
+#define INVALID_PTR ((void *)0x01)
 
 #define ASSERT_IS_ALIGNED(ptr, alignment)                                      \
     ASSERT_EQ(((uintptr_t)ptr % alignment), 0)
+
+typedef enum purge_t {
+    PURGE_NONE = 0,
+    PURGE_LAZY = 1,
+    PURGE_FORCE = 2,
+} purge_t;
 
 static umf_os_memory_provider_params_t UMF_OS_MEMORY_PROVIDER_PARAMS_TEST = {
     /* .protection = */ UMF_PROTECTION_READ | UMF_PROTECTION_WRITE,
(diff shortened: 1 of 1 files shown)


ldorau:
Add tests for alignment in OS memory provider and refactor tests to use gtest's `TEST_P()`.

test/provider_os_memory.cpp
```diff
@@ -39,396 +47,243 @@ static const char *Native_error_str[] = {
...
-        UMF_OS_MEMORY_PROVIDER_PARAMS_TEST;
+    auto ret =
+        umfMemoryProviderCreate(provider_ops, provider_params, &hProvider);
+    EXPECT_EQ(ret, UMF_RESULT_SUCCESS);
+    EXPECT_NE(hProvider, nullptr);
 
-    // wrong NUMA flags
-    os_memory_provider_params.numa_flags = (unsigned int)-1;
+    // capture provider and destroy it
+    auto providerDestructor = [](umf_memory_provider_handle_t hProvider) {
```
igchor, line 71: nit: you don't need to create a custom destructor here, you can just pass `umfMemoryProviderDestroy` directly to unique_ptr ctor, like this: `return umf::provider_unique_handle_t(hProvider, &umfMemoryProviderDestroy);`

This was needed for poolCreateExt because we wanted to destroy both pool and provider at the same time.

test/provider_os_memory.cpp
```diff
@@ -39,396 +47,243 @@ static const char *Native_error_str[] = {
...
-    umf_result = umfMemoryProviderCreate(&UMF_OS_MEMORY_PROVIDER_OPS,
-                                         &os_memory_provider_params,
-                                         &os_memory_provider);
+TEST_P(umfProviderTest, provider_os_memory_get_min_page_size) {
+    size_t min_page_size;
+    umf_result_t umf_result = umfMemoryProviderGetMinPageSize(
+        provider.get(), nullptr, &min_page_size);
     ASSERT_EQ(umf_result, UMF_RESULT_SUCCESS);
-    ASSERT_NE(os_memory_provider, nullptr);
+    ASSERT_LE(min_page_size, SIZE_4100);
```
igchor, line 241: Why compare with 4100? Should we even have a hardcoded page size here? There can be some systems with different page sizes right?

igchor (approved):

test/provider_os_memory.cpp
ldorau: Done

test/provider_os_memory.cpp
ldorau: Done

ldorau:
@igchor review again, please

bratpiorka (approved):

PR #105, opened 2024-01-05
Use hwloc instead of libnuma for os_provider

igchor:
This allows us to get rid of libnuma as a dependency.

One missing features from hwloc is lack of support for: `UMF_NUMA_FLAGS_MOVE_ALL`. Do we care about this?"
Also, I think we need more tests before we can consider merging this. Something where we actually verify if memory is bound to a specific NUAMA node.


igchor:
CI fails because I don't know how to best install hwloc on windows and macos yet ;d

ldorau:
> CI fails because I don't know how to best install hwloc on windows and macos yet ;d

I think we can keep OS provider disabled on Windows and MacOS.

bratpiorka:
@igchor I'm not sure about this change. WinAPI already supports NUMA (https://learn.microsoft.com/en-us/windows/win32/procthread/numa-support#numa-api) and with your changes, the UMF would require an additional dependency. Maybe we could use hwloc instead of libnuma only for Linux?

igchor:
> @igchor I'm not sure about this change. WinAPI already supports NUMA (https://learn.microsoft.com/en-us/windows/win32/procthread/numa-support#numa-api) and with your changes, the UMF would require an additional dependency. Maybe we could use hwloc instead of libnuma only for Linux?

I thought we'll need to rely on hwloc anyway for topology discovery (e.g. to implement HOST memspace).

igchor:
> > CI fails because I don't know how to best install hwloc on windows and macos yet ;d
> 
> I think we can keep OS provider disabled on Windows and MacOS as it is currently for now.

Ok. Done.

bratpiorka:
> > @igchor I'm not sure about this change. WinAPI already supports NUMA (https://learn.microsoft.com/en-us/windows/win32/procthread/numa-support#numa-api) and with your changes, the UMF would require an additional dependency. Maybe we could use hwloc instead of libnuma only for Linux?
> 
> I thought we'll need to rely on hwloc anyway for topology discovery (e.g. to implement HOST memspace).

OK but what I'm thinking about is that installing hwloc on Windows could be not as easy as for linux. I think we should add short description to the readme how to do that. Also, we should think which hwloc version we would require. Recently hwloc team released 2.10 version with some nice features supporting heterogeneous memory. Maybe for a stand-alone version (without oneAPI base toolkit) we should ship hwloc binaries inside deps or something.

igchor:
> > > @igchor I'm not sure about this change. WinAPI already supports NUMA (https://learn.microsoft.com/en-us/windows/win32/procthread/numa-support#numa-api) and with your changes, the UMF would require an additional dependency. Maybe we could use hwloc instead of libnuma only for Linux?
> > 
> > 
> > I thought we'll need to rely on hwloc anyway for topology discovery (e.g. to implement HOST memspace).
> 
> OK but what I'm thinking about is that installing hwloc on Windows could be not as easy as for linux. I think we should add short description to the readme how to do that. Also, we should think which hwloc version we would require. Recently hwloc team released 2.10 version with some nice features supporting heterogeneous memory. Maybe for a stand-alone version (without oneAPI base toolkit) we should ship hwloc binaries inside deps or something.

For now, the OS provider is not implemented for Windows anyway - when we do implement it, then yes, I agree, we should extend the readme. The version of hwloc we need will probably depend on the discovery use-case. For OS provider, we could even use 2.0

vinser52:
> Maybe for a stand-alone version (without oneAPI base toolkit) we should ship hwloc binaries inside deps or something.

@bratpiorka what do you mean by "standalone version"? We never discussed the distribution model for UMF except as part of oneAPI or source code on GitHub?

bratpiorka:
> > Maybe for a stand-alone version (without oneAPI base toolkit) we should ship hwloc binaries inside deps or something.
> 
> @bratpiorka what do you mean by "standalone version"? We never discussed the distribution model for UMF except as part of oneAPI or source code on GitHub?

Ok I was thinking about the far future. Sergey - do you know is hwloc lib is added into oneAPI?

vinser52:
> Sergey - do you know is hwloc lib is added into oneAPI?

Yes, it was added for TCM needs.

src/CMakeLists.txt
```diff
@@ -10,17 +10,16 @@ set(UMF_SOURCES
     memory_pool.c
     memory_provider.c
     memory_provider_get_last_failed.c
-    provider/provider_os_memory.c
```
bratpiorka, line 13: please do not change any windows - related code here

.github/workflows/basic.yml
```diff
@@ -33,7 +33,7 @@ jobs:
       - name: Install apt packages
         run: |
           sudo apt-get update
-          sudo apt-get install -y clang cmake libnuma-dev libjemalloc-dev libtbb-dev
```
bratpiorka, line 36: please update the README

bratpiorka (changes requested):

bratpiorka:
please rebase - also I think we could set this PR to "ready to review"

src/CMakeLists.txt
igchor: Why? I don't see any reason for keeping files that just have stubs. 

Also, this PR makes it necessary to store extra state (hwloc_topology) in OS provider struct. It's easier to keep it in an OS-agnostic way vs defining a struct that is OS-specific. We will probably move the hwloc_topology to another place anyway in future (to some library ctor or `initialize` method so we can support discovery).

bratpiorka: ok let's get back to this when we start adding support for windows

igchor:
I've rebased this PR and verified it works fine with tests from: https://github.com/oneapi-src/unified-memory-framework/pull/160 I think we can consider mergining it now. There are still some tests that would be useful (e.g. https://github.com/oneapi-src/unified-memory-framework/pull/144) but I don't think we need to actually wait on those.

.github/workflows/basic.yml
```diff
@@ -81,7 +81,7 @@ jobs:
...
+          sudo apt-get install -y clang cmake libhwloc-dev libnuma-dev libjemalloc-dev libtbb-dev
```
PatKamin, line 84: `linuma-dev` can be removed now?

igchor: It's still needed in tests. We could use hwloc in tests as well but I think using numa gives us more confidence that the integration is done correctly.

bratpiorka (approved):

PatKamin (approved):

PR #109, opened 2024-01-08
Add asserts for C tests

DamianDuy:
This PR adds asserts for C tests (using the implementation from pmemstream: https://github.com/pmem/pmemstream/blob/master/tests/common/unittest.h) and uses them in a c test for disjoint pool.

### DIFF
--- a/test/common/test_helpers.h
+++ b/test/common/test_helpers.h
@@ -1,11 +1,13 @@
-// Copyright (C) 2023 Intel Corporation
+// Copyright (C) 2023-2024 Intel Corporation
 // Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
-// This file contains tests for UMF pool API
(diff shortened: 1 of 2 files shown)


test/common/test_helpers.h
```diff
@@ -1,19 +1,91 @@
...
 
+#include <stdarg.h>
+#include <stdio.h>
 #include <umf/base.h>
 #include <umf/memory_pool.h>
 #include <umf/memory_provider_ops.h>
 
+#ifdef __cplusplus
+#include <stdexcept>
+static inline void UT_FATAL(const char *format, ...) {
```
igchor, line 17: You can just use the C-version of the UT_FATAL (the one under `#else`) since this is only expected to be used by C code anyway.

ldorau: @DamianDuy You are adding asserts for C tests, so you should add only the C version of `UT_FATAL`, not C++

DamianDuy: Done.

bratpiorka (approved):

PR #111, opened 2024-01-08
Add test for make uninstall target

ldorau:
Requires:
- [x] #108 
- [x] #110


test/test_make_uninstall.sh
```diff
@@ -0,0 +1,56 @@
...
+MATCH_UNINSTALLED_FILES="${WORKSPACE}/test/test_make_uninstall.txt"
+UNINSTALLED_FILES="${BUILD_DIR}/uninstall-files.txt"
+
+cd ${BUILD_DIR}
+make uninstall
+
+cd ${INSTALL_DIR}
+find | sort > ${UNINSTALLED_FILES}
+
+# check if 'make install' installed all and only required files
```
bratpiorka, line 52: please update the comment - 'make uninstall'? Also could you add a comment how does this work? Does this script checks if given directories are empty?

test/test_make_uninstall.txt
```diff
@@ -0,0 +1,11 @@
+.
```
bratpiorka, line 1: how does this work? why do we add '.' entry here?

ldorau: This file is an output of the `find | sort` command. `.` stands for the current directory.

test/test_make_uninstall.sh
ldorau: Done.
Yes, The `find` command prints out all types of general files (so also directories) found in the given directory.
After `make uninstall` there should be only empty directories left.
Done.

igchor (approved):

PR #117, opened 2024-01-09
Fix return values from jemalloc hooks

igchor:
For OS provider, merge, split and commits are nops. This means we should return false, not true from
those hooks.

Ref: https://github.com/memkind/memkind/blob/c425635feb0367c43da8fa0deadb50a4bc127988/src/memkind_arena.c

@kswiecicki this should fix the ASAN issue.


igchor:
Eventually, we'll need to implement split, merge, and commit as functions in the provider. I've added TODOs for now.

bratpiorka (approved):

bratpiorka:
FYI: I added a JIRA task related to added TODO

PR #118, opened 2024-01-10
Refactor the only .cpp file to pure C

PatKamin:
~~Dropping a call to critnib_delete(), leaving a memory leakage. However, the provider memory tracker object is not destroyed earlier than an application closure anyway.~~


src/provider/provider_tracking_linux.c
```diff
@@ -0,0 +1,25 @@
...
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ *
+ */
+
+#include "critnib.h"
+#include "provider_tracking.h"
+
+#include <stddef.h>
+
+critnib *TRACKER = NULL;
```
igchor, line 15: nit: this var can be static I think

src/provider/provider_tracking.c
```diff
@@ -233,6 +219,10 @@ umf_memory_provider_ops_t UMF_TRACKING_MEMORY_PROVIDER_OPS = {
 umf_result_t umfTrackingMemoryProviderCreate(
     umf_memory_provider_handle_t hUpstream, umf_memory_pool_handle_t hPool,
     umf_memory_provider_handle_t *hTrackingProvider) {
+#if defined(_WIN32) && !defined(UMF_SHARED_LIBRARY)
```
igchor, line 222: nit: I think it might be better to always implement `umfTrackingMemoryProviderInit` and just do nothing there if we're on linux or we build a static lib. This way we don't have to duplicate this condition averywhere.

src/provider/provider_tracking_windows.c
```diff
@@ -22,13 +24,17 @@ BOOL APIENTRY DllMain(HINSTANCE, DWORD fdwReason, LPVOID lpvReserved) {
...
-critnib *TRACKER = TRACKER_INSTANCE.map;
+INIT_ONCE init_once_flag = INIT_ONCE_STATIC_INIT;
+
+BOOL CALLBACK providerInit(PINIT_ONCE InitOnce, PVOID Parameter,
+                           PVOID *lpContext) {
+    TRACKER = critnib_new();
+    return TRACKER ? TRUE : FALSE;
+}
+
+void umfTrackingMemoryProviderInit(void) {
```
igchor, line 35: It would be good to have a comment stating that we intentionally leak the tracker.

igchor (changes requested):

src/provider/provider_tracking.c
```diff
@@ -233,6 +219,10 @@ umf_memory_provider_ops_t UMF_TRACKING_MEMORY_PROVIDER_OPS = {
...
+    umfTrackingMemoryProviderInit();
+#endif
+
     umf_tracking_memory_provider_t params;
     params.hUpstream = hUpstream;
     params.hTracker = umfMemoryTrackerGet();
```
bratpiorka, line 228: please check here if hTracker != NULL

src/provider/provider_tracking_windows.c
```diff
@@ -22,13 +24,17 @@ BOOL APIENTRY DllMain(HINSTANCE, DWORD fdwReason, LPVOID lpvReserved) {
...
-    critnib *map;
-} tracker_t;
-tracker_t TRACKER_INSTANCE;
-critnib *TRACKER = TRACKER_INSTANCE.map;
+INIT_ONCE init_once_flag = INIT_ONCE_STATIC_INIT;
+
+BOOL CALLBACK providerInit(PINIT_ONCE InitOnce, PVOID Parameter,
+                           PVOID *lpContext) {
+    TRACKER = critnib_new();
+    return TRACKER ? TRUE : FALSE;
```
bratpiorka, line 32: please try to register here custom deinit function and use atexit()

src/provider/provider_tracking_linux.c
PatKamin: Oh, I've changed the Windows file, forgot about this one. Thanks, now it's done.

bratpiorka (approved):

igchor (approved):

PR #119, opened 2024-01-10
Fix arena_extent_split() and arena_extent_merge()

ldorau:
Fix a regression introduced by commit 459a492de41a647d6dab2364d2f15c9603bf7a71
With this commit the jemalloc_pool test fails very often, see:
https://github.com/ldorau/unified-memory-framework/actions/workflows/basic.yml?query=branch%3ATest_real_make_install
This patch fixes this issue.

Ref: 459a492de41a647d6dab2364d2f15c9603bf7a71

PR #120, opened 2024-01-10
Test make install with benchmark

ldorau:


ldorau:
Builds failed because of (fix): https://github.com/oneapi-src/unified-memory-framework/pull/119

test/test_make_install_with_bench.sh
```diff
@@ -0,0 +1,70 @@
...
+	exit 1
+fi
+
+if [ ! -d $INSTALL_DIR ]; then
+	echo "Incorrect <install_dir> (it should exist): $INSTALL_DIR"
+	echo $USAGE_STR
+	exit 1
+fi
+
+if [ "$BUILD_TYPE" != "Release" -a "$BUILD_TYPE" != "Debug" ]; then
```
lukaszstolarczuk, line 38: Maybe I'm not aware of something - but why is it so important to verify it's either of these two? What if I use e.g. `RelWithDebInfo`...?

test/test_make_install_with_bench.sh
```diff
@@ -0,0 +1,70 @@
...
+set -e
+
+USAGE_STR="Usage: $(basename $0) <workspace_dir> <build_dir> <install_dir> <build_type:Release|Debug>"
+
+if [ "$4" == "" ]; then
+	echo $USAGE_STR
+	exit 1
+fi
+
+WORKSPACE=$(realpath $1)
```
lukaszstolarczuk, line 15: isn't it the workspace like a two levels (`../../`) up from the script file...?

test/test_make_install_with_bench.sh
```diff
@@ -0,0 +1,70 @@
...
+./benchmark/ubench && echo "Benchmark succeeded as expected" || exit 1
+
+# remove libumf.so
+ldd ./benchmark/ubench
+rm -f $(ldd ./benchmark/ubench | grep -e "libumf.so" | awk '{print $3}')
+
+./benchmark/ubench && exit 1 || echo "Benchmark failed as expected"
+LD_LIBRARY_PATH=${INSTALL_DIR}/lib/ ./benchmark/ubench && echo "Benchmark succeeded as expected" || exit 1
+
+echo "Success"
```
lukaszstolarczuk, line 70: not sure if needed, but you're not cleaning (e.g. install dir) after the test

test/test_make_install_with_bench.sh
```diff
@@ -0,0 +1,70 @@
...
+echo "WORKSPACE=$WORKSPACE"
+echo "BUILD_DIR=$BUILD_DIR"
+echo "INSTALL_DIR=$INSTALL_DIR"
+echo "BUILD_TYPE=$BUILD_TYPE"
+echo "BUILD_TYPE_STR=$BUILD_TYPE_STR"
+
+set -ex
+
+# clean the build directory
+rm -rf ${BUILD_DIR}/*
```
lukaszstolarczuk, line 55: I know in CI you're entering this dir, but if this script is used manually it may be in a wrong dir - make sure to use `cd`/`pushd`

test/test_make_install_with_bench.sh
```diff
@@ -0,0 +1,70 @@
...
+ls -al
+
+cmake .. -DCMAKE_INSTALL_PREFIX="${INSTALL_DIR}" -DCMAKE_BUILD_TYPE=${BUILD_TYPE} -DUMF_BUILD_SHARED_LIBRARY=ON -DUMF_BUILD_TESTS=OFF -DUMF_BUILD_BENCHMARKS=ON
+make -j $(nproc)
+make install
+./benchmark/ubench && echo "Benchmark succeeded as expected" || exit 1
+
+# remove libumf.so
+ldd ./benchmark/ubench
+rm -f $(ldd ./benchmark/ubench | grep -e "libumf.so" | awk '{print $3}')
```
lukaszstolarczuk, line 65: hmm... if I run this on my PC it may remove my UMF installed in the OS...? perhaps move the file and move it back afterwards...?
// or only make sure to remove from the BUILD_DIR...?

test/test_make_install_with_bench.sh
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash
```
lukaszstolarczuk, line 1: hmm... a little generic question first: why can't this be a part of the other script `test_make_install.sh`?

.github/workflows/basic.yml
```diff
@@ -93,6 +93,11 @@ jobs:
         working-directory: ${{env.BUILD_DIR}}
         run: ${{github.workspace}}/test/test_make_uninstall.sh ${{github.workspace}} ${{env.BUILD_DIR}} ${{env.INSTL_DIR}}
 
+      - name: Test make install with benchmark
```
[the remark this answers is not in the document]
bratpiorka: @lukaszstolarczuk we have a task for that: UMFW-198

test/test_make_install_with_bench.sh
bratpiorka: I think this should be separate tests but could partially use the test_make_install.sh so we don't duplicate the code

bratpiorka (changes requested):

igchor:
@ldorau, what does this test verify exactly? We're just using LD_LIBRARY_PATH to test if using a different location for the .so works. However, it doesn't test whether we can compile against the installed lib, right? (the binary is still linked/compiled against the libumf within the same CMake project).

I think a better test would be to have a standalone CMake project (we can make the benchmark a standalone project, similarly like we've done for all examples in pmemstream/libpmemobj-cpp:  https://github.com/pmem/pmemstream/blob/master/examples/01_basic_iterate/CMakeLists.txt in addition to having the bench as part of the umf CMake project). Then we can just install libumf and verify that we this standalone benchmark will find it using find_package(umf). And this will verify that both compilation and linking works fine.

igchor:
We might also want to add a test for verifying whether FetchContent works with UMF but this is a separate thing. (@bratpiorka 
 this is what I mentioned to you, do we have a task for that already?)

ldorau:
> I think a better test would be to have a standalone CMake project (we can make the benchmark a standalone project, similarly like we've done for all examples in pmemstream/libpmemobj-cpp: https://github.com/pmem/pmemstream/blob/master/examples/01_basic_iterate/CMakeLists.txt in addition to having the bench as part of the umf CMake project). Then we can just install libumf and verify that we this standalone benchmark will find it using find_package(umf). And this will verify that both compilation and linking works fine.

I agree. You are right.

ldorau:
It was a wrong way ...

PR #124, opened 2024-01-10
[Security] Add scorecard.yml

wlemkows:


.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,56 @@
...
+name: Scorecard supply-chain security
+on:
+  # For Branch-Protection check. Only the default branch is supported. See
+  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#branch-protection
+  branch_protection_rule:
+  # To guarantee Maintained check is occasionally updated. See
+  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#maintained
+  workflow_dispatch:
+  schedule:
+    - cron: '36 1 * * 6'
```
PatKamin, line 14: nit: Add a comment with the schedule deciphered, like: Every Saturday at 1:36 UTC

.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,56 @@
+# This workflow uses actions that are not certified by GitHub. They are provided
+# by a third-party and are governed by separate terms of service, privacy
+# policy, and support documentation.
+
+name: Scorecard supply-chain security
+on:
+  # For Branch-Protection check. Only the default branch is supported. See
+  # https://github.com/ossf/scorecard/blob/main/docs/checks.md#branch-protection
+  branch_protection_rule:
```
PatKamin, line 9: nit: Colons are probably redundant here and in `workflow_dispatch`.

.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,56 @@
...
+      - name: "Upload artifact"
+        uses: actions/upload-artifact@3cea5372237819ed00197afe530f5a7ea3e805c8 # v3.1.0
+        with:
+          name: SARIF file
+          path: results.sarif
+          retention-days: 5
+
+      # Upload the results to GitHub's code scanning dashboard.
+      - name: "Upload to code-scanning"
+        uses: github/codeql-action/upload-sarif@17573ee1cc1b9d061760f3a006fc4aac4f944fd5 # v2.2.4
```
PatKamin, line 54: Do we need to state versions of actions in this workflow with such detail?

.github/workflows/scorecard.yml
lukaszstolarczuk: there's probably more stuff that's not needed, but I guess this workflow was generated, and we may leave it that way..

.github/workflows/scorecard.yml
lukaszstolarczuk: heh, hashes are enforced by scorecard 😉

bratpiorka (approved):

.github/workflows/scorecard.yml
lukaszstolarczuk: yes, I agree. I can go both ways, though.

wlemkows: Done.

.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,56 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      # Needed to upload the results to code-scanning dashboard.
+      security-events: write
+      # Needed to publish results and get a badge (see publish_results below).
+      id-token: write
+
+    steps:
+      - name: "Checkout code"
+        uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # v3.1.0
```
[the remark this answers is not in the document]
wlemkows: done

.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,57 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      # Needed to upload the results to code-scanning dashboard.
+      security-events: write
+      # Needed to publish results and get a badge (see publish_results below).
+      id-token: write
+
+    steps:
+      - name: "Checkout code"
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
```
KFilipek, line 34: Why do you use full SHA1 instead of using a tagged version?
In this case: [v4.1.1](https://github.com/actions/checkout/tree/v4.1.1)

bratpiorka: hashes are enforced by scorecard

wlemkows: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-third-party-actions
Please read **Pin actions to a full length commit SHA**

PR #126, opened 2024-01-10
Add flags param to umfPoolCreate

igchor:


igchor:
@vinser52 @bratpiorka as agreed, this PR extend umfPoolCreate with flags param. This replaces https://github.com/oneapi-src/unified-memory-framework/pull/97

include/umf/memory_pool.h
```diff
@@ -30,17 +30,31 @@ struct umf_memory_pool_ops_t;
 ///
 typedef struct umf_memory_pool_ops_t umf_memory_pool_ops_t;
 
+/// @brief Supported pool creation flags
+typedef uint32_t umf_pool_create_flags_t;
+typedef enum umf_pool_create_flag_t {
+    UMF_POOL_CREATE_FLAG_OWN_PROVIDER =
```
bratpiorka, line 36: could we add an enum UMF_POOL_CREATE_FLAG_DEFAULT? I don't like passing 0 as flags

bratpiorka (approved):

vinser52 (approved):

include/umf/memory_pool.h
vinser52: from my point of view `0` is more clear. If the user sees `UMF_POOL_CREATE_FLAG_DEFAULT`, it is unclear which combination of flags is the default. And if the user wants to set some flag do you need to append it to default (e.g. `UMF_POOL_CREATE_FLAG_DEFAULT | SOME_FLAG`) or just pass this single flag?

bratpiorka: ok we could call this as "UMF_POOL_CREATE_FLAG_NONE" - that way we don't have the problems of appending a single flag. What I like about passing enum instead of just "0" is that when I read it, I don't have to wonder what argument I set, because it's more self-explanatory. But the PR has been approved by me, so @igchor, if you like as it is, you can merge this PR

igchor: I also prefer just passing 0, especially since it's shorter and create already has a lot of params. However, I'm fine adding the `UMF_POOL_CREATE_FLAG_NONE` to the enum, this is C, so passing both `UMF_POOL_CREATE_FLAG_NONE` and `0` will work.

igchor: Done.

PR #129, opened 2024-01-11
Fix memory leak in the freeErrorPropagation disjoint_pool test

ldorau:


bratpiorka (approved):

test/pools/disjoint_pool.cpp
```diff
@@ -58,8 +58,13 @@ TEST_F(test, freeErrorPropagation) {
     void *ptr = umfPoolMalloc(pool, size);
 
     freeReturn = UMF_RESULT_ERROR_MEMORY_PROVIDER_SPECIFIC;
```
lukaszstolarczuk, line 60: this `freeReturn` name is confusing, perhaps rename to `expectedRet` or something...?

test/pools/disjoint_pool.cpp
```diff
@@ -58,8 +58,13 @@ TEST_F(test, freeErrorPropagation) {
...
+    // this umfPoolFree() will not free the memory
```
lukaszstolarczuk, line 61: and why is that...?

// that's me asking, not knowing what's inside UMF repo 😉 you're calling umfPoolFree twice and the second one magically will do the freeing?  :D

ldorau: Yes, take a look 30 lines above:
```c
        umf_result_t free(void *ptr, [[maybe_unused]] size_t size) noexcept {
            if (freeReturn == UMF_RESULT_SUCCESS) {
                ::free(ptr);
            }
            return freeReturn;
        }
```

test/pools/disjoint_pool.cpp
ldorau: I think it is OK, take a look 30 lines above:
```c
        umf_result_t free(void *ptr, [[maybe_unused]] size_t size) noexcept {
            if (freeReturn == UMF_RESULT_SUCCESS) {
                ::free(ptr);
            }
            return freeReturn;
        }
```

lukaszstolarczuk: I'm not saying it's wrong, I'm just suggesting to rename the var. Now I see it's defined waay above, so perhaps we could rename it as a separate change/commit... but it'd be nice

test/pools/disjoint_pool.cpp
lukaszstolarczuk: heh, I was looking for some explanation (at best, added to the comment), but if you don't have it for me then no biggie, it's ok.

test/pools/disjoint_pool.cpp
ldorau: I meant this variable holds a value that **free()** will **return**, so I think `freeReturn` is a good name.

test/pools/disjoint_pool.cpp
ldorau: What do you want me to explain more? What needs to be explained?

igchor (approved):

test/pools/disjoint_pool.cpp
KFilipek: This rename should apply to the whole file, but this PR doesn't touch any other line and still keeps the convention.

KFilipek (approved):

test/pools/disjoint_pool.cpp
ldorau: > I'm not saying it's wrong, I'm just suggesting to rename the var. Now I see it's defined waay above, so perhaps we could rename it as a separate change/commit... but it'd be nice

@lukaszstolarczuk BTW this PR regards fixing memory leaks. If you want to rename a variable, you can submit a pull request.

lukaszstolarczuk: > I meant this variable holds a value that **free()** will **return**, so I think `freeReturn` is a good name.

Confusing for me was the usage in this part:

`EXPECT_EQ(freeRet, freeReturn);` 😉 

but, ok, I agree, it can be done as a separate change

test/pools/disjoint_pool.cpp
lukaszstolarczuk: why do we need a second (exactly the same) free call, one after another?

but as I said, if it's just me or you don't have a good explanation for me, then ok, let's move on :)

bratpiorka: @lukaszstolarczuk This test works in the way that we modify the default behavior of the free() so we can force it to either do the actual free or return the expected error code. When we call this function for the first time we expect it to fail (and we test that) while the second time we do the cleanup to avoid memory leak.

lukaszstolarczuk: And that's the reasonable answer - can we put a comment like this into the description of the test, or near the second free?
Since I was looking for the first time ever at this test, that was certainly not clear to me. Perhaps, if the first free was called e.g. `mock_free` then it would make more sense... just saying 😄

bratpiorka: I will create a separate small PR for that today

PR #132, opened 2024-01-11
Add TSAN annotations to pool_scalable

igchor:
To fix false positives from TSAN.

Found by @kswiecicki. Can you verify that this fixes the issue?

### DIFF
--- a/src/pool/pool_scalable.c
+++ b/src/pool/pool_scalable.c
@@ -1,6 +1,6 @@
 /*
  *
- * Copyright (C) 2023 Intel Corporation
+ * Copyright (C) 2023-2024 Intel Corporation
  *
  * Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
  * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
@@ -24,6 +24,7 @@
 #include <tbb/scalable_allocator.h>
 
 #include "utils_common.h"
+#include "utils_sanitizers.h"
 
 typedef void *(*raw_alloc_tbb_type)(intptr_t, size_t *);
 typedef void (*raw_free_tbb_type)(intptr_t, void *, size_t);
@@ -179,6 +180,7 @@ static void *tbb_malloc(void *pool, size_t size) {
         }
         return NULL;
     }
+    utils_annotate_acquire(pool);
     return ptr;
 }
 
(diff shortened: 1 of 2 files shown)


kswiecicki:
> To fix false positives from TSAN.
> 
> Found by @kswiecicki. Can you verify that this fixes the issue?

It does fix the issue in `tbb_calloc`. Wouldn't it be a good idea to also add those annotations in other tbb callbacks that acquire or release memory?

src/pool/pool_scalable.c
```diff
@@ -191,6 +192,13 @@ static void *tbb_calloc(void *pool, size_t num, size_t size) {
         return NULL;
     }
 
+    // synchronizes with release in tbb_free:
+    // if another thread modified memory allocated from tbb and called free
+    // and we end up getting this memory from tbb_malloc above we need to
+    // tell asan that the memset below does not conflict with any preceding
```
bratpiorka, line 198: tell TSAN? still after reading this I don't understand what is the scenario that reproduces this issue. Also, why we don't add such annotation to tbb_malloc?

bratpiorka (changes requested):

igchor:
> > To fix false positives from TSAN.
> > Found by @kswiecicki. Can you verify that this fixes the issue?
> 
> It does fix the issue in `tbb_calloc`. Wouldn't it be a good idea to also add those annotations in other tbb callbacks that acquire or release memory?

Good point, done. Note that I switched from calling acquire/release on 'ptr' to calling them on 'pool'. The reason is that it's possible for part of the allocation that is freed to be assigned to another thread. In that case, the 'ptr' will be different but the writes will still overlap.

src/pool/pool_scalable.c
igchor: Consider the following case:
```
thread 1                          |  thread 2
ptr = alloc(size) // ptr == 0xAAA |
memset(ptr, 0, size)              |          
free(ptr);                        | 
                                  | ptr = alloc(size) // ptr == 0xAAA
                                  | memset(ptr, 1, size);
                                  | free(ptr);
```

TSAN will report that there is a race on 'ptr' because it thinks that allocation and memset in thread 2 can happen before free(ptr) is called in thread 1 (it doesn't know that there's a happens-before relation between free and alloc, this is what I'm telling it using those annotations).

TBB actually supports TSAN but only if you pass appropriate flag during compilation. So the alternative to this PR could be to just use custom-compiled version of TBB but then you'd need to enforce this in CMake somehow (which does not sound like a good idea).

bratpiorka: Ok I got it, thanks for the comprehensive explanation.

bratpiorka (approved):

ldorau (approved):

PR #134, opened 2024-01-12
Fix skipping memspace_numa tests

ldorau:
Fix skipping memspace_numa tests.
`GTEST_SKIP()` called in `numa_nodes_test` does not skip `numa_memspace_test` and `numa_memspace_provider_test` tests.

It fixes the following test failure:
```
[----------] 1 test from numa_memspace_test
[ RUN      ] numa_memspace_test.provider_from_numa_memspace
/home/ldorau/work/unified-memory-framework/test/memspace_numa.cpp:21: Skipped
Failed to initialize libnuma
/home/ldorau/work/unified-memory-framework/test/memspace_numa.cpp:43: Failure
Expected equality of these values:
  ret
    Which is: 3
  UMF_RESULT_SUCCESS
    Which is: 0
umf_test-memspace_numa: /home/ldorau/work/unified-memory-framework/src/memspace.c:101: umfMemspaceDestroy: Assertion `memspace' failed.


90% tests passed, 1 tests failed out of 10

Label Time Summary:
umf    =   2.71 sec*proc (10 tests)

Total Test time (real) =   2.72 sec

The following tests FAILED:
         10 - umf-memspace_numa (Subprocess aborted)
Errors while running CTest
```

### DIFF
--- a/test/memspace_numa.cpp
+++ b/test/memspace_numa.cpp
@@ -37,17 +37,19 @@ struct numa_nodes_test : ::umf_test::test {
 struct numa_memspace_test : ::numa_nodes_test {
     void SetUp() override {
         ::numa_nodes_test::SetUp();
-
-        enum umf_result_t ret = umfMemspaceCreateFromNumaArray(
-            nodeIds.data(), nodeIds.size(), &memspace);
-        ASSERT_EQ(ret, UMF_RESULT_SUCCESS);
-        ASSERT_NE(memspace, nullptr);
+        if (nodeIds.size()) {
+            enum umf_result_t ret = umfMemspaceCreateFromNumaArray(
+                nodeIds.data(), nodeIds.size(), &memspace);
+            ASSERT_EQ(ret, UMF_RESULT_SUCCESS);
+            ASSERT_NE(memspace, nullptr);
+        }
     }
 
     void TearDown() override {
         ::numa_nodes_test::TearDown();
-
-        umfMemspaceDestroy(memspace);
+        if (memspace) {
+            umfMemspaceDestroy(memspace);
+        }
     }
 
     umf_memspace_handle_t memspace = nullptr;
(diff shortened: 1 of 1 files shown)


ldorau:
@igchor please review - maybe you know a better way to fix this issue than that?

test/memspace_numa.cpp
```diff
@@ -32,12 +35,17 @@ struct numa_nodes_test : ::umf_test::test {
...
 
     std::vector<size_t> nodeIds;
+    int skip_this_test;
 };
 
 struct numa_memspace_test : ::numa_nodes_test {
     void SetUp() override {
         ::numa_nodes_test::SetUp();
 
+        if (skip_this_test) {
```
bratpiorka, line 45: could we just create a helper function that checks condition from SetUp() and call it here?

bratpiorka (changes requested):

test/memspace_numa.cpp
```diff
@@ -66,6 +82,10 @@ struct numa_memspace_provider_test : ::numa_memspace_test {
     void TearDown() override {
         ::numa_memspace_test::TearDown();
 
+        if (skip_this_test) {
```
PatKamin, line 85: Skipping tests should be done before they begin (during SetUp()), not at the cleaning step (during TearDown()).

test/memspace_numa.cpp
PatKamin: All other modified test classes call `::numa_nodes_test::SetUp()` anyway, so there should be only one check in  `::numa_nodes_test::SetUp()`. All `skuip_this_test` flags can be removed.

bratpiorka: @PatKamin ok originally we called GTEST_SKIP() only in SetUp() and this didn't work - see the description of the PR

igchor:
> @igchor please review - maybe you know a better way to fix this issue than that?

I think the problem is this: `GTEST_SKIP()` only skips tests, but it does not affect any code that is in the SetUp() function. So, even if GTEST_SKIP() is called in the `numa_nodes_test::SetUp()`, `umfMemspaceCreateFromNumaArray` will still be called and fail.

To fix this, you can just put memspace/provider creation under an if:

```
struct numa_memspace_test : ::numa_nodes_test {
    void SetUp() override {
        ::numa_nodes_test::SetUp();

        if (nodeIds.size()) {
            enum umf_result_t ret = umfMemspaceCreateFromNumaArray(
                nodeIds.data(), nodeIds.size(), &memspace);
            ASSERT_EQ(ret, UMF_RESULT_SUCCESS);
            ASSERT_NE(memspace, nullptr);
        }
    }
    ...
```

and you might also need to modify TearDown():

```
    void TearDown() override {
        ::numa_nodes_test::TearDown();
        if (memspace)
            umfMemspaceDestroy(memspace);
    }
```

test/memspace_numa.cpp
ldorau: So try and remove it ;-) I would not add it if it were not needed :-)

ldorau:
> > @igchor please review - maybe you know a better way to fix this issue than that?
> 
> I think the problem is this: `GTEST_SKIP()` only skips tests, but it does not affect any code that is in the SetUp() function. So, even if GTEST_SKIP() is called in the `numa_nodes_test::SetUp()`, `umfMemspaceCreateFromNumaArray` will still be called and fail.

Thanks! This is a lot simpler :-)
Done.

bratpiorka (approved):

PatKamin (approved):

Issue #133, opened 2024-01-12
[CI] Add code style check


KFilipek:
Shortly, add code style check to the CI.

Source:
[> IMHO, code style check should be a part of this (newly added in this PR) "FastBuild" job.
](https://github.com/oneapi-src/unified-memory-framework/pull/115#discussion_r1450117686)

PR #136, opened 2024-01-12
[CI] Add coding style check as prerequisite

KFilipek:
This PR introduces a coding style check.

### DIFF
--- a/.github/workflows/pr_push.yml
+++ b/.github/workflows/pr_push.yml
@@ -48,10 +48,34 @@ jobs:
         - name: Build
           run: >
             cmake --build ${{github.workspace}}/build --config Release -j ${{matrix.nproc}}
+
+    CodeStyle:
+      name: Coding style
+      runs-on: ubuntu-latest
+
+      steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Install apt packages
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y cmake clang-format-15
+
+      - name: Configure CMake
+        run: >
+          cmake
+          -B ${{github.workspace}}/build
+          -DUMF_FORMAT_CODE_STYLE=ON
+
+      - name: Build
+        run: >
+          cmake --build build --target clang-format-check  
+
     Spellcheck:
         uses: oneapi-src/unified-memory-framework/.github/workflows/spellcheck.yml@main
     Build:
-        needs: [Spellcheck, FastBuild]
+        needs: [Spellcheck, FastBuild, CodeStyle]
         uses: oneapi-src/unified-memory-framework/.github/workflows/basic.yml@main
     Benchmark:
         needs: [Build]
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -24,7 +24,7 @@ jobs:
             shared_library: 'OFF'
           - os: 'ubuntu-22.04'
             build_type: Release
-            compiler: {c: clang, cxx: clang++}            
+            compiler: {c: clang, cxx: clang++}
             shared_library: 'OFF'
           - os: 'ubuntu-22.04'
             build_type: Release
@@ -156,13 +156,9 @@ jobs:
           cmake
           -B ${{env.BUILD_DIR}}
           -DCMAKE_BUILD_TYPE=Release
-          -DUMF_FORMAT_CODE_STYLE=ON
+          -DUMF_FORMAT_CODE_STYLE=OFF
           -DUMF_DEVELOPER_MODE=ON
           -DUMF_ENABLE_POOL_TRACKING=ON
-
-      - name: Run code-style check     
-        run: |
-          cmake --build ${{env.BUILD_DIR}} --target clang-format-check
 
       - name: Build UMF
         run: |


vinser52 (approved):

.github/workflows/pr_push.yml
```diff
@@ -48,6 +48,31 @@ jobs:
         - name: Build
           run: >
             cmake --build ${{github.workspace}}/build --config Release -j ${{matrix.nproc}}
+
+    CodeStyle:
```
PatKamin, line 52: Add this as a requirement to the Build job.

.github/workflows/pr_push.yml
```diff
@@ -48,6 +48,31 @@ jobs:
...
+      runs-on: ubuntu-latest
+
+      steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Install apt packages
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y cmake clang-format-15 libjemalloc-dev libnuma-dev libtbb-dev
```
PatKamin, line 63: Libs are redundant as you don't compile the sources, just disable building parts of umf which require these libs.

.github/workflows/pr_push.yml
```diff
@@ -48,6 +48,31 @@ jobs:
...
+      - name: Configure CMake
+        run: >
+          cmake
+          -B ${{github.workspace}}/build
+          -DUMF_FORMAT_CODE_STYLE=ON
+
+      - name: Build
+        working-directory: ${{github.workspace}}/build
+        run: >
+          make clang-format-check
```
PatKamin, line 74: nit: We could use cmake everywhere for code portability:
```
cmake --build build --target clang-format-check
```
You can drop the change of the working directory then.

.github/workflows/pr_push.yml
KFilipek: Done

.github/workflows/pr_push.yml
KFilipek: Done

.github/workflows/pr_push.yml
```diff
@@ -48,6 +48,31 @@ jobs:
...
+      - name: Configure CMake
+        run: >
+          cmake
+          -B ${{github.workspace}}/build
+          -DUMF_FORMAT_CODE_STYLE=ON
+
+      - name: Build
+        run: >
+          cmake --build build --target clang-format-check  
+
```
lukaszstolarczuk, line 74: redundant empty line

lukaszstolarczuk:
Please add to the commit msg info about #133

`Fixes: #133`

.github/workflows/pr_push.yml
lukaszstolarczuk, line 74: I can see that clang-format-check is run on Mac (in basic.yml) - should we get rid of it now...?

.github/workflows/pr_push.yml
```diff
@@ -48,6 +48,31 @@ jobs:
...
+      - name: Install apt packages
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y cmake clang-format-15
+
+      - name: Configure CMake
+        run: >
+          cmake
+          -B ${{github.workspace}}/build
+          -DUMF_FORMAT_CODE_STYLE=ON
```
ldorau, line 69: The coding style of the whole repository should be checked, so all components turned off by default should be also enabled:
- UMF_BUILD_LIBUMF_POOL_DISJOINT
- UMF_BUILD_LIBUMF_POOL_JEMALLOC
- UMF_BUILD_LIBUMF_POOL_SCALABLE
- UMF_BUILD_BENCHMARKS

ldorau, line 69: The coding style check is so fast that - in my opinion - (apart from this build at the beginning that no doubt is needed) the coding style check could be added to all basic builds, just to be sure that coding style is verified in all possible configurations.

ldorau (changes requested):

.github/workflows/pr_push.yml
bratpiorka: The coding style is checked for all h, hpp, c, cpp etc files in the project. There is no need to run it multiple times. If we decided to use separate workflows we should keep them short and  ideally each should be responsible for something different.

lukaszstolarczuk: does it make any difference...? we use glob pattern to find files in repo for scanning

.github/workflows/pr_push.yml
KFilipek: Done

.github/workflows/pr_push.yml
KFilipek: Done

.github/workflows/pr_push.yml
KFilipek: Done

PatKamin (approved):

lukaszstolarczuk (approved):

.github/workflows/pr_push.yml
ldorau: Right

ldorau: Right

ldorau (approved):

PR #138, opened 2024-01-13
Expose 'metadata_use_provider' for jemalloc pool

igchor:
TODO: the test doesn't seem to work... That is, even if I don't set `metadat_use_provider=false` jemalloc doesn't seem to allocate metadata using hooks.


igchor:
Closing until we have a reliable test

PR #139, opened 2024-01-13
Link jemalloc pool and test with libc explicitly to avoid recursive calls to jemalloc (tmp fix)

igchor:
This forces the linker to link with libc before linking with jemalloc. Because of that, jemalloc will not override symbols for malloc/free.

Without this patch, we would call `malloc` from within the `arena_extent_alloc`  (jemalloc hook) that would be intercepted by jemalloc.

This fixes the issue with debug jemalloc when running tests with tracking enabled:
```
<jemalloc>: Should own 0 locks of rank >= 14: extent_grow(17)
Aborted (core dumped)
```

### DIFF
--- a/src/pool/CMakeLists.txt
+++ b/src/pool/CMakeLists.txt
@@ -30,10 +30,12 @@ endif()
 # libumf_pool_jemalloc
 if(UMF_BUILD_LIBUMF_POOL_JEMALLOC)
     if(LINUX)
+        # TODO: we link directly with libc to force using malloc from glibc
+        # This is a temporary fix for https://github.com/oneapi-src/unified-memory-framework/issues/161
         add_umf_library(NAME jemalloc_pool
                         TYPE STATIC
                         SRCS pool_jemalloc.c
-                        LIBS jemalloc umf_utils)
+                        LIBS c jemalloc umf_utils)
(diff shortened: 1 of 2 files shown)


pbalcer (approved):

bratpiorka:
Interesting, I thought in this case the linking would fail because of multiple definitions of the same symbols. Anyway - you added a TODO in the code - what would be the final solution?

vinser52:
I would extend `TODO` you added with the detailed explanations like you posted for this PR:

> This forces the linker to link with libc before linking with jemalloc. Because of that, jemalloc will not override symbols for malloc/free.

> Without this patch, we would call malloc from within the arena_extent_alloc (jemalloc hook) that would be intercepted by jemalloc.

> This fixes the issue with debug jemalloc when running tests with tracking enabled:

> <jemalloc>: Should own 0 locks of rank >= 14: extent_grow(17)
> Aborted (core dumped)

ldorau (approved):
@igchor Please add a detailed comment to the code like @vinser52 suggested above

igchor:
> @igchor Please add a detailed comment to the code like @vinser52 suggested above

Done. I've created an issue and added link to that issue in the comments.

bratpiorka (approved):

PR #140, opened 2024-01-13
Implement umfMemoryProviderAllocation[Split/Merge]

igchor:
That is needed to support jemalloc (avoid leaking memory).

@kswiecicki could you verify if this solves all the memory leaks? On my machine, it seems to work fine after those changes.

### DIFF
--- a/include/umf/memory_provider.h
+++ b/include/umf/memory_provider.h
@@ -156,6 +156,34 @@ umfMemoryProviderGetName(umf_memory_provider_handle_t hProvider);
 ///
 UMF_EXPORT umf_memory_provider_handle_t umfGetLastFailedMemoryProvider(void);
 
+///
+/// @brief Splits a coarse grain allocation into 2 adjacent allocations that
+///        can be managed (freed) separately.
+/// @param hProvider handle to the memory provider
+/// @param ptr pointer to the beginning of the allocation
+/// @param totalSize total size of the allocation to be split
+/// @param firstSize size of the first new allocation, second allocation
+//         has a size equal to totalSize - firstSize
+/// @return UMF_RESULT_SUCCESS on success or appropriate error code on failure
+///
+UMF_EXPORT umf_result_t
+umfMemoryProviderAllocationSplit(umf_memory_provider_handle_t hProvider,
+                                 void *ptr, size_t totalSize, size_t firstSize);
+
+///
+/// @brief Merges two coarse grain allocations into a single allocation that
+///        can be managed (freed) as a whole.
+/// @param hProvider handle to the memory provider
+/// @param lowPtr pointer to the first allocation
+/// @param highPtr pointer to the second allocation (should be > lowPtr)
+/// @param totalSize size of a new merged allocation. Should be equal
+///        to the sum of sizes of allocations beginning at lowPtr and highPtr
+/// @return UMF_RESULT_SUCCESS on success or appropriate error code on failure
+///
+UMF_EXPORT umf_result_t
+umfMemoryProviderAllocationMerge(umf_memory_provider_handle_t hProvider,
+                                 void *lowPtr, void *highPtr, size_t totalSize);
+
 #ifdef __cplusplus
 }
 #endif
--- a/include/umf/memory_provider_ops.h
+++ b/include/umf/memory_provider_ops.h
@@ -139,6 +139,32 @@ typedef struct umf_memory_provider_ops_t {
     /// @return pointer to a string containing the name of the \p provider
     ///
     const char *(*get_name)(void *provider);
+
+    ///
+    /// @brief Splits a coarse grain allocation into 2 adjacent allocations that
+    ///        can be managed (freed) separately.
+    /// @param hProvider handle to the memory provider
+    /// @param ptr pointer to the beginning of the allocation
+    /// @param totalSize total size of the allocation to be split
+    /// @param firstSize size of the first new allocation, second allocation
+    //         has a size equal to totalSize - firstSize
+    /// @return UMF_RESULT_SUCCESS on success or appropriate error code on failure
+    ///
+    umf_result_t (*allocation_split)(void *hProvider, void *ptr,
+                                     size_t totalSize, size_t firstSize);
+
+    ///
+    /// @brief Merges two coarse grain allocations into a single allocation that
+    ///        can be managed (freed) as a whole.
+    /// @param hProvider handle to the memory provider
+    /// @param lowPtr pointer to the first allocation
+    /// @param highPtr pointer to the second allocation (should be > lowPtr)
+    /// @param totalSize size of a new merged allocation. Should be equal
+    ///        to the sum of sizes of allocations beginning at lowPtr and highPtr
+    /// @return UMF_RESULT_SUCCESS on success or appropriate error code on failure
+    ///
+    umf_result_t (*allocation_merge)(void *hProvider, void *lowPtr,
+                                     void *highPtr, size_t totalSize);
 } umf_memory_provider_ops_t;
 
 #ifdef __cplusplus
--- a/src/pool/pool_jemalloc.c
+++ b/src/pool/pool_jemalloc.c
@@ -210,15 +210,15 @@ static bool arena_extent_split(extent_hooks_t *extent_hooks, void *addr,
                                size_t size, size_t size_a, size_t size_b,
                                bool committed, unsigned arena_ind) {
     (void)extent_hooks; // unused
-    (void)addr;         // unused
-    (void)size;         // unused
-    (void)size_a;       // unused
-    (void)size_b;       // unused
     (void)committed;    // unused
-    (void)arena_ind;    // unused
+    (void)size_b;
 
-    // TODO: add this function to the provider API to support Windows and USM
-    return true; // true means failure (unsupported)
+    assert(size_a + size_b == size);
+
+    jemalloc_memory_pool_t *pool = get_pool_by_arena_index(arena_ind);
+    assert(pool);
+    return umfMemoryProviderAllocationSplit(pool->provider, addr, size,
+                                            size_a) != UMF_RESULT_SUCCESS;
 }
 
 // arena_extent_merge - an extent merge function conforms to the extent_merge_t type and optionally
@@ -231,15 +231,13 @@ static bool arena_extent_merge(extent_hooks_t *extent_hooks, void *addr_a,
                                size_t size_a, void *addr_b, size_t size_b,
                                bool committed, unsigned arena_ind) {
     (void)extent_hooks; // unused
-    (void)addr_a;       // unused
-    (void)addr_b;       // unused
-    (void)size_a;       // unused
-    (void)size_b;       // unused
     (void)committed;    // unused
-    (void)arena_ind;    // unused
 
-    // TODO: add this function to the provider API to support Windows and USM
-    return true; // true means failure (unsupported)
+    jemalloc_memory_pool_t *pool = get_pool_by_arena_index(arena_ind);
+    assert(pool);
+    return umfMemoryProviderAllocationMerge(pool->provider, addr_a, addr_b,
+                                            size_a + size_b) !=
+           UMF_RESULT_SUCCESS;
 }
 
 // The extent_hooks_t structure comprises function pointers which are described individually below.
--- a/src/provider/provider_tracking.c
+++ b/src/provider/provider_tracking.c
@@ -9,6 +9,7 @@
 
 #include "provider_tracking.h"
 #include "critnib.h"
+#include "utils_concurrency.h"
 
 #include <umf/memory_pool.h>
 #include <umf/memory_provider.h>
@@ -16,6 +17,7 @@
 
 #include <assert.h>
 #include <errno.h>
+#include <stdio.h>
 #include <stdlib.h>
 
 typedef struct tracker_value_t {
@@ -32,7 +34,7 @@ static umf_result_t umfMemoryTrackerAdd(umf_memory_tracker_handle_t hTracker,
     value->pool = pool;
     value->size = size;
 
-    int ret = critnib_insert((critnib *)hTracker, (uintptr_t)ptr, value, 0);
+    int ret = critnib_insert(hTracker->map, (uintptr_t)ptr, value, 0);
 
     if (ret == 0) {
         return UMF_RESULT_SUCCESS;
@@ -59,7 +61,7 @@ static umf_result_t umfMemoryTrackerRemove(umf_memory_tracker_handle_t hTracker,
     // umfMemoryTrackerRemove call with the same ptr value.
     (void)size;
 
-    void *value = critnib_remove((critnib *)hTracker, (uintptr_t)ptr);
+    void *value = critnib_remove(hTracker->map, (uintptr_t)ptr);
     if (!value) {
         // This should not happen
         // TODO: add logging here
@@ -77,7 +79,7 @@ umfMemoryTrackerGetPool(umf_memory_tracker_handle_t hTracker, const void *ptr) {
 
     uintptr_t rkey;
     tracker_value_t *rvalue;
-    int found = critnib_find((critnib *)hTracker, (uintptr_t)ptr, FIND_LE,
+    int found = critnib_find(hTracker->map, (uintptr_t)ptr, FIND_LE,
                              (void *)&rkey, (void **)&rvalue);
     if (!found) {
         return NULL;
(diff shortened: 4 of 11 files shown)


src/provider/provider_tracking.c
```diff
@@ -119,6 +120,134 @@ static umf_result_t trackingAlloc(void *hProvider, size_t size,
     return ret;
 }
 
+static umf_result_t trackingAllocationSplit(void *provider, void *ptr,
```
pbalcer, line 123: Given that we use a tree for tracking, it might be simpler to avoid preemptively splitting/merging pages, and instead handle that during inserts/removals.

include/umf/memory_provider_ops.h
```diff
@@ -139,6 +139,11 @@ typedef struct umf_memory_provider_ops_t {
     /// @return pointer to a string containing the name of the \p provider
     ///
     const char *(*get_name)(void *provider);
+
+    umf_result_t (*allocation_split)(void *provider, void *ptr,
+                                     size_t totalSize, size_t leftSize);
+    umf_result_t (*allocation_merge)(void *provider, void *leftPtr,
```
pbalcer, line 145: you are making and assumption that "left" means "numerically lower address", right? I'd call it `lowPtr`, `highPtr`.

pbalcer (commented):
as for concurrency, I think if we don't add a lock we are creating a footgun. I can imagine a scenario where for example, split is called concurrently twice on the same block.

So unless, we add some mechanism for exclusive ownerships of ranges of memory, I think we should add a lock.

src/provider/provider_tracking.c
vinser52: it might be error-prone concerning IPC handle support. I prefer to do it explicitly when a pool manager calls merge/split functions (like implemented right now).
In general, it is an interesting case, the IPC handle is created per coarse-grain allocation (returned by the memory provider) and if two coarse-grain allocations are merged what should we do with corresponding IPC handles if them were created before? I will think more about that. I know that L0 just does not support split/merge functionality but tracking memory provider should work for all cases.

src/provider/provider_tracking.c
```diff
@@ -119,6 +120,134 @@ static umf_result_t trackingAlloc(void *hProvider, size_t size,
...
+                                            size_t totalSize, size_t leftSize) {
+    umf_tracking_memory_provider_t *p =
+        (umf_tracking_memory_provider_t *)provider;
+
+    tracker_value_t *splitValue =
```
vinser52, line 128: if you allocate `splitValue` just before `critnib_insert` then you do not need to `free(splitValue)` in every error cases below.

src/provider/provider_tracking.c
```diff
@@ -119,6 +120,134 @@ static umf_result_t trackingAlloc(void *hProvider, size_t size,
...
+    return UMF_RESULT_SUCCESS;
+}
+
+static umf_result_t trackingAllocationMerge(void *provider, void *leftPtr,
+                                            void *rightPtr, size_t totalSize) {
+    umf_tracking_memory_provider_t *p =
+        (umf_tracking_memory_provider_t *)provider;
+
+    tracker_value_t *mergedValue =
+        (tracker_value_t *)malloc(sizeof(tracker_value_t));
```
vinser52, line 188: if you allocate `mergedValue` just before `critnib_insert` then you do not need to `free(mergedValue)` in every error case below.

src/provider/provider_tracking.c
```diff
@@ -119,6 +120,134 @@ static umf_result_t trackingAlloc(void *hProvider, size_t size,
...
+    tracker_value_t *splitValue =
+        (tracker_value_t *)malloc(sizeof(tracker_value_t));
+    splitValue->pool = p->pool;
+    splitValue->size = leftSize;
+
+    tracker_value_t *value =
+        (tracker_value_t *)critnib_get((critnib *)p->hTracker, (uintptr_t)ptr);
+    if (!value) {
+        free(splitValue);
+        fprintf(stderr, "tracking split: no such value\n");
```
vinser52, line 137: I would add `TODO` comment that logger should be called instead of just `fprintf`

src/provider/provider_tracking.c
```diff
@@ -119,6 +120,134 @@ static umf_result_t trackingAlloc(void *hProvider, size_t size,
...
+        return UMF_RESULT_ERROR_INVALID_ARGUMENT;
+    }
+    if (value->size != totalSize) {
+        free(splitValue);
+        fprintf(stderr, "tracking split: %zu != %zu\n", value->size, totalSize);
```
vinser52, line 142: I would add `TODO` comment that logger should be called instead of just `fprintf`

src/provider/provider_tracking.c
```diff
@@ -119,6 +120,134 @@ static umf_result_t trackingAlloc(void *hProvider, size_t size,
...
+        return UMF_RESULT_ERROR_INVALID_ARGUMENT;
+    }
+
+    umf_result_t ret = umfMemoryProviderAllocationSplit(p->hUpstream, ptr,
+                                                        totalSize, leftSize);
+    if (ret != UMF_RESULT_SUCCESS) {
+        free(splitValue);
+        fprintf(stderr,
```
vinser52, line 150: I would add `TODO` comment that logger should be called instead of just `fprintf`

src/provider/provider_tracking.c
```diff
@@ -119,6 +120,134 @@ static umf_result_t trackingAlloc(void *hProvider, size_t size,
...
+
+    if (!mergedValue) {
+        return UMF_RESULT_ERROR_OUT_OF_HOST_MEMORY;
+    }
+
+    tracker_value_t *leftValue = (tracker_value_t *)critnib_get(
+        (critnib *)p->hTracker, (uintptr_t)leftPtr);
+    if (!leftValue) {
+        free(mergedValue);
+        fprintf(stderr, "tracking merge: no left value\n");
```
vinser52, line 200: I would add `TODO` comment that logger should be called instead of just `fprintf`

src/provider/provider_tracking.c
```diff
@@ -119,6 +120,134 @@ static umf_result_t trackingAlloc(void *hProvider, size_t size,
...
+        return UMF_RESULT_ERROR_INVALID_ARGUMENT;
+    }
+    tracker_value_t *rightValue = (tracker_value_t *)critnib_get(
+        (critnib *)p->hTracker, (uintptr_t)rightPtr);
+    if (!rightValue) {
+        free(mergedValue);
+        fprintf(stderr, "tracking merge: no right value\n");
```
vinser52, line 207: I would add `TODO` comment that logger should be called instead of just `fprintf`

src/provider/provider_tracking.c
```diff
@@ -119,6 +120,134 @@ static umf_result_t trackingAlloc(void *hProvider, size_t size,
...
+        return UMF_RESULT_ERROR_INVALID_ARGUMENT;
+    }
+    if (leftValue->pool != rightValue->pool) {
+        free(mergedValue);
+        fprintf(stderr, "tracking merge: pool mismatch\n");
```
vinser52, line 212: I would add `TODO` comment that logger should be called instead of just `fprintf`

src/provider/provider_tracking.c
```diff
@@ -119,6 +120,134 @@ static umf_result_t trackingAlloc(void *hProvider, size_t size,
...
+        return UMF_RESULT_ERROR_INVALID_ARGUMENT;
+    }
+    if (leftValue->size + rightValue->size != totalSize) {
+        // TODO: should we allow leftValue->size + rightValue->size > totalSize?
+        fprintf(stderr, "tracking merge: leftValue->size + rightValue->size != "
```
vinser52, line 217: I would add `TODO` comment that logger should be called instead of just `fprintf`

src/provider/provider_tracking.c
```diff
@@ -119,6 +120,134 @@ static umf_result_t trackingAlloc(void *hProvider, size_t size,
...
+        fprintf(stderr, "tracking merge: no right value\n");
+        return UMF_RESULT_ERROR_INVALID_ARGUMENT;
+    }
+    if (leftValue->pool != rightValue->pool) {
+        free(mergedValue);
+        fprintf(stderr, "tracking merge: pool mismatch\n");
+        return UMF_RESULT_ERROR_INVALID_ARGUMENT;
+    }
+    if (leftValue->size + rightValue->size != totalSize) {
+        // TODO: should we allow leftValue->size + rightValue->size > totalSize?
```
vinser52, line 216: How it is possible?

ldorau:
I confirm that the memory leak described in https://github.com/oneapi-src/unified-memory-framework/issues/147 does not occur with this change, because `arena_extent_split()` succeeds now.

kswiecicki:
> @kswiecicki could you verify if this solves all the memory leaks? On my machine, it seems to work fine after those changes.

It does, ASan no longer reports any leaks in jemalloc.

igchor:
> as for concurrency, I think if we don't add a lock we are creating a footgun. I can imagine a scenario where for example, split is called concurrently twice on the same block.
> 
> So unless we add some mechanism for exclusive ownerships of ranges of memory, I think we should add a lock.

Hm, should only add a mutex for split/merge calls or should we also protect trackingFree? I think freeing a memory region that is being split/merged can just be UB, right?

src/provider/provider_tracking.c
igchor: Well, someone (implementer of a pool) can call split with wrong arguments (size that does not match existing regions). It's a programmer error but the check is pretty cheap and I think quite useful.

src/provider/provider_tracking.c
igchor: Hm, do we really want to add "TODO"s to every fprintf call? I think we have a few without annotations already. When we will finally have a logging framework we'll do `grep fprintf` anyway I guess.

src/provider/provider_tracking.c
igchor: Yeah, I was thinking about this as well. Doing this lazily in removal would definitely be simpler (and probably faster) but as Sergey said, might be problematic for IPC. I think we can revisit this once we have IPC support merged.

include/umf/memory_provider_ops.h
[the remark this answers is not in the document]
ldorau: That's right :-)

ldorau:
This PR is required by #153

igchor:
> > as for concurrency, I think if we don't add a lock we are creating a footgun. I can imagine a scenario where for example, split is called concurrently twice on the same block.
> > So unless we add some mechanism for exclusive ownerships of ranges of memory, I think we should add a lock.
> 
> Hm, should only add a mutex for split/merge calls or should we also protect trackingFree? I think freeing a memory region that is being split/merged can just be UB, right?

I've added mutex for split/merge only for now.

src/provider/provider_tracking.c
```diff
@@ -119,6 +121,159 @@ static umf_result_t trackingAlloc(void *hProvider, size_t size,
...
+    ret = umfMemoryProviderAllocationSplit(p->hUpstream, ptr, totalSize,
+                                           firstSize);
+    if (ret != UMF_RESULT_SUCCESS) {
+        fprintf(stderr,
+                "tracking split: umfMemoryProviderAllocationSplit failed\n");
+        goto err;
+    }
+
+    void *highPtr = (void *)(((uintptr_t)ptr) + firstSize);
+    size_t rightSize = totalSize - firstSize;
```
ldorau, line 163: > size_t rightSize = totalSize - firstSize;

(`lowSize` and `highSize`) OR (`firstSize` and `secondSize`) OR (`leftSize` and `rightSize`) - do not mix them, please

src/pool/pool_jemalloc.c
```diff
@@ -210,15 +210,14 @@ static bool arena_extent_split(extent_hooks_t *extent_hooks, void *addr,
...
-    (void)size;         // unused
-    (void)size_a;       // unused
-    (void)size_b;       // unused
     (void)committed;    // unused
-    (void)arena_ind;    // unused
+    (void)size_b;
 
-    // TODO: add this function to the provider API to support Windows and USM
-    return true; // true means failure (unsupported)
+    assert(size_a + size_b == size);
```
bratpiorka, line 216: please assert that size, size_a and size_b > 0 and addr != NULL

src/provider/provider_tracking.c
```diff
@@ -119,6 +121,159 @@ static umf_result_t trackingAlloc(void *hProvider, size_t size,
...
 }
 
+static umf_result_t trackingAllocationSplit(void *provider, void *ptr,
+                                            size_t totalSize,
+                                            size_t firstSize) {
+    umf_result_t ret = UMF_RESULT_ERROR_UNKNOWN;
+    umf_tracking_memory_provider_t *p =
+        (umf_tracking_memory_provider_t *)provider;
+
+    tracker_value_t *splitValue =
```
bratpiorka, line 131: add a check for splitValue != NULL

src/provider/provider_tracking.c
```diff
@@ -119,6 +121,159 @@ static umf_result_t trackingAlloc(void *hProvider, size_t size,
...
+    if (ret != UMF_RESULT_SUCCESS) {
+        fprintf(stderr,
+                "tracking split: umfMemoryProviderAllocationSplit failed\n");
+        goto err;
+    }
+
+    void *highPtr = (void *)(((uintptr_t)ptr) + firstSize);
+    size_t rightSize = totalSize - firstSize;
+
+    // We'll have duplicate entry for the range [highPtr, rightValue->size] but this is fine,
```
ldorau, line 165: 1) `have duplicate` -> `have to duplicate` ?
2) `highPtr, rightValue` ? - maybe `highPtr` and `highValue` ?

src/provider/provider_tracking.c
```diff
@@ -119,6 +121,159 @@ static umf_result_t trackingAlloc(void *hProvider, size_t size,
     return ret;
 }
 
+static umf_result_t trackingAllocationSplit(void *provider, void *ptr,
+                                            size_t totalSize,
+                                            size_t firstSize) {
+    umf_result_t ret = UMF_RESULT_ERROR_UNKNOWN;
+    umf_tracking_memory_provider_t *p =
```
bratpiorka, line 128: could we stay with the full name? During the review I had to scroll up to see what is "p". If you want to keep the code short it would be better to add an alias for "tracker = provider->hTracker"

src/provider/provider_tracking.c
```diff
@@ -119,6 +121,159 @@ static umf_result_t trackingAlloc(void *hProvider, size_t size,
...
+                "tracking split: umfMemoryProviderAllocationSplit failed\n");
+        goto err;
+    }
+
+    void *highPtr = (void *)(((uintptr_t)ptr) + firstSize);
+    size_t rightSize = totalSize - firstSize;
+
+    // We'll have duplicate entry for the range [highPtr, rightValue->size] but this is fine,
+    // the value is the same anyway and we forbid splitting/removing that range concurrently
+    ret = umfMemoryTrackerAdd(p->hTracker, p->pool, highPtr, rightSize);
```
ldorau, line 167: `highPtr, rightSize` - mixed

src/provider/provider_tracking.c
```diff
@@ -119,6 +121,159 @@ static umf_result_t trackingAlloc(void *hProvider, size_t size,
...
+    if (ret != UMF_RESULT_SUCCESS) {
+        fprintf(stderr, "tracking split: umfMemoryTrackerAdd failed\n");
+        // TODO: what now? should we rollback the split? This is can only happen due to ENOMEM
```
ldorau, line 170: `This is can only happen` -> `This can only happen`

include/umf/memory_provider.h
```diff
@@ -156,6 +156,34 @@ umfMemoryProviderGetName(umf_memory_provider_handle_t hProvider);
 ///
 UMF_EXPORT umf_memory_provider_handle_t umfGetLastFailedMemoryProvider(void);
 
+///
+/// @brief Splits a coarse grain allocation into 2 adjacent allocations that
+///        can be managed (freed) separately.
+/// @param hProvider handle to the memory provider
+/// @param ptr pointer to the beginning of the allocation
+/// @param totalSize total size of the allocation to be split
```
bratpiorka, line 164: I see that in split/merge methods we don't want to have redundant arguments, so why we need "totalSize"? Shouldn't the totalSize be just the size of the alloc pointed by ptr? What if totalSize would be less/more than allocation?

include/umf/memory_provider.h
```diff
@@ -156,6 +156,34 @@ umfMemoryProviderGetName(umf_memory_provider_handle_t hProvider);
...
+umfMemoryProviderAllocationSplit(umf_memory_provider_handle_t hProvider,
+                                 void *ptr, size_t totalSize, size_t firstSize);
+
+///
+/// @brief Merges two coarse grain allocations into a single allocations that
+///        can be managed (freed) as a whole.
+/// @param hProvider handle to the memory provider
+/// @param lowPtr pointer to the first allocation
+/// @param highPtr pointer to the second allocation (should be > lowPtr)
+/// @param totalSize size of a new merged allocation. Should be equal
```
bratpiorka, line 179: why we need the totalSize argument?

src/provider/provider_tracking.c
```diff
@@ -119,6 +121,159 @@ static umf_result_t trackingAlloc(void *hProvider, size_t size,
...
+}
+
+static umf_result_t trackingAllocationMerge(void *provider, void *lowPtr,
+                                            void *highPtr, size_t totalSize) {
+    umf_result_t ret = UMF_RESULT_ERROR_UNKNOWN;
+    umf_tracking_memory_provider_t *p =
+        (umf_tracking_memory_provider_t *)provider;
+
+    tracker_value_t *mergedValue =
+        (tracker_value_t *)malloc(sizeof(tracker_value_t));
```
ldorau, line 202: handle malloc failure

src/provider/provider_tracking.c
```diff
@@ -119,6 +121,159 @@ static umf_result_t trackingAlloc(void *hProvider, size_t size,
...
+    mergedValue->pool = p->pool;
+    mergedValue->size = totalSize;
+
+    if (!mergedValue) {
```
ldorau, line 206: move that before the assignment

src/provider/provider_tracking.c
```diff
@@ -119,6 +121,159 @@ static umf_result_t trackingAlloc(void *hProvider, size_t size,
...
+    size_t rightSize = totalSize - firstSize;
+
+    // We'll have duplicate entry for the range [highPtr, rightValue->size] but this is fine,
+    // the value is the same anyway and we forbid splitting/removing that range concurrently
+    ret = umfMemoryTrackerAdd(p->hTracker, p->pool, highPtr, rightSize);
+    if (ret != UMF_RESULT_SUCCESS) {
+        fprintf(stderr, "tracking split: umfMemoryTrackerAdd failed\n");
+        // TODO: what now? should we rollback the split? This is can only happen due to ENOMEM
+        // so it's unlikely but probably the best solution would be to try to preallocate everything
+        // (value and critnib nodes) before calling umfMemoryProviderAllocationSplit.
```
bratpiorka, line 172: note: I added an issue for that

src/provider/provider_tracking.c
```diff
@@ -119,6 +121,159 @@ static umf_result_t trackingAlloc(void *hProvider, size_t size,
...
+        return UMF_RESULT_ERROR_OUT_OF_HOST_MEMORY;
+    }
+
+    int r = util_mutex_lock(p->hTracker->splitMergeMutex);
+    if (r) {
+        goto err_lock;
+    }
+
+    tracker_value_t *leftValue =
+        (tracker_value_t *)critnib_get(p->hTracker->map, (uintptr_t)lowPtr);
```
ldorau, line 216: `lowPtr` and `leftValue` - mixed names (low and high OR left and right)

src/provider/provider_tracking.c
```diff
@@ -119,6 +121,159 @@ static umf_result_t trackingAlloc(void *hProvider, size_t size,
...
+    if (!leftValue) {
+        fprintf(stderr, "tracking merge: no left value\n");
+        ret = UMF_RESULT_ERROR_INVALID_ARGUMENT;
+        goto err;
+    }
+    tracker_value_t *rightValue =
+        (tracker_value_t *)critnib_get(p->hTracker->map, (uintptr_t)highPtr);
```
ldorau, line 223: `highPtr` and `rightValue` - mixed

src/provider/provider_tracking.c
```diff
@@ -119,6 +121,159 @@ static umf_result_t trackingAlloc(void *hProvider, size_t size,
...
+
+    ret = umfMemoryProviderAllocationMerge(p->hUpstream, lowPtr, highPtr,
+                                           totalSize);
+    if (ret != UMF_RESULT_SUCCESS) {
+        fprintf(stderr,
+                "tracking merge: umfMemoryProviderAllocationMerge failed\n");
+        goto err;
+    }
+
+    // We'll have duplicate entry for the range [highPtr, rightValue->size] but this is fine,
```
ldorau, line 249: `have duplicate` -> `have to duplicate`

ldorau, line 249: `highPtr, rightValue` - mixed

src/provider/provider_tracking.c
```diff
@@ -119,6 +121,159 @@ static umf_result_t trackingAlloc(void *hProvider, size_t size,
...
+err_lock:
+    free(splitValue);
+    return ret;
+}
+
+static umf_result_t trackingAllocationMerge(void *provider, void *lowPtr,
+                                            void *highPtr, size_t totalSize) {
+    umf_result_t ret = UMF_RESULT_ERROR_UNKNOWN;
+    umf_tracking_memory_provider_t *p =
+        (umf_tracking_memory_provider_t *)provider;
```
bratpiorka, line 199: it would be better to do "tracker = provider->hTracker"

src/provider/provider_tracking.c
```diff
@@ -119,6 +121,159 @@ static umf_result_t trackingAlloc(void *hProvider, size_t size,
...
+    ret = umfMemoryProviderAllocationMerge(p->hUpstream, lowPtr, highPtr,
+                                           totalSize);
+    if (ret != UMF_RESULT_SUCCESS) {
+        fprintf(stderr,
+                "tracking merge: umfMemoryProviderAllocationMerge failed\n");
+        goto err;
+    }
+
+    // We'll have duplicate entry for the range [highPtr, rightValue->size] but this is fine,
+    // the value is the same anyway and we forbid splitting/removing that range concurrently
```
ldorau, line 250: splitting -> merging ?

src/provider/provider_tracking.c
```diff
@@ -119,6 +121,159 @@ static umf_result_t trackingAlloc(void *hProvider, size_t size,
...
+    return ret;
+}
+
+static umf_result_t trackingAllocationMerge(void *provider, void *lowPtr,
+                                            void *highPtr, size_t totalSize) {
+    umf_result_t ret = UMF_RESULT_ERROR_UNKNOWN;
+    umf_tracking_memory_provider_t *p =
+        (umf_tracking_memory_provider_t *)provider;
+
+    tracker_value_t *mergedValue =
```
bratpiorka, line 201: please check for "mergedValue != NULL"

src/provider/provider_tracking.h
```diff
@@ -14,14 +14,41 @@
...
+    }
+
+    handle->map = critnib_new();
+    handle->splitMergeMutex = util_mutex_create();
+
+    return handle;
+}
+
+static inline void umfMemoryTrackerDestroy(umf_memory_tracker_handle_t handle) {
+    critnib_delete(handle->map);
```
bratpiorka, line 47: add an assert + return if handle == NULL

bratpiorka (changes requested):

ldorau (changes requested):

src/provider/provider_tracking.c
vinser52: Make sense, but in case of merge do we really need to allocate a new element? You can just update the left record with the new size and remove the right record. Did I miss something?

src/pool/pool_jemalloc.c
```diff
@@ -210,15 +210,14 @@ static bool arena_extent_split(extent_hooks_t *extent_hooks, void *addr,
...
+
+    jemalloc_memory_pool_t *pool = get_pool_by_arena_index(arena_ind);
+    return umfMemoryProviderAllocationSplit(pool->provider, addr, size,
```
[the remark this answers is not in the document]
igchor: Done.

src/pool/pool_jemalloc.c
```diff
@@ -231,15 +230,12 @@ static bool arena_extent_merge(extent_hooks_t *extent_hooks, void *addr_a,
...
-    (void)addr_b;       // unused
-    (void)size_a;       // unused
-    (void)size_b;       // unused
     (void)committed;    // unused
-    (void)arena_ind;    // unused
 
-    // TODO: add this function to the provider API to support Windows and USM
-    return true; // true means failure (unsupported)
+    jemalloc_memory_pool_t *pool = get_pool_by_arena_index(arena_ind);
+    return umfMemoryProviderAllocationMerge(pool->provider, addr_a, addr_b,
```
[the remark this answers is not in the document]
igchor: Done.

src/provider/provider_tracking.c
igchor: Done.

src/provider/provider_tracking.c
igchor: See my other answer about updating the values. Without this alloc we would have to modify the value in-place which means we would have to use atomic instructions.

src/provider/provider_tracking.c
igchor: Just removing now, since split/merge is called under the lock.

src/provider/provider_tracking.h
igchor: Can't return since this is done in the lib destructor anyway and we cannot handle a failure there.

bratpiorka: Ok but if handle == NULL then we crash when we try to deref it. Please just add an early return for this case.

include/umf/memory_provider.h
vinser52: Here are my thoughts about this question:

1. We should not just copy jemalloc extent API because for me it looks jemalloc-specific. And it looks like we are not doing that.
2. Regarding `totalSize`. For me, it looks similar to `umfMemoryProviderFree(..., void *ptr, size_t size)` API that also requires passing the `size` of allocation pointed by `ptr`. But not all memory provider implementations need/use it: e.g. mmap implementation uses `size`, but L0 ignores it.
3. For me, bullet 2 means that UMF requires that the pool manager remember the size of each particular coarse-grain allocation but memory providers are not required to remember the size of each coarse-grain allocation. Since we already require to pass the `size` to `umfMemoryProviderFree` function what is the difference here? 
4. Regarding what if `totalSize != actual total size`, the same question is applicable for `umfMemoryProviderFree`. On our side we can do (at least debug) verification using the tracking memory provider - there is such information in the tracker.
5. Another question regarding input parameter validation, should the `firstSize` be a multiple of the page size?

bratpiorka: ok I agree with bullet 3

src/provider/provider_tracking.h
igchor: Right, good point. Done.

include/umf/memory_provider.h
igchor: Regarding 5: I think it is implied because I'm saying that the pointer needs to point to the beginning of the allocation and all allocations need to be page aligned.

bratpiorka (approved):

bratpiorka:
@vinser52 @pbalcer @ldorau all I need 2nd approve to merge this PR - could you check if anything is missing here?

ldorau (approved):

src/provider/provider_tracking.h
```diff
@@ -10,18 +10,61 @@
...
+
 #ifdef __cplusplus
 extern "C" {
 #endif
 
 extern umf_memory_provider_ops_t UMF_TRACKING_MEMORY_PROVIDER_OPS;
 
+struct umf_memory_tracker_t {
+    critnib *map;
+    os_mutex_t *splitMergeMutex;
```
vinser52, line 30: Global split/merge lock might kill scalability. As I understand the lock is needed only when concurrent split/merge occurs for the same memory buffer.

pbalcer: well, mmap is also fully synchronized, so I wouldn't worry that much. This shouldn't be a super common operation. I'd start with a simple mutex and then we can think about locking out ranges of memory if there's a perf problem.

src/provider/provider_tracking.c
```diff
@@ -119,6 +121,164 @@ static umf_result_t trackingAlloc(void *hProvider, size_t size,
...
+    void *highPtr = (void *)(((uintptr_t)ptr) + firstSize);
+    size_t secondSize = totalSize - firstSize;
+
+    // We'll have a duplicate entry for the range [highPtr, highValue->size] but this is fine,
+    // the value is the same anyway and we forbid removing that range concurrently
+    ret = umfMemoryTrackerAdd(provider->hTracker, provider->pool, highPtr,
+                              secondSize);
+    if (ret != UMF_RESULT_SUCCESS) {
+        fprintf(stderr, "tracking split: umfMemoryTrackerAdd failed\n");
+        // TODO: what now? should we rollback the split? This can only happen due to ENOMEM
```
pbalcer, line 175: yes, preallocation makes sense - should be doable once we have the base allocator.

PR #145, opened 2024-01-15
require specific versions of sphinxcontrib to gen docs

bratpiorka:
this PR fixes building HTML docs

### DIFF
--- a/third_party/requirements.txt
+++ b/third_party/requirements.txt
@@ -2,6 +2,11 @@
 # Formatting the source code
 clang-format==15.0.7
 # Generating HTML documentation
+sphinxcontrib_applehelp==1.0.4
+sphinxcontrib_devhelp==1.0.2
+sphinxcontrib_htmlhelp==2.0.1
+sphinxcontrib_serializinghtml==1.1.5
+sphinxcontrib_qthelp==1.0.3
 breathe==4.35.0
 sphinx==4.5.0
 sphinx_book_theme==0.3.3


ldorau (approved):

PatKamin:
Please, provide a link to the html docs preview.

lukaszstolarczuk (commented):
Do you know why did it stop working exactly...? It looks like it was out of the blue, after Igor's change...

My concern is, that bumping Sphinx version may lead to stop working for other OSes (like Rocky Linux or OpenSUSE - that was my experience with dockers on UR). I'm not sure how much you care about it 😉

third_party/requirements.txt
```diff
@@ -2,6 +2,6 @@
 # Formatting the source code
 clang-format==15.0.7
 # Generating HTML documentation
-breathe==4.35.0
-sphinx==4.5.0
-sphinx_book_theme==0.3.3
+breathe>=4.35.0
+sphinx>=5.0.0
+sphinx_book_theme>=0.3.3
```
lukaszstolarczuk, line 7: Please, update the commit message with some details why do we need this change.

bratpiorka:
> Please, provide a link to the html docs preview.

@PatKamin when I require specific versions of sphinx contrib I fix the building while the layout doesn't change, see: https://bratpiorka.github.io/unified-memory-framework/api.html#memory-provider

third_party/requirements.txt
bratpiorka: I changed my PR because I wanted to keep the current layout

bratpiorka:
> Do you know why did it stop working exactly...? It looks like it was out of the blue, after Igor's change...
> 
> My concern is, that bumping Sphinx version may lead to stop working for other OSes (like Rocky Linux or OpenSUSE - that was my experience with dockers on UR). I'm not sure how much you care about it 😉
.
@lukaszstolarczuk I'm not sure... the only thing I saw is the different version of github image provisioner. Maybe Python requirements were searched and installed in a different order? Anyway, I kept the current version of Sphinx

lukaszstolarczuk (approved):

PR #146, opened 2024-01-15
Tracking provider cannot change behavior of upstream provider

ldorau:
Tracking provider cannot change behavior of the upstream provider. Tracking provider should be transparent for the upstream provider.

### DIFF
--- a/src/provider/provider_tracking.c
+++ b/src/provider/provider_tracking.c
@@ -109,11 +109,11 @@ static umf_result_t trackingAlloc(void *hProvider, size_t size,
         return ret;
     }
 
-    ret = umfMemoryTrackerAdd(p->hTracker, p->pool, *ptr, size);
-    if (ret != UMF_RESULT_SUCCESS && p->hUpstream) {
-        if (umfMemoryProviderFree(p->hUpstream, *ptr, size)) {
-            // TODO: LOG
-        }
+    umf_result_t ret2 = umfMemoryTrackerAdd(p->hTracker, p->pool, *ptr, size);
+    if (ret2 != UMF_RESULT_SUCCESS) {
+        // DO NOT call umfMemoryProviderFree() here, because the tracking provider
+        // cannot change behaviour of the upstream provider.
+        // TODO: LOG
     }
 
     return ret;
@@ -131,7 +131,9 @@ static umf_result_t trackingFree(void *hProvider, void *ptr, size_t size) {
     if (ptr) {
         ret = umfMemoryTrackerRemove(p->hTracker, ptr, size);
         if (ret != UMF_RESULT_SUCCESS) {
-            return ret;
+            // DO NOT return an error here, because the tracking provider
+            // cannot change behaviour of the upstream provider.
+            // TODO: LOG
         }
     }
 


src/provider/provider_tracking.c
```diff
@@ -131,7 +131,9 @@ static umf_result_t trackingFree(void *hProvider, void *ptr, size_t size) {
     if (ptr) {
         ret = umfMemoryTrackerRemove(p->hTracker, ptr, size);
         if (ret != UMF_RESULT_SUCCESS) {
-            return ret;
+            // DO NOT return an error here, because the tracking provider
```
bratpiorka, line 134: Totally agree

bratpiorka (approved):

src/provider/provider_tracking.c
vinser52, line 134: I looked into `umfMemoryTrackerRemove` implementation and it can fail only if the state of the memory tracker is invalid. So probably we should add assert here and just crash.

vinser52:
Could you please also fix the `trackingAlloc` function and align it with this PR. This part looks strange:
```
ret = umfMemoryTrackerAdd(p->hTracker, p->pool, *ptr, size);
    if (ret != UMF_RESULT_SUCCESS && p->hUpstream) {
        if (umfMemoryProviderFree(p->hUpstream, *ptr, size)) {
            // TODO: LOG
        }
    }
```
if we are not able to add new block to the tracker we should not call `umfMemoryProviderFree`, right?

src/provider/provider_tracking.c
ldorau: See: https://github.com/oneapi-src/unified-memory-framework/pull/141#issuecomment-1890259875
"It looks like older version of jemalloc called `arena_extent_dalloc()` with an address that was not coming from any previous `arena_extend_alloc()` for some reason."

With assert here and https://github.com/oneapi-src/unified-memory-framework/pull/141 merged all jemalloc_pool test would crash here, so it is not a good idea. 
We could add assert inside a debug ifdef, as a debug feature only .

ldorau:
> Could you please also fix the `trackingAlloc` function and align it with this PR. This part looks strange:
> if we are not able to add new block to the tracker we should not call `umfMemoryProviderFree`, right?

Right. Done.

src/provider/provider_tracking.c
vinser52: If `arena_extent_dalloc()` is called with an address that was not coming from any previous `arena_extend_alloc()` can we fix the memory tracker map? My point is that we should keep data in the tracker map in a valid state otherwise, the tracker is in the invalid state. I asked clarify question in [#141 (comment)](https://github.com/oneapi-src/unified-memory-framework/pull/141#issuecomment-1893582613)

src/provider/provider_tracking.c
```diff
@@ -110,10 +110,10 @@ static umf_result_t trackingAlloc(void *hProvider, size_t size,
     }
 
     ret = umfMemoryTrackerAdd(p->hTracker, p->pool, *ptr, size);
-    if (ret != UMF_RESULT_SUCCESS && p->hUpstream) {
-        if (umfMemoryProviderFree(p->hUpstream, *ptr, size)) {
-            // TODO: LOG
-        }
+    if (ret != UMF_RESULT_SUCCESS) {
```
vinser52, line 113: If allocation succeeds but `umfMemoryTrackerAdd` fails the `ret` will contain an error code even if the upstream memory provider returned `UMF_RESULT_SUCCESS`. The same applies to the free path below.

ldorau: Done. The free path below is OK.

src/provider/provider_tracking.c
ldorau: jemalloc sometimes calls `arena_extend_dalloc()` for an address that does not come from any previous `arena_extend_alloc()` call. How can we fix that? I think the memory tracker map is OK and it does not need to be fixed. It is just an issue in jemalloc.

vinser52: my guess was that if jemalloc calls `arena_extend_dalloc()` and passes a sub-range of previously allocated and added to the tracker region then we can fix the tracker state by removing this sub-range from the tracker. But if the completely different region is deleted with `arena_extend_dalloc()` then I agree with you - nothing can be done on our side.

vinser52 (approved):

src/provider/provider_tracking.c
```diff
@@ -131,7 +131,9 @@ static umf_result_t trackingFree(void *hProvider, void *ptr, size_t size) {
     if (ptr) {
         ret = umfMemoryTrackerRemove(p->hTracker, ptr, size);
```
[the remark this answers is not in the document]
ldorau: > Actually, I think we can just make umfMemorTrackerRemove return void. We already have a comment inside it that the failure 'should not happen'. We can just make this check an assert:

With assert and https://github.com/oneapi-src/unified-memory-framework/pull/141 merged all CI builds will fail.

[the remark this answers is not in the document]
bratpiorka: @igchor I think we should leave this as it is, with returning umf_result_t. I've talked to Lukasz and we have an idea of how to better WA the problems with jemalloc. However, let's solve this problem by starting with a split/merge implementation and then adding dalloc.

src/provider/provider_tracking.c
```diff
@@ -109,11 +109,11 @@ static umf_result_t trackingAlloc(void *hProvider, size_t size,
         return ret;
     }
 
-    ret = umfMemoryTrackerAdd(p->hTracker, p->pool, *ptr, size);
-    if (ret != UMF_RESULT_SUCCESS && p->hUpstream) {
-        if (umfMemoryProviderFree(p->hUpstream, *ptr, size)) {
-            // TODO: LOG
-        }
+    umf_result_t ret2 = umfMemoryTrackerAdd(p->hTracker, p->pool, *ptr, size);
```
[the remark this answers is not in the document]
bratpiorka: I added the issue #151 for this. I also added a task for optimizing mallocs (UMFW-218).

igchor (approved):

PR #152, opened 2024-01-17
Use UMF_NUMA_MODE_BIND as the default NUMA mode in memory_target_numa.c

ldorau:
Use UMF_NUMA_MODE_BIND instead of UMF_NUMA_MODE_DEFAULT as the default NUMA mode in memory_target_numa.c, because for UMF_NUMA_MODE_DEFAULT the nodemask and maxnode arguments would have to specify the empty set of nodes. With UMF_NUMA_MODE_DEFAULT and non-empty the nodemask and maxnode arguments the bind() function fails with EINVAL.

Fixes: #149


ldorau:
Postponed until 2-nodes machine is available in the CI

PR #153, opened 2024-01-17
Assert if tracking provider is not empty

ldorau:


bratpiorka:
great job with finding this problems :)

igchor:
Hm, I'm not sure if this is correct. We have a single critnib (hTracker) for the entire library and tracking provider is per-pool (trackingFinalize will be called on each umfPoolDestroy). 

This means that if we have multiple pools opened (and with active allocations), this assert will fail, even, if we are destroying a pool that did all necessary cleanup.

bratpiorka:
@igchor ok and what about moving this code to tracker destructor? providerFini() and deleteLibTracker()

igchor:
> @igchor ok and what about moving this code to tracker destructor? providerFini() and deleteLibTracker()

Yes, that should work.

But I think checking the leaks per each memory pool would be also good.

We can implement a provider in tests, that will be a wrapper and will forward all allocs/free to the upstream provider + will count all allocs and frees. If the counter is not 0 at the end, report an error. You can see here, how you can implement a custom provider easily: https://github.com/oneapi-src/unified-memory-framework/blob/main/test/memoryPoolAPI.cpp#L205

ldorau:
@igchor checking the leaks per each memory pool done :-)

igchor:
> @igchor checking the leaks per each memory pool done :-)

Looks good. I would just put those checks (especially the iteration in trackingFinalize) under some define (maybe just `#ifndef NDEBUG` or some custom one which we would set in tests)  We don't want to do those checks in a production version as there might be thousands / millions of entries in the critnib.

ldorau:
> #ifndef NDEBUG

@igchor  Done.

igchor (approved):

ldorau:
See a version with all required fixes: https://github.com/oneapi-src/unified-memory-framework/pull/155

ldorau:
Requires the following PRs to be merged in order to pass all CI builds (see #155):
- [ ] #141 
- [ ] #140

vinser52 (approved):

bratpiorka:
@ldorau is this PR ready to review? Could you rebase and re-run checks?

ldorau:
> @ldorau is this PR ready to review? Could you rebase and re-run checks?

Rebased and checks re-run, but not all of them pass yet :-(

src/provider/provider_tracking.c
```diff
@@ -404,3 +434,27 @@ void umfTrackingMemoryProviderGetUpstreamProvider(
         (umf_tracking_memory_provider_t *)hTrackingProvider;
     *hUpstream = p->hUpstream;
 }
+
+void umfTrackingMemoryProviderFini(void *tracker) {
+#ifndef NDEBUG
+    uintptr_t rkey;
```
bratpiorka, line 440: you repeat the same code - could you create a function for that?

ldorau: Done

bratpiorka (approved):

PR #160, opened 2024-01-18
Add OS memory provider config tests

PatKamin:
Tests implemented for a single NUMA node:
- tests for protection flags except the UMF_PROTECTION_EXEC flag,
- tests for numa modes except unsupported values,
- tests for visibility.

Tests for numa flags are not part of this PR, I think they should be implemented for multiple numa nodes systems.


.github/workflows/basic.yml
```diff
@@ -103,7 +103,7 @@ jobs:
       - name: Run tests
         working-directory: ${{env.BUILD_DIR}}
         run: |
-          ctest --output-on-failure
+          ctest --output-on-failure --verbose
```
ldorau, line 106: Why `--verbose` ? `--output-on-failure` is enough - it prints the verbose output when a test fails.

ldorau (changes requested):

test/provider_os_memory_config.cpp
```diff
@@ -0,0 +1,192 @@
...
+        /* .maxnode = */ 0,
+        /* .numa_mode = */ UMF_NUMA_MODE_DEFAULT,
+        /* .numa_flags = */ 0,
+
+        // others
+        /* .traces = */ 1,
+    };
+    size_t size = 128;
+    void *ptr = nullptr;
+    char *dest = nullptr;
```
igchor, line 36: any reason why this isn't a std::string or something similar?

test/provider_os_memory_config.cpp
```diff
@@ -0,0 +1,192 @@
...
+    // pages may not be accessed - PROT_NONE
+    params.protection = UMF_PROTECTION_NONE;
+    create_provider(&params);
+    allocate_memory();
+
+    // read failure
+    EXPECT_DEATH(copy_ptr_data(), "");
+
+    // write failure
+    EXPECT_DEATH(memset(ptr, 'A', size), "");
```
bratpiorka, line 95: will this line be executed after the failure for copy?

test/provider_os_memory_config.cpp
```diff
@@ -0,0 +1,192 @@
...
+    allocate_memory();
+
+    // read success
+    read_memory();
+
+    // write failure
+    EXPECT_DEATH(memset(ptr, 'A', size), "");
+}
+
+TEST_F(providerConfigTest, protection_flag_write) {
```
igchor, line 111: perhaps also test a combination of read and write

test/provider_os_memory_config.cpp
PatKamin: Yes, after the expected read failure, this line will be executed with expected write failure. Read or write success will fail this test case.

test/provider_os_memory_config.cpp
PatKamin: Done

bratpiorka (approved):

test/provider_os_memory_config.cpp
```diff
@@ -0,0 +1,202 @@
...
+
+    void allocate_memory() {
+        auto res =
+            umfMemoryProviderAlloc(provider, size, os_get_page_size(), &ptr);
+        ASSERT_EQ(res, UMF_RESULT_SUCCESS);
+        ASSERT_NE(ptr, nullptr);
+    }
+
+    void read_memory() {
+        dest.reserve(size);
```
igchor, line 66: this only reserves memory but does not change the size of the string, you should just call:
`dest.assing(ptr, size)` here.

test/provider_os_memory_config.cpp
```diff
@@ -0,0 +1,202 @@
...
+        memcpy(dest.data(), ptr, size);
+        for (int i = 0; i < size; ++i) {
```
igchor, line 68: nit: I think you can use memcmp or something similar here.

test/provider_os_memory_config.cpp
```diff
@@ -0,0 +1,202 @@
...
+            EXPECT_EQ(*(static_cast<char *>(ptr) + i), *(dest.data() + i));
+        }
+    }
+
+    void write_memory(std::string_view str) {
+        memset(ptr, '\0', size);
+        str.copy((char *)ptr, str.size());
+        EXPECT_EQ(std::string_view((char *)ptr), str);
```
igchor, line 76: I think you should initialize the string_view using 'size' as well. Otherwise, the view will contain 0 elements (the ctor that takes only a ptr, stops after finding a NULL terminator).

test/provider_os_memory_config.cpp
PatKamin: Thanks, done.

test/provider_os_memory_config.cpp
PatKamin: Done

test/provider_os_memory_config.cpp
PatKamin: But the ptr's data should already begin with the string passed to this function after `str.copy()`. And the string will be null terminated. I think it's sufficient to check just the beginning of the allocated memory to see if write is even possible on this page.

igchor: Of course, you are right.

igchor (approved):

bratpiorka (approved):

ldorau (approved):

Issue #162, opened 2024-01-18
Add missing checks for provider/pool handles in API functions


igchor:
We should verify that handles passed to umfMemoryProvider* and umfPool* functions are not NULL.

PR #163, opened 2024-01-18
Add checks for pool and provider handles

igchor:
closes https://github.com/oneapi-src/unified-memory-framework/issues/162

### DIFF
--- a/src/critnib/critnib.c
+++ b/src/critnib/critnib.c
@@ -58,6 +58,7 @@
 #include <stddef.h>
 
 #include "critnib.h"
+#include "utils_common.h"
 #include "utils_concurrency.h"
 
 /*
--- a/src/memory_provider.c
+++ b/src/memory_provider.c
@@ -8,6 +8,8 @@
  */
 
 #include "memory_provider_internal.h"
+#include "utils_common.h"
+
 #include <umf/memory_provider.h>
 
 #include <assert.h>
@@ -64,6 +66,7 @@ checkErrorAndSetLastProvider(umf_result_t result,
 
 umf_result_t umfMemoryProviderAlloc(umf_memory_provider_handle_t hProvider,
                                     size_t size, size_t alignment, void **ptr) {
+    UMF_CHECK((hProvider != NULL), UMF_RESULT_ERROR_INVALID_ARGUMENT);
     umf_result_t res =
         hProvider->ops.alloc(hProvider->provider_priv, size, alignment, ptr);
     checkErrorAndSetLastProvider(res, hProvider);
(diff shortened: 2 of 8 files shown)


src/memory_provider.c
```diff
@@ -80,17 +84,20 @@ umf_result_t umfMemoryProviderFree(umf_memory_provider_handle_t hProvider,
 void umfMemoryProviderGetLastNativeError(umf_memory_provider_handle_t hProvider,
                                          const char **ppMessage,
                                          int32_t *pError) {
+    ASSERT(hProvider != NULL);
```
bratpiorka, line 87: UMF_CHECK?

test/memoryPoolAPI.cpp
```diff
@@ -282,3 +285,37 @@ TEST_F(test, getLastFailedMemoryProvider) {
                   umfMemoryProviderGetName(umfGetLastFailedMemoryProvider())),
               "provider2");
 }
+
+struct poolHandleCheck
+    : umf_test::test,
+      ::testing::WithParamInterface<
+          std::function<std::variant<void *, umf_result_t, size_t>(void)>> {};
+
+TEST_P(poolHandleCheck, poolHandleCheckAll) {
```
bratpiorka, line 294: Could you add few comments on how this works? For someone with a "moderate" level of C++ knowledge? :) You are testing here the arguments of the functions or the result?

bratpiorka (changes requested):

src/critnib/critnib.c
```diff
@@ -58,6 +58,7 @@
 #include <stddef.h>
 
 #include "critnib.h"
+#include "utils_common.h"
```
vinser52, line 61: You added include but looks like do not use anything in this src file.

src/memory_provider.c
igchor: I cannot use it here since this function returns void.

src/critnib/critnib.c
igchor: I moved some misc helpers from concurrency.h to common.h that critnib uses.

test/memoryPoolAPI.cpp
igchor: Done. I've added some comments.

A longer explanation:
The idea here is to just verify if calling each of the functions listed below (umfPoolMalloc, umfPoolFree, ...) with NULL handle returns an appropriate error (NULL in case of a function returning void*, INVALID_ARGUMENT for functions returning status). But there is a problem: since those functions have different return types and different parameters we cannot just instantiate this TEST_P with function pointers directly. We need to have a single function pointer type. To do that I unify the return type using std::variant and `withGeneratedArgs` to unify parameters (which basically just sets each argument to NULL/zero). Now, we have a single function pointer signature (`std::variant<void *, umf_result_t, size_t>(void)`) and we can bind it to each umf function we are interested in.

I actually want to use a similar approach for testing the parameter validation in general where we could just define what combination or arguments we want to test.

bratpiorka (approved):

KFilipek (approved):

vinser52 (approved):

PR #164, opened 2024-01-19
extend UMF_DEVELOPER_MODE compiler params

bratpiorka:
requires #165

### DIFF
--- a/cmake/helpers.cmake
+++ b/cmake/helpers.cmake
@@ -36,6 +36,11 @@ function(add_umf_target_compile_options name)
         if(UMF_DEVELOPER_MODE)
             target_compile_options(${name} PRIVATE
                 -Werror
+                -Wextra
+                -Wshadow
+                -Wno-implicit-fallthrough
+                -Wformat
+                -Wformat-security
                 -fno-omit-frame-pointer
                 -fstack-protector-strong
             )
--- a/src/pool/pool_disjoint.cpp
+++ b/src/pool/pool_disjoint.cpp
@@ -313,7 +313,7 @@ class DisjointPool::AllocImpl {
     std::vector<std::unique_ptr<Bucket>> Buckets;
 
     // Configuration for this instance
-    umf_disjoint_pool_params_t params;
+    umf_disjoint_pool_params_t Params;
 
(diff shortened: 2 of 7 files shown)


cmake/helpers.cmake
```diff
@@ -36,6 +36,11 @@ function(add_umf_target_compile_options name)
         if(UMF_DEVELOPER_MODE)
             target_compile_options(${name} PRIVATE
                 -Werror
+                -Wextra
+                -Wshadow
+                -Wno-implicit-fallthrough
```
PatKamin, line 41: Do we use a fallthrough switch statement somewhere? I think the code is cleaner with explicit breaks in switches.

cmake/helpers.cmake
```diff
@@ -36,6 +36,11 @@ function(add_umf_target_compile_options name)
         if(UMF_DEVELOPER_MODE)
             target_compile_options(${name} PRIVATE
                 -Werror
+                -Wextra
+                -Wshadow
```
igchor, line 40: I don't think we should use `-Wshadow` (especially for C++) where it's perfectly fine to have class/struct members named the same way as e.g. function params.

bratpiorka: oneCCL enables this flag during compilation and if they did this and compiled UMF, they would get errors

cmake/helpers.cmake
bratpiorka: no, but oneCCL enables this flag so I would leave it but as you suggest - we'll probably use [[fallthrough]] attribute for such cases

cmake/helpers.cmake
igchor: Ok, then perhaps we should add some comments why are we adding those flags?

PatKamin: I wonder, shouldn't it be the third party repo's problem that it enforces some compiler flags globally on included repositories such as UMF? In the end, we can't control all repos and their flags.

That's the reason why we use modified versions of `add_executable` and `add_library`. By doing this, we don't enforce flags on eventual subprojects.

bratpiorka:
all - I'll try to change cmake options at oneCCL side instead of changing our flags. If I got problems I will re-open this PR

PR #165, opened 2024-01-22
fix for shadowed declaration error

bratpiorka:
fixes for shadowed decl errors when compiled with -Wshadow flag

### DIFF
--- a/src/pool/pool_disjoint.cpp
+++ b/src/pool/pool_disjoint.cpp
@@ -313,7 +313,7 @@ class DisjointPool::AllocImpl {
     std::vector<std::unique_ptr<Bucket>> Buckets;
 
     // Configuration for this instance
-    umf_disjoint_pool_params_t params;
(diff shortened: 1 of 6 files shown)


ldorau (changes requested):
Please describe what problem you want to solve here (I do not understand why this change is needed - it should be explained in the commit message).

igchor:
> Please describe what problem you want to solve here (I do not understand why this change is needed - it should be explained in the commit message).

I guess this is to fix  `-Wshadow` warning but I don't think this flag makes much sense for C++ tbh.

bratpiorka:
@ldorau I updated the commit message

lukaszstolarczuk (commented):
:+1:

igchor (approved):

bratpiorka:
all - I'll try to change cmake options at oneCCL side instead of changing our flags. If I got problems I will re-open this PR

PR #166, opened 2024-01-22
Add base allocator

ldorau:
Add a MT-safe fixed-size-pool base allocator.

### DIFF
--- a/.github/workflows/.spellcheck-conf.toml
+++ b/.github/workflows/.spellcheck-conf.toml
@@ -1,3 +1,3 @@
 [default]
-# Don't correct the "ASSERTne" word
-extend-ignore-words-re = ["ASSER", "Tne"]
+# Don't correct the following words:
+extend-ignore-words-re = ["ASSER", "Tne", "ba", "BA"]
--- a/src/base_alloc/base_alloc.c
+++ b/src/base_alloc/base_alloc.c
@@ -1,0 +1,230 @@
+/*
+ * Copyright (C) 2024 Intel Corporation
+ *
+ * Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+*/
+
+#include <assert.h>
+
+#include "base_alloc.h"
+#include "base_alloc_internal.h"
+#include "utils_common.h"
+#include "utils_concurrency.h"
+
+// minimum size of a single pool of the base allocator
+#define MINIMUM_POOL_SIZE (ba_os_get_page_size())
+
+// minimum number of chunks used to calculate the size of pools
+#define MINIMUM_CHUNK_COUNT (128)
+
+// alignment of the base allocator
+#define MEMORY_ALIGNMENT (8)
+
+typedef struct umf_ba_chunk_t umf_ba_chunk_t;
+typedef struct umf_ba_next_pool_t umf_ba_next_pool_t;
+
+// memory chunk of size 'chunk_size'
+struct umf_ba_chunk_t {
+    umf_ba_chunk_t *next;
+    char user_data[];
+};
+
+// metadata is set and used only in the main (the first) pool
+struct umf_ba_main_pool_meta_t {
+    size_t pool_size; // size of each pool (argument of each ba_os_alloc() call)
+    size_t chunk_size;         // size of all memory chunks in this pool
+    os_mutex_t *free_lock;     // lock of free_list
+    umf_ba_chunk_t *free_list; // list of free chunks
+#ifndef NDEBUG
+    size_t n_pools;
+    size_t n_allocs;
+    size_t n_chunks;
+#endif /* NDEBUG */
+};
+
+// the main pool of the base allocator (there is only one such pool)
+struct umf_ba_pool_t {
+    // address of the beginning of the next pool (a list of allocated pools to be freed in umf_ba_destroy())
+    umf_ba_next_pool_t *next_pool;
+
+    // metadata is set and used only in the main (the first) pool
+    struct umf_ba_main_pool_meta_t metadata;
+
+    // data area of the main pool (the first one) starts here
+    char data[];
+};
+
+// the "next" pools of the base allocator (pools allocated later, when we run out of the memory of the main pool)
+struct umf_ba_next_pool_t {
+    // address of the beginning of the next pool (a list of allocated pools to be freed in umf_ba_destroy())
+    umf_ba_next_pool_t *next_pool;
+
+    // data area of all pools except of the main (the first one) starts here
+    char data[];
+};
+
+#ifndef NDEBUG
+static void ba_debug_checks(umf_ba_pool_t *pool) {
+    // count pools
+    size_t n_pools = 1;
+    umf_ba_next_pool_t *next_pool = pool->next_pool;
+    while (next_pool) {
+        n_pools++;
+        next_pool = next_pool->next_pool;
+    }
+    assert(n_pools == pool->metadata.n_pools);
+
+    // count chunks
+    size_t n_free_chunks = 0;
+    umf_ba_chunk_t *next_chunk = pool->metadata.free_list;
+    while (next_chunk) {
+        n_free_chunks++;
+        next_chunk = next_chunk->next;
+    }
+    assert(n_free_chunks == pool->metadata.n_chunks - pool->metadata.n_allocs);
+}
+#endif /* NDEBUG */
+
+// ba_divide_memory_into_chunks - divide given memory into chunks of chunk_size and add them to the free_list
+static void ba_divide_memory_into_chunks(umf_ba_pool_t *pool, void *ptr,
+                                         size_t size) {
+    assert(pool->metadata.free_list == NULL);
+    assert(size > pool->metadata.chunk_size);
+
+    char *data_ptr = ptr;
+    size_t size_left = size;
+
+    umf_ba_chunk_t *current_chunk = (umf_ba_chunk_t *)data_ptr;
+    umf_ba_chunk_t *prev_chunk = current_chunk;
+
+    while (size_left >= pool->metadata.chunk_size) {
+        current_chunk = (umf_ba_chunk_t *)data_ptr;
+        prev_chunk->next = current_chunk;
+
+        data_ptr += pool->metadata.chunk_size;
+        size_left -= pool->metadata.chunk_size;
+        prev_chunk = current_chunk;
+#ifndef NDEBUG
+        pool->metadata.n_chunks++;
+#endif /* NDEBUG */
+    }
+
+    current_chunk->next = NULL;
+    pool->metadata.free_list = ptr; // address of the first chunk
+}
+
+umf_ba_pool_t *umf_ba_create(size_t size) {
+    size_t chunk_size = align_size(size, MEMORY_ALIGNMENT);
+    size_t mutex_size = align_size(util_mutex_get_size(), MEMORY_ALIGNMENT);
+
+    size_t metadata_size = sizeof(struct umf_ba_main_pool_meta_t);
+    size_t pool_size = sizeof(void *) + metadata_size + mutex_size +
+                       (MINIMUM_CHUNK_COUNT * chunk_size);
+    if (pool_size < MINIMUM_POOL_SIZE) {
+        pool_size = MINIMUM_POOL_SIZE;
+    }
+
+    pool_size = align_size(pool_size, ba_os_get_page_size());
+
+    umf_ba_pool_t *pool = (umf_ba_pool_t *)ba_os_alloc(pool_size);
+    if (!pool) {
+        return NULL;
+    }
+
+    pool->metadata.pool_size = pool_size;
+    pool->metadata.chunk_size = chunk_size;
+    pool->next_pool = NULL; // this is the only pool now
+#ifndef NDEBUG
+    pool->metadata.n_pools = 1;
+    pool->metadata.n_allocs = 0;
+    pool->metadata.n_chunks = 0;
+#endif /* NDEBUG */
+
+    char *data_ptr = (char *)&pool->data;
+    size_t size_left = pool_size - offsetof(umf_ba_pool_t, data);
+
+    // allocate and init free_lock
+    pool->metadata.free_lock = util_mutex_init(data_ptr);
+    if (!pool->metadata.free_lock) {
+        ba_os_free(pool, pool_size);
+        return NULL;
+    }
+
+    data_ptr += mutex_size;  // free_lock is here
+    size_left -= mutex_size; // for free_lock
+
+    pool->metadata.free_list = NULL;
+    ba_divide_memory_into_chunks(pool, data_ptr, size_left);
+
+    return pool;
+}
+
+void *umf_ba_alloc(umf_ba_pool_t *pool) {
+    util_mutex_lock(pool->metadata.free_lock);
+    if (pool->metadata.free_list == NULL) {
+        umf_ba_next_pool_t *new_pool =
+            (umf_ba_next_pool_t *)ba_os_alloc(pool->metadata.pool_size);
+        if (!new_pool) {
+            util_mutex_unlock(pool->metadata.free_lock);
+            return NULL;
+        }
+
+        // add the new pool to the list of pools
+        new_pool->next_pool = pool->next_pool;
+        pool->next_pool = new_pool;
+
+#ifndef NDEBUG
+        pool->metadata.n_pools++;
+#endif /* NDEBUG */
+
+        size_t size =
+            pool->metadata.pool_size - offsetof(umf_ba_next_pool_t, data);
+        ba_divide_memory_into_chunks(pool, &new_pool->data, size);
+    }
+
+    umf_ba_chunk_t *chunk = pool->metadata.free_list;
+    pool->metadata.free_list = pool->metadata.free_list->next;
+#ifndef NDEBUG
+    pool->metadata.n_allocs++;
+    ba_debug_checks(pool);
+#endif /* NDEBUG */
+    util_mutex_unlock(pool->metadata.free_lock);
+
+    return chunk;
+}
+
+void umf_ba_free(umf_ba_pool_t *pool, void *ptr) {
+    if (ptr == NULL) {
+        return;
+    }
+
+    umf_ba_chunk_t *chunk = (umf_ba_chunk_t *)ptr;
+
+    util_mutex_lock(pool->metadata.free_lock);
+    chunk->next = pool->metadata.free_list;
+    pool->metadata.free_list = chunk;
+#ifndef NDEBUG
+    pool->metadata.n_allocs--;
+    ba_debug_checks(pool);
+#endif /* NDEBUG */
+    util_mutex_unlock(pool->metadata.free_lock);
+}
+
+void umf_ba_destroy(umf_ba_pool_t *pool) {
+#ifndef NDEBUG
+    assert(pool->metadata.n_allocs == 0);
+    ba_debug_checks(pool);
+#endif /* NDEBUG */
+    size_t size = pool->metadata.pool_size;
+    umf_ba_next_pool_t *current_pool;
+    umf_ba_next_pool_t *next_pool = pool->next_pool;
+    while (next_pool) {
+        current_pool = next_pool;
+        next_pool = next_pool->next_pool;
+        ba_os_free(current_pool, size);
+    }
+
+    util_mutex_destroy_not_free(pool->metadata.free_lock);
+    ba_os_free(pool, size);
+}
--- a/src/base_alloc/base_alloc.h
+++ b/src/base_alloc/base_alloc.h
@@ -1,0 +1,22 @@
+/*
+ * Copyright (C) 2024 Intel Corporation
+ *
+ * Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+*/
+
+/* A MT-safe fixed-size-pool base allocator */
+
+#ifndef UMF_BASE_ALLOC_H
+#define UMF_BASE_ALLOC_H 1
+
+#include <stddef.h>
+
+typedef struct umf_ba_pool_t umf_ba_pool_t;
+
+umf_ba_pool_t *umf_ba_create(size_t size);
+void *umf_ba_alloc(umf_ba_pool_t *pool);
+void umf_ba_free(umf_ba_pool_t *pool, void *ptr);
+void umf_ba_destroy(umf_ba_pool_t *pool);
+
+#endif /* UMF_BASE_ALLOC_H */
--- a/src/base_alloc/base_alloc_windows.c
+++ b/src/base_alloc/base_alloc_windows.c
@@ -1,0 +1,21 @@
+/*
+ * Copyright (C) 2024 Intel Corporation
+ *
+ * Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+*/
+
+#include <windows.h>
+
+void *ba_os_alloc(size_t size) {
+    return VirtualAlloc(NULL, size, MEM_RESERVE | MEM_COMMIT, PAGE_READWRITE);
+}
+
+void ba_os_free(void *ptr, size_t size) { VirtualFree(ptr, 0, MEM_RELEASE); }
+
+size_t ba_os_get_page_size(void) { return 4096; /* TODO */ }
+
+//
+// TODO:
+// 1) implement ba_os_get_page_size()
+//
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -39,6 +39,7 @@ function(add_umf_test)
     target_include_directories(${TEST_TARGET_NAME} PRIVATE
         ${UMF_CMAKE_SOURCE_DIR}/include
         ${UMF_CMAKE_SOURCE_DIR}/src
+        ${UMF_CMAKE_SOURCE_DIR}/src/base_alloc
         ${UMF_TEST_DIR}/common
         ${UMF_TEST_DIR})
 
@@ -108,4 +109,12 @@ if(UMF_BUILD_OS_MEMORY_PROVIDER AND LINUX) # OS-specific functions are implement
     add_umf_test(NAME provider_os_memory_config
                  SRCS provider_os_memory_config.cpp
                  LIBS umf_utils numa)
+endif()
+
+# TODO: add Windows tests
+if(LINUX)
+    # the base_alloc test uses linux pthreads
+    add_umf_test(NAME base_alloc
+                SRCS test_base_alloc.c
+                LIBS umf_utils)
 endif()
--- a/test/test_base_alloc.c
+++ b/test/test_base_alloc.c
@@ -1,0 +1,73 @@
+/*
+ * Copyright (C) 2024 Intel Corporation
+ *
+ * Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+*/
+
+#include <assert.h>
+#include <pthread.h>
+#include <stdio.h>
+#include <string.h>
+#include <sys/syscall.h>
+#include <unistd.h>
+
+#include "base_alloc.h"
+#include "test_helpers.h"
+
+#define NTHREADS 10
+#define ITERATIONS 1000
+#define ALLOCATION_SIZE 16
+
+static void *start_routine(void *arg) {
+    unsigned char *ptrs[ITERATIONS];
+    umf_ba_pool_t *pool = (umf_ba_pool_t *)arg;
+
+    long TID = syscall(SYS_gettid);
+
+    for (int i = 0; i < ITERATIONS; i++) {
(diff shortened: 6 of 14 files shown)


src/base_alloc/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+if(LINUX)
+    set(BA_SOURCES ${BA_SOURCES} ${BA_SOURCES_LINUX})
+elseif(WINDOWS)
+    set(BA_SOURCES ${BA_SOURCES} ${BA_SOURCES_WINDOWS})
+elseif(MACOSX)
+    set(BA_SOURCES ${BA_SOURCES} ${BA_SOURCES_MACOSX})
+endif()
+
+if(CMAKE_BUILD_TYPE STREQUAL "Debug")
+	add_compile_options(-DDEBUG)
```
bratpiorka, line 30: fix indentation, also - shouldn't this be added to top - level cmake?

src/base_alloc/base_alloc.c
```diff
@@ -0,0 +1,140 @@
...
+struct umf_ba_pool_t {
+    size_t aligned_size;
+    ba_list_item_t *free_list;
+    ba_list_item_t *alloc_list;
+#ifdef DEBUG
+    long tid;
+#endif /* DEBUG */
+};
+
+static int ba_fill_pool(umf_ba_pool_t *pool) {
```
bratpiorka, line 37: add a one-sentence description what this function does

src/base_alloc/base_alloc.c
```diff
@@ -0,0 +1,140 @@
...
+
+typedef struct ba_list_item_t ba_list_item_t;
+
+struct ba_list_item_t {
+    ba_list_item_t *next;
+    void *addr;
+};
+
+struct umf_ba_pool_t {
+    size_t aligned_size;
```
bratpiorka, line 29: Is this the size of all elements that this pool manages? Please add comments that explain this.

src/base_alloc/base_alloc.c
```diff
@@ -0,0 +1,140 @@
...
+
+    pool->free_list = last_list_item;
+
+    return 0;
+}
+
+umf_ba_pool_t *umf_ba_create(size_t size) {
+    size_t aligned_size;
+    // align size to MEMORY_ALIGNMENT bytes
+    size_t rest = size & (MEMORY_ALIGNMENT - 1);
```
bratpiorka, line 71: do we have a helper function for calculating aligned address?

src/base_alloc/base_alloc.c
```diff
@@ -0,0 +1,140 @@
...
+    if (!pool_memory) {
+        return -1;
+    }
+
+    size_t pool_size_left = BA_POOL_SIZE;
+    ba_list_item_t *last_list_item = NULL;
+
+    while (pool_size_left >= pool->aligned_size) {
+        ba_list_item_t *new_list_item =
+            (ba_list_item_t *)linear_alloc(sizeof(ba_list_item_t));
```
bratpiorka, line 50: I think this code would be much simpler if we keep metadata and user data in one place. This way we wouldn't require linear_alloc at all - @pbalcer what do you think?

bratpiorka (changes requested):

src/base_alloc/CMakeLists.txt
igchor: I think NDEBUG is always defined for release builds, so cannot we just use `#ifndef NDEBUG`?

src/base_alloc/base_alloc.c
```diff
@@ -0,0 +1,140 @@
...
+    pool->alloc_list = item;
+
+    assert(item->addr);
+    void *ptr = item->addr;
+    item->addr = NULL;
+
+    return ptr;
+}
+
+void umf_ba_free(umf_ba_pool_t *pool, void *ptr) {
```
igchor, line 121: I don't think we can forbid user from calling free from a different thread (than the one used for allocating `ptr`). We will need to support this case for critnib and memory tracking.

src/base_alloc/base_alloc.h
```diff
@@ -0,0 +1,24 @@
...
+/* A MT-safe fixed-size-pool base allocator */
+
+#ifndef UMF_BASE_ALLOC_H
+#define UMF_BASE_ALLOC_H 1
+
+#include <stddef.h>
+
+typedef struct umf_ba_pool_t umf_ba_pool_t;
+
+struct umf_ba_pool_t;
```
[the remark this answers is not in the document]
ldorau: No, removed

src/base_alloc/base_alloc_windows.c
```diff
@@ -0,0 +1,27 @@
+/*
+ * Copyright (C) 2024 Intel Corporation
+ *
+ * Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+*/
+
+#include <Windows.h>
```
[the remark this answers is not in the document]
ldorau: Done

test/CMakeLists.txt
```diff
@@ -106,3 +107,10 @@ if(UMF_BUILD_OS_MEMORY_PROVIDER AND LINUX) # OS-specific functions are implement
                  SRCS memspace_numa.cpp
                  LIBS numa)
 endif()
+
+if(LINUX)
+    # the base_alloc test uses linux pthreads
```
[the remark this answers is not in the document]
ldorau: Done

src/base_alloc/CMakeLists.txt
ldorau: Done.

test/CMakeLists.txt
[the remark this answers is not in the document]
igchor: Why do we need to have separate windows and linux tests anyway? Can't you just write them in C++? (what way you can use std::thread that works everywhere).

src/base_alloc/base_alloc.c
ldorau: It seems so. I've just dreamt I could make it lock-free ;-) ...

ldorau: Done

src/base_alloc/base_alloc.c
ldorau: Added.

src/base_alloc/base_alloc.c
ldorau: Done

src/base_alloc/base_alloc.c
ldorau: Done

src/base_alloc/base_alloc.c
bratpiorka: still, we could make a single mmap() call instead of two, but let's wait for Piotr comment

pbalcer: yea, I agree with Rafal, this would be simpler (and faster) with an intrusive list embedded in the returned data.

src/base_alloc/base_alloc.c
```diff
@@ -0,0 +1,148 @@
...
+        pool_size_left -= pool->aligned_size;
+        pool_memory += pool->aligned_size;
+    }
+
+    pool->free_list = last_list_item;
+
+    return 0;
+}
+
+umf_ba_pool_t *umf_ba_create(size_t size) {
```
pbalcer, line 71: I thought about this a bit, and I think this could be solved by using mmap here and using the remaining data for populating the list:
```
struct ba_pool_meta {
   ...
}
struct ba_pool {
     struct struct ba_pool_meta meta;
     char data[pagesize - sizeof(struct ba_pool_meta)];
}

ba_pool *create(size_t size) {
   ba_pool *pool = mmap(4k);
   pool_fill_with(&pool->data, sizeof(pool->data));
}
```

.github/workflows/.spellcheck-conf.toml
```diff
@@ -1,3 +1,3 @@
 [default]
 # Don't correct the "ASSERTne" word
-extend-ignore-words-re = ["ASSER", "Tne"]
+extend-ignore-words-re = ["ASSER", "Tne", "ba", "BA"]
```
lukaszstolarczuk, line 3: I guess you can update the comment above 😉

ldorau: Done

src/base_alloc/base_alloc.c
ldorau: Done

src/base_alloc/base_alloc.c
igchor: I think we could only have a mutex around the `ba_os_alloc` call and all `next` and `free_list` pointers modifications could be done using compare exchange. Free could also be lock-free then.

Of course we can do it in another PR

src/base_alloc/base_alloc.c
```diff
@@ -0,0 +1,129 @@
...
+            return NULL;
+        }
+        ba_divide_memory_into_chunks(pool, new_pool, BA_POOL_SIZE);
+    }
+
+    ba_mem_chunk_t *chunk = pool->metadata.free_list;
+    pool->metadata.free_list = pool->metadata.free_list->next;
+    util_mutex_unlock(pool->metadata.free_lock);
+
+    return &chunk->user_data;
```
igchor, line 111: Do we actually need to waste 8 bytes per each allocation? `next` pointer is only used for blocks that are free so can't we just return `(void*) &chunk->next;` here to the user? Or would that be against strict aliasing policy?

src/base_alloc/base_alloc.c
```diff
@@ -0,0 +1,129 @@
...
+    // allocate and init free_lock
+    pool->metadata.free_lock = util_mutex_init(pool_memory);
+    if (!pool->metadata.free_lock) {
+        return NULL;
+    }
+
+    pool_memory += util_mutex_get_size(); // free_lock is here
+    size_left -= util_mutex_get_size();   // for free_lock
+
+    util_mutex_lock(pool->metadata.free_lock);
```
igchor, line 88: Why do we need a lock here? Is this memory visible/accessible by anyone else at this moment?

src/base_alloc/base_alloc.c
```diff
@@ -0,0 +1,129 @@
...
+    pool->metadata.free_list = NULL;
+    ba_divide_memory_into_chunks(pool, pool_memory, size_left);
```
igchor, line 90: Do we need to actually divide all the memory eagerly at the beginning?

I think we could get rid of ba_divide_memory_into_chunks and just do 'bump' allocation for the first `pool_size / aligned_size` allocations. That is, have an `allocated_size` variable which would initially be 0. Every time we allocate, if `allocated_size + aligned_size` is <= `pool_size` then we just return `pool->base_address + allocated_size`. If `allocated_size + aligned_size` >= `pool_size` we look at the free list.

And the free list is only populated on `umf_ba_free`.

Not sure if this is worth doing, tough - the overhead of `ba_os_alloc` probably makes the ba_divide_memory_into_chunks unnoticable.

src/base_alloc/base_alloc_os_mem.c
```diff
@@ -0,0 +1,66 @@
...
+
+typedef struct os_mem_t {
+    os_mem_region_t region[SIZE_OS_MEM_REGIONS_BUFFER];
+    int index;
+} os_mem_t;
+
+static __TLS os_mem_t OS_mem;
+
+void *ba_os_alloc(size_t size) {
+    // assert(OS_mem.index < SIZE_OS_MEM_REGIONS_BUFFER - 1);
```
igchor, line 31: why is this commneted out? we should have some check like this, right? (and probably return an error, even in release build)

src/base_alloc/base_alloc.c
ldorau: Right, it is not needed here. Removed.
Done.

src/base_alloc/base_alloc.c
ldorau: > Do we actually need to waste 8 bytes per each allocation?

What do you mean is wasted here? A `next` pointer? I'm afraid we can't get rid of it.

src/base_alloc/base_alloc_os_mem.c
ldorau: I am not sure if we should assert in this case, when we can allocate more memory, but we cannot only save this pointer in the array, so we will leak this memory ...

ldorau:
Working on a new version, please PAUSE review.

src/base_alloc/base_alloc_os_mem.c
ldorau: Removed.
Done.

ldorau:
Ready for review

src/base_alloc/base_alloc.c
pbalcer: hm, I think Igor is right. We don't need to get rid of it - it's still needed when the block is free, but can be overwritten with user data when it's given out as allocation. So when you remove  `&chunk->user_data` (and the reverse in free) things should just keep working?

src/base_alloc/base_alloc.c
pbalcer: I mean, I'd rather have the code be simpler and preemptively split it in a cold path, then have to special-case the hot path.

pbalcer (approved):

test/test_base_alloc.c
```diff
@@ -0,0 +1,72 @@
...
+    return NULL;
+}
+
+int main() {
+    pthread_t thread[NTHREADS];
+    umf_ba_pool_t *pool = umf_ba_create(ALLOCATION_SIZE);
+
+    for (int i = 0; i < NTHREADS; i++) {
+        int ret = pthread_create(&thread[i], NULL, start_routine, pool);
+        if (ret) {
```
lukaszstolarczuk, line 54: shouldn't you use `ASSERT_*` macros...?

test/test_base_alloc.c
```diff
@@ -0,0 +1,72 @@
...
+    for (int i = 0; i < NTHREADS; i++) {
+        void *retval;
+        int ret = pthread_join(thread[i], &retval);
+        if (ret) {
+            fprintf(stderr, "pthread_join() failed!\n");
+            return -1;
+        }
+    }
+
+    printf("TEST PASSED\n");
```
lukaszstolarczuk, line 69: not needed, I guess

src/base_alloc/base_alloc.c
```diff
@@ -0,0 +1,167 @@
...
+
+#include <assert.h>
+
+#include "base_alloc.h"
+#include "base_alloc_internal.h"
+#include "utils_common.h"
+#include "utils_concurrency.h"
+
+// minimum size of a single pool of the base allocator,
+#define MINIMUM_POOL_SIZE (ba_os_get_page_size())
```
lukaszstolarczuk, line 16: should we have this documented somehow...? (perhaps we can do it, when BA is actually used)

src/base_alloc/base_alloc.c
igchor: It will happen on each allocation that needs to call `ba_os_alloc` (when free list is empty). But yeah, it might actually be faster anyway since we have less atomics.

src/base_alloc/base_alloc.c
igchor: I think we could just remove this assert in `ba_divide_memory_into_chunks` and once all the memory is divided just do compare exchange on pool->metadata.free_list. If it's empty, it will be equivalent to current logic, if it's not, then this would be similar to how we would insert to the list in `umf_ba_free`.

So, there would be 3 CAS operations:
- in umf_ba_free() when inserting an item to the list
- in umf_ba_alloc() when taking an item from the list
- in ba_divide_memory_into_chunks() when inserting a bunch items to the list

src/base_alloc/base_alloc.c
igchor: Yeah, that was my thinking. In other words, the `ba_mem_chunk_t` could be an union instead of a struct. (but C forbids having a flexible array member in union so we would probably just need to cast the allocation to `ba_mem_chunk_t*` in free).

igchor:
@ldorau  A general question: do we ever free the memory allocated by `ba_os_alloc` (before calling `umf_ba_destroy`)? I think we might need some way to limit the amount of pooled memory, similarly as we have in disjoint_pool (MaxSize, which is a soft-limit, meaning there is always at most MaxSize preallocated free memory in the pool).

test/test_base_alloc.c
ldorau: Done

src/base_alloc/base_alloc.c
ldorau: OK, I got it now.
Done.

src/base_alloc/base_alloc.c
ldorau: @igchor is it resolved now?

src/base_alloc/base_alloc.c
ldorau: @igchor Is that what you meant?
https://github.com/ldorau/unified-memory-framework/commit/ef34eef7058b4a150a8bcc7b49d1a6d9bf6294c4

src/base_alloc/base_alloc.c
```diff
@@ -0,0 +1,167 @@
...
+
+struct umf_ba_pool_meta_t {
+    // address of the beginning of this pool (allocated region)
+    umf_ba_pool_t *pool_addr;
+    // address of the beginning of the next pool (a list of allocated pools to be freed in umf_ba_destroy())
+    umf_ba_pool_t *next_pool;
+
+    // the following fields are set and used only in the main (the first) pool
+    size_t pool_size; // size of each pool (argument of each ba_os_alloc() call)
+    size_t pool_usable_size;   // usable size of each pool
```
bratpiorka, line 39: I think pool_usable_size is not needed at all in this struct. It doesn't change after allocating new pool, right? It could be calculated when needed from other constants.

src/base_alloc/base_alloc.c
```diff
@@ -0,0 +1,167 @@
...
+typedef struct ba_mem_chunk_t ba_mem_chunk_t;
+
+// memory chunk of size 'chunk_size'
+struct ba_mem_chunk_t {
+    ba_mem_chunk_t *next;
+};
+
+struct umf_ba_pool_meta_t {
+    // address of the beginning of this pool (allocated region)
+    umf_ba_pool_t *pool_addr;
```
bratpiorka, line 33: why this is umf_ba_pool_t and not umf_ba_pool_meta_t? You got a circular dependency

src/base_alloc/base_alloc.c
```diff
@@ -0,0 +1,167 @@
...
+    pool_size = align_size(pool_size, ba_os_get_page_size());
+
+    umf_ba_pool_t *pool = (umf_ba_pool_t *)ba_os_alloc(pool_size);
+    if (!pool) {
+        return NULL;
+    }
+
+    pool->metadata.pool_addr = pool;
+    pool->metadata.pool_size = pool_size;
+    pool->metadata.pool_usable_size = pool_size - metadata_size;
```
bratpiorka, line 94: if you create functions calc_pool_size(size) and calc_metadata_size(size) you don't need to keep these values here

bratpiorka (changes requested):

src/base_alloc/base_alloc.c
igchor: Yes, exactly. 

I would only propose implementing an os-agnostic version of `__sync_bool_compare_and_swap` in utils_concurreny.h and using that. Also when you're reading the old `free_list` value you should use util_atomic_load_acquire.

igchor: And there's also a question whether we should get rid of the mutex around `ba_os_alloc` in `umf_ba_alloc`.  Without the mutex, if `pool->metadata.free_list == NULL` and 2 (or more) threads happen to call `umf_ba_alloc` at the same time, we can have multiple `ba_os_alloc` calls. Everything should be consistent but we'll possibly allocate more than needed.

I think it would be better to leave the mutex there. (the compare exchange operations are still needed to synchronize with free).

igchor: One more thing: the `if (pool->metadata.free_list == NULL)` check inside `umf_ba_alloc` should probably be moved to the `do {} while` loop because some other thread might allocate the last free chunk from the free list while our thread is trying to do CAS.

ldorau: OK, so let's do that in a separate PR.

src/base_alloc/base_alloc.c
ldorau: @bratpiorka please resolve

src/base_alloc/base_alloc.c
ldorau: @pbalcer please resolve

test/test_base_alloc.c
ldorau: @lukaszstolarczuk please resolve

test/test_base_alloc.c
ldorau: @lukaszstolarczuk please resolve

src/base_alloc/base_alloc.c
bratpiorka: @lukaszstolarczuk base allocator wouldn't be public, similar as critnib

lukaszstolarczuk: @bratpiorka, yeah, I get that, but it will affect pools using this base

@ldorau, in the docs, I guess 😉

test/test_base_alloc.c
```diff
@@ -0,0 +1,71 @@
...
+
+    for (int i = 0; i < ITERATIONS; i++) {
+        for (int k = 0; k < ALLOCATION_SIZE; k++) {
+            if (*(ptrs[i] + k) != ((i + TID) & 0xFF)) {
+                fprintf(stderr,
+                        "i = %i k = %i, *(ptrs[i] + k) = %i != ((i + TID) & "
+                        "0xFF) = %li\n",
+                        i, k, *(ptrs[i] + k), ((i + TID) & 0xFF));
+            }
+            assert(*(ptrs[i] + k) == ((i + TID) & 0xFF));
```
lukaszstolarczuk, line 41: what about `ASSERT_*` here?

test/test_base_alloc.c
lukaszstolarczuk: heh, no need to ping me twice on the same issue :wink:

test/test_base_alloc.c
ldorau: Right, I've missed that.
Done

src/base_alloc/base_alloc.c
```diff
@@ -0,0 +1,167 @@
...
+// minimum number of chunks used to calculate the size of pools
+#define MINIMUM_CHUNK_COUNT (128)
+
+// alignment of the base allocator
+#define MEMORY_ALIGNMENT (8)
+
+typedef struct ba_mem_chunk_t ba_mem_chunk_t;
+
+// memory chunk of size 'chunk_size'
+struct ba_mem_chunk_t {
```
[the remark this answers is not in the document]
ldorau: Renamed to `umf_ba_chunk_t`.
Done

src/base_alloc/base_alloc.c
ldorau: It is not a circular dependency, It is only a pointer stored, but actually it is not necessary, so I removed it.
Done.

src/base_alloc/base_alloc.c
bratpiorka: I think it would be worth documenting if this would be exposed in cmake options or something. Now it isn't and we don't plan to do that. So documenting this inside a code as a comment is IMO enough.

src/base_alloc/base_alloc.c
```diff
@@ -0,0 +1,167 @@
...
+    ba_mem_chunk_t *next;
+};
+
+struct umf_ba_pool_meta_t {
+    // address of the beginning of this pool (allocated region)
+    umf_ba_pool_t *pool_addr;
+    // address of the beginning of the next pool (a list of allocated pools to be freed in umf_ba_destroy())
+    umf_ba_pool_t *next_pool;
+
+    // the following fields are set and used only in the main (the first) pool
```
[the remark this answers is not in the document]
ldorau: Done

src/base_alloc/base_alloc.c
ldorau: Done

src/base_alloc/base_alloc.c
```diff
@@ -0,0 +1,167 @@
...
+    pool->metadata.next_pool = NULL; // this is the only pool now
+    pool->metadata.chunk_size = chunk_size;
+
+    char *pool_memory = (char *)&pool->data;
+    size_t size_left = pool->metadata.pool_usable_size;
+
+    // allocate and init free_lock
+    pool->metadata.free_lock = util_mutex_init(pool_memory);
+    if (!pool->metadata.free_lock) {
+        return NULL;
```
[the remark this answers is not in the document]
ldorau: Right. Done

src/base_alloc/base_alloc.c
ldorau: `pool_size` is calculated once and stored only in one place now. This is simpler and faster.

src/base_alloc/base_alloc.c
```diff
@@ -0,0 +1,167 @@
...
+    umf_ba_pool_t *current_pool;
+    umf_ba_pool_t *next_pool = pool->metadata.next_pool;
+    while (next_pool) {
+        current_pool = next_pool;
+        next_pool = next_pool->metadata.next_pool;
+        ba_os_free(current_pool->metadata.pool_addr, size);
+    }
+
+    util_mutex_destroy_not_free(pool->metadata.free_lock);
+    ba_os_free(pool, size);
```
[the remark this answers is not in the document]
ldorau: The free list should be full here :-). I will add a debug check.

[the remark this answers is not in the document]
ldorau: Done

src/base_alloc/base_alloc.c
```diff
@@ -0,0 +1,192 @@
...
+*/
+
+#include <assert.h>
+
+#include "base_alloc.h"
+#include "base_alloc_internal.h"
+#include "utils_common.h"
+#include "utils_concurrency.h"
+
+// minimum size of a single pool of the base allocator,
```
KFilipek, line 15: Unnecessary ',' at the end of line

src/base_alloc/base_alloc.c
ldorau: New version: https://github.com/ldorau/unified-memory-framework/commit/f88053c040377613fcef64a64dfdeea0b2216740 - will be submitted in a separate PR

src/base_alloc/base_alloc.c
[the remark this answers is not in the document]
ldorau: Done now

bratpiorka (approved):

src/base_alloc/base_alloc.c
```diff
@@ -0,0 +1,228 @@
...
+};
+
+struct umf_ba_pool_t {
+    // address of the beginning of the next pool (a list of allocated pools to be freed in umf_ba_destroy())
+    umf_ba_next_pool_t *next_pool;
+
+    // metadata is set and used only in the main (the first) pool
+    struct umf_ba_main_pool_meta_t metadata;
+
+    // data area of of the main (the first one) starts here
```
lukaszstolarczuk, line 53: 1. `of of`
2. the main what? (missing `pool` I guess)

KFilipek (approved):

src/base_alloc/base_alloc.c
```diff
@@ -0,0 +1,228 @@
...
+    os_mutex_t *free_lock;     // lock of free_list
+    umf_ba_chunk_t *free_list; // list of free chunks
+#ifndef NDEBUG
+    size_t n_pools;
+    size_t n_allocs;
+    size_t n_chunks;
+#endif /* NDEBUG */
+};
+
+struct umf_ba_pool_t {
```
lukaszstolarczuk, line 46: you could add a general comment above this struct (and `umf_ba_next_pool_t` as well). The distinction between "main" and the rest of the pools could be more clear.

src/base_alloc/base_alloc.c
ldorau: Done

src/base_alloc/base_alloc.c
ldorau: Done

igchor:
> > @ldorau A general question: do we ever free the memory allocated by `ba_os_alloc` (before calling `umf_ba_destroy`)?
> 
> No, we don't.
> 
> > I think we might need some way to limit the amount of pooled memory, similarly as we have in disjoint_pool (MaxSize, which is a soft-limit, meaning there is always at most MaxSize preallocated free memory in the pool).
> 
> Why? What would be this limit for?

To avoid a situation where we would hold a large amount of memory in the pool even if it is not used. Consider a situation where we have a lot of allocations to ba_alloc (.e.g. because the user is allocating a lot of coarse grain allocations and we use tracking) and then all that memory is freed (umf_ba_free). The memory would still be in the pool until the application is finished. This might not be a problem for most workloads but I think it would be good to have some upper limit on how much (free) memory we want to keep in the pool.

I don't think we need to implement it right now, but let's keep in mind.

igchor (approved):

bratpiorka:
@ldorau please rebase so I can merge these changes

ldorau:
> @ldorau please rebase so I can merge these changes

Done

ldorau:
Rebased again

PR #167, opened 2024-01-22
Update C++17 requirement cases

PatKamin:
C++17 is required to build UMF only when two flags are enabled:
- UMF_BUILD_LIBUMF_POOL_DISJOINT,
- UMF_BUILD_TESTS.

In other cases, all is needed is a C compiler.

### DIFF
--- a/.github/workflows/pr_push.yml
+++ b/.github/workflows/pr_push.yml
@@ -11,15 +11,16 @@ jobs:
         name: Fast build
         strategy:
             matrix:
-                os: [ubuntu-latest, windows-latest]
+                os: [ubuntu-latest]
+                nproc: ['$(nproc)']
+                pool_scalable: ['ON']
+                disjoint: ['ON', 'OFF']
+                jemalloc: ['ON']
                 include:
-                    - os: ubuntu-latest
-                      nproc: $(nproc)
-                      pool_scalable: 'ON'
-                      jemalloc: 'ON'
                     - os: windows-latest
                       nproc: $Env:NUMBER_OF_PROCESSORS
                       pool_scalable: 'OFF'
+                      disjoint: 'OFF'
                       jemalloc: 'OFF'
         runs-on: ${{matrix.os}}
 
@@ -42,8 +43,9 @@ jobs:
             -DUMF_DEVELOPER_MODE=ON
             -DUMF_ENABLE_POOL_TRACKING=ON
             -DUMF_BUILD_LIBUMF_POOL_SCALABLE=${{matrix.pool_scalable}}
-            -DUMF_BUILD_LIBUMF_POOL_DISJOINT=ON
+            -DUMF_BUILD_LIBUMF_POOL_DISJOINT=${{matrix.disjoint}}
             -DUMF_BUILD_LIBUMF_POOL_JEMALLOC=${{matrix.jemalloc}}
+            -DUMF_BUILD_TESTS=OFF
 
         - name: Build
           run: >
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,17 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 cmake_minimum_required(VERSION 3.14.0 FATAL_ERROR)
-project(unified-memory-framework VERSION 0.1.0)
-
-set(CMAKE_CXX_STANDARD 17)
-set(CMAKE_CXX_STANDARD_REQUIRED YES)
-
-include(CTest)
-include(CMakePackageConfigHelpers)
-include(GNUInstallDirs)
-
-list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
-include(helpers)
+project(unified-memory-framework VERSION 0.1.0 LANGUAGES C)
 
 # Build Options
 option(UMF_BUILD_SHARED_LIBRARY "Build UMF as shared library" OFF)
@@ -30,6 +20,28 @@ option(USE_ASAN "Enable AddressSanitizer checks" OFF)
 option(USE_UBSAN "Enable UndefinedBehaviorSanitizer checks" OFF)
 option(USE_TSAN "Enable ThreadSanitizer checks" OFF)
 option(USE_MSAN "Enable MemorySanitizer checks" OFF)
+
+# For using the options listed in the OPTIONS_REQUIRING_CXX variable
+# a C++17 compiler is required. Moreover, if these options are not set,
+# CMake will set up a strict C build, without C++ support.
+set(OPTIONS_REQUIRING_CXX
+    "UMF_BUILD_TESTS"
+    "UMF_BUILD_LIBUMF_POOL_DISJOINT")
+foreach(option_name ${OPTIONS_REQUIRING_CXX})
+    if(${option_name})
+        enable_language(CXX)
+        set(CMAKE_CXX_STANDARD 17)
+        set(CMAKE_CXX_STANDARD_REQUIRED YES)
+        break()
+    endif()
+endforeach()
+
+include(CTest)
+include(CMakePackageConfigHelpers)
+include(GNUInstallDirs)
+
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
+include(helpers)
 
 if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
     set(LINUX TRUE)
--- a/README.md
+++ b/README.md
@@ -72,11 +72,14 @@ The `UMF_BUILD_LIBUMF_POOL_SCALABLE` option has to be turned `ON` to build this 
 ### Requirements
 
 Required packages:
-- C++ compiler with C++17 support
+- C compiler
 - [CMake](https://cmake.org/) >= 3.14.0
 
 For development and contributions:
 - clang-format-15.0 (can be installed with `python -m pip install clang-format==15.0.7`)
+
+For building tests and Disjoint Pool:
+- C++ compiler with C++17 support
 
 ### Benchmark
 
--- a/cmake/helpers.cmake
+++ b/cmake/helpers.cmake
@@ -135,7 +135,9 @@ macro(add_sanitizer_flag flag)
 
     # Check C and CXX compilers for given sanitizer flag.
     check_c_compiler_flag("${SANITIZER_FLAG}" "C_${check_name}")
-    check_cxx_compiler_flag("${SANITIZER_FLAG}" "CXX_${check_name}")
+    if (CMAKE_CXX_COMPILE_FEATURES)
+        check_cxx_compiler_flag("${SANITIZER_FLAG}" "CXX_${check_name}")
+    endif()
     if (NOT ${C_${check_name}} OR NOT ${CXX_${check_name}})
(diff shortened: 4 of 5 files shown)


igchor (approved):

bratpiorka (approved):

cmake/helpers.cmake
```diff
@@ -137,8 +137,10 @@ macro(add_sanitizer_flag flag)
 
     # Check C and CXX compilers for given sanitizer flag.
     check_c_compiler_flag("${SANITIZER_FLAG}" "C_${check_name}")
-    check_cxx_compiler_flag("${SANITIZER_FLAG}" "CXX_${check_name}")
-    if (${C_${check_name}} OR ${CXX_${check_name}})
+    if (CMAKE_CXX_STANDARD_REQUIRED)
+        check_cxx_compiler_flag("${SANITIZER_FLAG}" "CXX_${check_name}")
+    endif()
+    if (${C_${check_name}} OR $<BOOL:CXX_${check_name}>)
```
lukaszstolarczuk, line 143: don't you need this `BOOL` a few lines below as well...?

README.md
```diff
@@ -67,12 +67,15 @@ The `UMF_BUILD_LIBUMF_POOL_SCALABLE` option has to be turned `ON` to build this
 ### Requirements
 
 Required packages:
-- C++ compiler with C++17 support
+- C compiler
```
lukaszstolarczuk, line 70: not sure how much you want to rely on this info... perhaps we should have a test for that? at least you can verify one time on the OS with no C++ compiler (pls send a link to the build, if go this way 😉 )

cmake/helpers.cmake
PatKamin: It's not needed at all after making some changes.

README.md
PatKamin: I've made a [test branch](https://github.com/PatKamin/unified-memory-framework/actions/runs/7625835325/job/20770973393) to run CI without a C++ compiler and it looks like the project won't compile because the `project` CMake function fails when no C++ compiler is detected.
So currently the UMF won't be compiled just because the current form of `project()` include implicit checks both for C and C++ compilers.

I've added an explicit C-only CMake setup in cases where none of the two flags requiring C++ is set. This will help in keeping the bulk of UMF a pure C code.

igchor (approved):

CMakeLists.txt
```diff
@@ -31,6 +20,29 @@ option(USE_UBSAN "Enable UndefinedBehaviorSanitizer checks" OFF)
 option(USE_TSAN "Enable ThreadSanitizer checks" OFF)
 option(USE_MSAN "Enable MemorySanitizer checks" OFF)
 
+set(CXX_REQUIRING_OPTIONS
```
KFilipek, line 23: Shouldn't be "CXX_REQUIRED_OPTIONS"?

KFilipek (changes requested):

CMakeLists.txt
```diff
@@ -31,6 +20,29 @@ option(USE_UBSAN "Enable UndefinedBehaviorSanitizer checks" OFF)
...
+    if(${option_name})
+        project(unified-memory-framework VERSION 0.1.0)
+        set(CMAKE_CXX_STANDARD 17)
+        set(CMAKE_CXX_STANDARD_REQUIRED YES)
+        break()
+    endif()
+endforeach()
+
+if(NOT CMAKE_CXX_STANDARD_REQUIRED)
+    project(unified-memory-framework VERSION 0.1.0 LANGUAGES C)
```
KFilipek, line 36: Instead of keeping two lines with hardcoded version "0.1.0" it can be done via `enable_language()` and then only enable C, I think that it should be also enabled for C++ variant of the execution path.
Source: https://cmake.org/cmake/help/latest/command/enable_language.html

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index e73bfa4..6f4fb4a 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception

 cmake_minimum_required(VERSION 3.14.0 FATAL_ERROR)
+project(unified-memory-framework VERSION 0.1.0 LANGUAGES C)

 # Build Options
 option(UMF_BUILD_SHARED_LIBRARY "Build UMF as shared library" OFF)
@@ -25,17 +26,13 @@ set(CXX_REQUIRING_OPTIONS
     "UMF_BUILD_LIBUMF_POOL_DISJOINT")
 foreach(option_name ${CXX_REQUIRING_OPTIONS})
     if(${option_name})
-        project(unified-memory-framework VERSION 0.1.0)
+        enable_language(CXX)
         set(CMAKE_CXX_STANDARD 17)
         set(CMAKE_CXX_STANDARD_REQUIRED YES)
         break()
     endif()
 endforeach()

-if(NOT CMAKE_CXX_STANDARD_REQUIRED)
-    project(unified-memory-framework VERSION 0.1.0 LANGUAGES C)
-endif()
-
 include(CTest)
 include(CMakePackageConfigHelpers)
 include(GNUInstallDirs)
```

KFilipek (changes requested):

CMakeLists.txt
lukaszstolarczuk: heh, I believe nope - is that some options are requiring CXX, not the other way around - maybe (but just maybe :D ) rename to `OPTIONS_REQUIRING_CXX` to be more clear 😉

lukaszstolarczuk, line 23: you may add a short comment here

PatKamin: Renamed

PatKamin: Done

CMakeLists.txt
PatKamin: Thanks, done.

cmake/helpers.cmake
```diff
@@ -135,16 +135,19 @@ macro(add_sanitizer_flag flag)
...
 
     # Check C and CXX compilers for sanitizer arguments.
     if (${SANITIZER_ARGS})
         check_c_compiler_flag("${SANITIZER_ARGS}" "C_HAS_SAN_ARGS")
-        check_cxx_compiler_flag("${SANITIZER_ARGS}" "CXX_HAS_SAN_ARGS")
-
+        if (CMAKE_CXX_COMPILE_FEATURES)
+            check_cxx_compiler_flag("${SANITIZER_ARGS}" "CXX_HAS_SAN_ARGS")
+        endif()
         if (NOT ${C_HAS_SAN_ARGS} OR NOT ${CXX_HAS_SAN_ARGS})
```
kswiecicki, line 151: This will probably fail when `UMF_BUILD_TESTS` and `UMF_BUILD_LIBUMF_POOL_DISJOINT` options are turned `OFF`.
Could you set the value of `CXX_HAS_SAN_ARGS` to one of the positive constants when CXX compiler should be ignored? (https://cmake.org/cmake/help/latest/command/if.html#constant)

lukaszstolarczuk: hmm.. I think you're on to something... CXX_HAS_SAN_ARGS should be empty here, meaning false. But why didn't it fail on the CI? maybe some values were cached...?

.github/workflows/pr_push.yml
```diff
@@ -16,10 +16,17 @@ jobs:
                     - os: ubuntu-latest
                       nproc: $(nproc)
                       pool_scalable: 'ON'
+                      disjoint: 'ON'
+                      jemalloc: 'ON'
+                    - os: ubuntu-latest
```
lukaszstolarczuk, line 21: ugh, I think `include` does not work in this case as we would like it to 😉  We ended up only with 2 FastBuild jobs. I think, you have to add some param to the matrix (beside `os`) to overwrite its value in the `include`.

FYI, docs: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrixinclude

PatKamin: That's odd. Refactored this a bit, let's see if it's any better.

cmake/helpers.cmake
PatKamin: It seems to work with both CMake config variables [turned off](https://github.com/oneapi-src/unified-memory-framework/actions/runs/7723525323/job/21053822402?pr=167)

kswiecicki: Alright, I see that it tests just the C compiler for appropriate sanitizer and doesn't fail, it seems to work correctly.

.github/workflows/pr_push.yml
kswiecicki: Now, there seem to be 2 ubuntu FastBuilds because of the `disjoint: ['ON', 'OFF']` matrix option. Couldn't we just run the job with disjoint pool set to `ON` and disregard the case where it's not built?

PatKamin: I wanted to test exactly this case, where both disjoint pool and tests are turned off. That's the only case in our CI where you can verify a clean C build without CMake support for C++.

cmake/helpers.cmake
lukaszstolarczuk: hmm... heh, do we test this part of the code in FastBuild?

PatKamin: No, that's why I've [linked](https://github.com/PatKamin/unified-memory-framework/actions/runs/7723653607/job/21054205754) to the modified branch.

lukaszstolarczuk: heh, ok, I'm blind I didn't see your links. Nonetheless, after our call I dug a little into the code and found the issue (issues to be precise) 😄 

Please see my patch here:
https://github.com/lukaszstolarczuk/unified-memory-framework/commit/83c4c331f0947ad195cadbe4c059219dab72246d

If you want some explanations about these, please contact me 😉 

You can see a few of my tries with debug prints, here:
https://github.com/lukaszstolarczuk/unified-memory-framework/actions

kswiecicki: But it was meant to be an `OR`, I've had a chat with @bratpiorka about this. Couldn't this cause a situation where code compiled with the C compiler is sanitized, while code compiled with the C++ compiler remains unsanitized, or vice versa?

lukaszstolarczuk: heh, while this may be a valid point, questions:
1. when could that happen?
2. sanitizers are mostly for our use case - is it something we could expect in our CI?
3. do you have a quick solution to that `if` to wrap this PR up...? 😄

PatKamin: I've updated the PR with a quick solution, should work now as expected.

kswiecicki: I thought that setting `CXX_HAS_SAN_ARGS` to a default value and leaving the `AND` as is was sufficient.

kswiecicki (approved):

cmake/helpers.cmake
lukaszstolarczuk: Now, the code looks neat, but TSAN started failing 😄

igchor (approved):

ldorau:
The CI builds are failing because of https://github.com/oneapi-src/unified-memory-framework/pull/199

PatKamin:
Turning to draft until #199 is merged.

bratpiorka:
@PatKamin #199 is merged

PatKamin:
Rebased

kswiecicki (approved):

bratpiorka (approved):

PR #177, opened 2024-01-26
Add linear base allocator

ldorau:
Add a MT-safe linear base allocator. It is useful for a few, small and different size allocations for a most/whole life-time of an application (free() is not available).

It is needed by https://github.com/oneapi-src/unified-memory-framework/pull/178 and https://github.com/oneapi-src/unified-memory-framework/pull/170.

PR #178, opened 2024-01-26
Use base allocator in critnib

ldorau:
Requires:
- [x] #177

### DIFF
--- a/src/critnib/critnib.c
+++ b/src/critnib/critnib.c
@@ -57,6 +57,8 @@
 #include <stdbool.h>
 #include <stddef.h>
 
+#include "base_alloc.h"
+#include "base_alloc_linear.h"
 #include "critnib.h"
 #include "utils_common.h"
 #include "utils_concurrency.h"
@@ -126,7 +128,11 @@ struct critnib {
 
     uint64_t remove_count;
 
-    struct os_mutex_t *mutex; /* writes/removes */
+    struct os_mutex_t mutex; /* writes/removes */
+
+    umf_ba_linear_pool_t *pool_linear;
+    umf_ba_pool_t *pool_nodes;
+    umf_ba_pool_t *pool_leaves;
 };
 
 /*
@@ -176,37 +182,63 @@ static inline unsigned slice_index(word key, sh_t shift) {
  * critnib_new -- allocates a new critnib structure
  */
 struct critnib *critnib_new(void) {
-    struct critnib *c = Zalloc(sizeof(struct critnib));
-    if (!c) {
+    umf_ba_linear_pool_t *pool_linear =
+        umf_ba_linear_create(0 /* minimal pool size */);
+    if (!pool_linear) {
         return NULL;
     }
 
-    c->mutex = util_mutex_create();
-    if (!c->mutex) {
-        free(c);
-        return NULL;
+    struct critnib *c =
+        umf_ba_linear_alloc(pool_linear, sizeof(struct critnib));
+    if (!c) {
+        goto err_destroy_pool_linear;
+    }
+
+    c->pool_linear = pool_linear;
+
+    void *mutex_ptr = util_mutex_init(&c->mutex);
+    if (!mutex_ptr) {
+        goto err_destroy_pool_linear;
+    }
+
+    c->pool_nodes = umf_ba_create(sizeof(struct critnib_node));
+    if (!c->pool_nodes) {
+        goto err_util_mutex_destroy;
+    }
+
+    c->pool_leaves = umf_ba_create(sizeof(struct critnib_leaf));
+    if (!c->pool_leaves) {
+        goto err_destroy_pool_nodes;
     }
 
     VALGRIND_HG_DRD_DISABLE_CHECKING(&c->root, sizeof(c->root));
     VALGRIND_HG_DRD_DISABLE_CHECKING(&c->remove_count, sizeof(c->remove_count));
 
     return c;
+
+err_destroy_pool_nodes:
+    umf_ba_destroy(c->pool_nodes);
+err_util_mutex_destroy:
+    util_mutex_destroy_not_free(&c->mutex);
+err_destroy_pool_linear:
+    umf_ba_linear_destroy(pool_linear); // free all its allocations and destroy
+    return NULL;
 }
 
 /*
  * internal: delete_node -- recursively free (to malloc) a subtree
  */
-static void delete_node(struct critnib_node *__restrict n) {
+static void delete_node(struct critnib *c, struct critnib_node *__restrict n) {
     if (is_leaf(n)) {
-        Free(to_leaf(n));
+        umf_ba_free(c->pool_leaves, to_leaf(n));
     } else {
         for (int i = 0; i < SLNODES; i++) {
             if (n->child[i]) {
-                delete_node(n->child[i]);
+                delete_node(c, n->child[i]);
             }
         }
 
-        Free(n);
+        umf_ba_free(c->pool_nodes, n);
     }
 }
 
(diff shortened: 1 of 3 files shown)


src/critnib/critnib.c
```diff
@@ -176,37 +182,64 @@ static inline unsigned slice_index(word key, sh_t shift) {
...
-    c->mutex = util_mutex_create();
+    struct critnib *c =
+        umf_ba_linear_alloc(pool_linear, sizeof(struct critnib));
+    if (!c) {
+        goto err_destroy_pool_linear;
+    }
+
+    c->pool_linear = pool_linear;
+
+    c->mutex = util_mutex_init(
```
bratpiorka, line 199: please split the allocation and initialization and handle possible errors from both calls

src/critnib/critnib.c
```diff
@@ -176,37 +182,64 @@ static inline unsigned slice_index(word key, sh_t shift) {
...
 }
 
 /*
  * internal: delete_node -- recursively free (to malloc) a subtree
  */
-static void delete_node(struct critnib_node *__restrict n) {
+static void delete_node(struct critnib *c, struct critnib_node *__restrict n) {
     if (is_leaf(n)) {
-        Free(to_leaf(n));
+        umf_ba_free(c->pool_leaves, to_leaf(n));
```
bratpiorka, line 234: I think we could remove also Malloc, Free and Zalloc symbols from code

bratpiorka (dismissed):

bratpiorka (commented):
others - please review as I'm on vacation

src/critnib/critnib.c
ldorau: Done

src/critnib/critnib.c
ldorau: None of these functions is used in the critnib code.

src/critnib/critnib.c
```diff
@@ -176,37 +182,68 @@ static inline unsigned slice_index(word key, sh_t shift) {
...
-    c->mutex = util_mutex_create();
+    struct critnib *c =
+        umf_ba_linear_alloc(pool_linear, sizeof(struct critnib));
+    if (!c) {
+        goto err_destroy_pool_linear;
+    }
+
+    c->pool_linear = pool_linear;
+
+    void *mutex_ptr = umf_ba_linear_alloc(pool_linear, util_mutex_get_size());
```
igchor, line 199: nit: since you are modyfing this part anyway, you can actually just refactor this part and store the mutex directly in the critnib stuct (just use util_mutex_init, without allocation). This way we could also get rid of the util_mutex_* functions that allocate/free memory and only leave (util_mutex_destroy and util_mutex_create)

src/critnib/critnib.c
igchor: Then we can also remove them from utils_common.h I think only critnib used them.

src/critnib/critnib.c
```diff
@@ -176,37 +182,68 @@ static inline unsigned slice_index(word key, sh_t shift) {
...
+    }
+
+    c->mutex = util_mutex_init(mutex_ptr);
     if (!c->mutex) {
-        free(c);
-        return NULL;
+        goto err_destroy_pool_linear;
+    }
+
+    c->pool_nodes = umf_ba_create(sizeof(struct critnib_node));
```
igchor, line 209: What are the sizes of `critnib_node` and `critnib_leaf`? I'm wondering if we should align them up to a cacheline to avoid false sharing.

src/critnib/critnib.c
```diff
@@ -176,37 +182,68 @@ static inline unsigned slice_index(word key, sh_t shift) {
  * critnib_new -- allocates a new critnib structure
  */
 struct critnib *critnib_new(void) {
-    struct critnib *c = Zalloc(sizeof(struct critnib));
-    if (!c) {
+    umf_ba_linear_pool_t *pool_linear =
+        umf_ba_linear_create(0 /* minimal pool size */);
```
igchor, line 186: shouldn't we explicitly request a proper size? It's probably unlikely but it may happen that those metada will take more than 4K in future

src/critnib/critnib.c
ldorau: Done

src/critnib/critnib.c
ldorau: Removed

src/critnib/critnib.c
ldorau: sizeof(struct critnib_node) = 144,
sizeof(struct critnib_leaf) = 16

src/critnib/critnib.c
ldorau: We allocate only `sizeof(struct critnib) = 352` bytes now, so 4K is more than enough now. We can increase that if we need more than 4K at any moment.

igchor: But will you remember to do that? ;d The 4k is not documented anywhere so perhaps we should at least add some comment / assert for the size as a reminder?

src/critnib/critnib.c
igchor: Ok, then we can probably leave it as is for now. We can consider this if we ever find that we need to optimize critnib.

igchor (approved):

src/critnib/critnib.c
ldorau: It will start failing then.

igchor: Ok, I forgot we don't have any builtin static_assert in C so we can only check it at runtime anyway.

PatKamin (approved):

KFilipek (changes requested):
Generally okay, but if the function `umf_ba_linear_destroy` needs an explanation every time that it also destroys the parent (the provider), it should have another name. If it's not the topic for this PR, please do it separately, I would be grateful.

ldorau:
> Generally okay, but if the function `umf_ba_linear_destroy` needs an explanation every time that it also destroys the parent (the provider), it should have another name. If it's not the topic for this PR, please do it separately, I would be grateful.

@KFilipek 
1) It does NOT need an explanation every time - it is only my good will that I added those comments.
2) It does NOT destroy any parent. `umf_ba_linear_destroy()` only **destroys** a mutex and **destroys** a pool itself (unmaps a range of memory) - so what would be a better name? - see its implementation:
```c
void umf_ba_linear_destroy(umf_ba_linear_pool_t *pool) {
    util_mutex_destroy_not_free(&pool->metadata.lock);
    ba_os_free(pool, pool->metadata.pool_size);
}
```

src/critnib/critnib.c
ldorau: Fixed

bb-umf:
> > Generally okay, but if the function `umf_ba_linear_destroy` needs an explanation every time that it also destroys the parent (the provider), it should have another name. If it's not the topic for this PR, please do it separately, I would be grateful.
> 
> @KFilipek
> 
> 1. It does NOT need an explanation every time - it is only my good will that I added those comments ;-) - I can remove them all.
> 2. It does NOT destroy any parent. `umf_ba_linear_destroy()` only **destroys** a mutex and **destroys** a pool itself (unmaps a range of memory) - so what would be a better name? - see its implementation (`ba_os_free()` is a wrapper for `munmap()`):
> 
> ```c
> void umf_ba_linear_destroy(umf_ba_linear_pool_t *pool) {
>     util_mutex_destroy_not_free(&pool->metadata.lock);
>     ba_os_free(pool, pool->metadata.pool_size);
> }
> ```

Currently, I'm not sure what I've seen, but you're completely right about your changes. Sorry for that. LGTM.

KFilipek:
> > Generally okay, but if the function `umf_ba_linear_destroy` needs an explanation every time that it also destroys the parent (the provider), it should have another name. If it's not the topic for this PR, please do it separately, I would be grateful.
> 
> @KFilipek
> 
> 1. It does NOT need an explanation every time - it is only my good will that I added those comments ;-) - I can remove them all.
> 2. It does NOT destroy any parent. `umf_ba_linear_destroy()` only **destroys** a mutex and **destroys** a pool itself (unmaps a range of memory) - so what would be a better name? - see its implementation (`ba_os_free()` is a wrapper for `munmap()`):
> 
> ```c
> void umf_ba_linear_destroy(umf_ba_linear_pool_t *pool) {
>     util_mutex_destroy_not_free(&pool->metadata.lock);
>     ba_os_free(pool, pool->metadata.pool_size);
> }
> ```

Currently, I'm not sure what I've seen previously, but you're right about your changes. Sorry for that.

KFilipek (approved):

ldorau:
@igchor Can we merge this PR?

PR #182, opened 2024-01-26
Add job with Intel C/C++ Compiler

KFilipek:
# Add job with Intel C/C++ Compiler
## Description
This PR introduces Intel C/C++ Compiler provided as a Docker image.

Used VM: Ubuntu 22.04
Used container: intel/oneapi:latest (icx/icpx 2024.02)

Other jobs are VM-based (Ubuntu 20.04/22.04, Windows 2019 or MacOSX) and those runners have other paths (github.workspace etc.) than steps executed under Docker in those VM's.
To avoid ambiguity there is a mounting between host and container.

### Notice 1:
The privileged option is related to NUMA requirements. Maybe it is possible to limit permission just by "--sys-cap SYS_CAP_NICE", but I don't test this option.
### Notice 2:
As we talked, it will be refactored and please keep it in mind during review.

### DIFF
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -7,6 +7,70 @@ env:
   INSTL_DIR : "${{github.workspace}}/build/install-dir"
 
 jobs:
+  icx-build:
+    name: Build - Intel C++ Compiler
+    strategy:
+      matrix:
+        build_type: [Debug, Release]
+        compiler: [{c: icx, cxx: icpx}]
+        pool_tracking: ['ON', 'OFF']
+        shared_library: ['OFF']
+        os_provider: ['ON']
+        sanitizers: [{asan: OFF, ubsan: OFF, tsan: OFF}]
+    runs-on: ubuntu-22.04
+    container:
+      image: intel/oneapi:latest
+      volumes:
+        - ${{github.workspace}}:${{github.workspace}}
+      options: "--privileged"
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v4
+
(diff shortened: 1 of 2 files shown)


.github/workflows/basic.yml
```diff
@@ -7,6 +7,70 @@ env:
   INSTL_DIR : "${{github.workspace}}/build/install-dir"
 
 jobs:
+  icx-build:
```
bratpiorka, line 10: I think it would be better to add icx/icpx as a "include" to ubuntu matrix (similar to clang or gcc-7) if this is possible

ldorau (approved):

lukaszstolarczuk:
> ### Notice 2:
> As we talked, it will be refactored and please keep it in mind during review.

What do you mean by that second notice?

KFilipek:
> > ### Notice 2:
> > As we talked, it will be refactored and please keep it in mind during review.
> 
> What do you mean exactly by that second notice?

As we agreed in the last meeting, workflow files need some refactoring.

lukaszstolarczuk:
> > > ### Notice 2:
> > > As we talked, it will be refactored and please keep it in mind during review.
> > 
> > 
> > What do you mean exactly by that second notice?
> 
> As we agreed in the last meeting, workflow files need some refactoring.

heh, you didn't give me much more details 😄

lukaszstolarczuk (approved):

.github/workflows/basic.yml
KFilipek: It can extend execution time a little bit, but I can do it on another PR.

PR #185, opened 2024-01-30
Add proxy_pool

igchor:
a pool that forwards all (supported) requests to the memory provider.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,12 @@ A memory provider that provides memory from an operating system.
 
 ## Memory pool managers
 
+### proxy_pool (part of libumf)
+
+This memory pool is distributed as part of libumf. It forwards all requests to the underlying
+memory provider. Currently umfPoolRealloc, umfPoolCalloc and umfPoolMallocUsableSize functions
+are not supported by the proxy pool.
+
 ### libumf_pool_disjoint
 
 TODO: Add a description
--- a/include/umf/pools/pool_proxy.h
+++ b/include/umf/pools/pool_proxy.h
@@ -1,0 +1,27 @@
+/*
+ *
+ * Copyright (C) 2024 Intel Corporation
+ *
+ * Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ *
+ */
+
+#ifndef UMF_PROXY_MEMORY_POOL_H
+#define UMF_PROXY_MEMORY_POOL_H 1
+
+#include <umf/base.h>
+#include <umf/memory_pool.h>
+#include <umf/memory_provider.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+UMF_EXPORT umf_memory_pool_ops_t *umfProxyPoolOps(void);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* UMF_PROXY_MEMORY_POOL_H */
--- a/src/pool/pool_proxy.c
+++ b/src/pool/pool_proxy.c
@@ -1,0 +1,126 @@
+/*
+ *
+ * Copyright (C) 2024 Intel Corporation
+ *
+ * Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ *
+ */
+
+#include <umf/memory_pool_ops.h>
+#include <umf/pools/pool_proxy.h>
+
+#include <assert.h>
+#include <stdlib.h>
+
+#include "utils_common.h"
+
+static __TLS umf_result_t TLS_last_allocation_error;
+
+struct proxy_memory_pool {
+    umf_memory_provider_handle_t hProvider;
+};
+
+static umf_result_t
+proxy_pool_initialize(umf_memory_provider_handle_t hProvider, void *params,
+                      void **ppPool) {
+    (void)params; // unused
+
+    // TODO: use ba_alloc here
+    struct proxy_memory_pool *pool = malloc(sizeof(struct proxy_memory_pool));
+    if (!pool) {
+        return UMF_RESULT_ERROR_OUT_OF_HOST_MEMORY;
+    }
+
+    pool->hProvider = hProvider;
+    *ppPool = (void *)pool;
+
+    return UMF_RESULT_SUCCESS;
+}
+
+static void proxy_pool_finalize(void *pool) { free(pool); }
+
+static void *proxy_aligned_malloc(void *pool, size_t size, size_t alignment) {
+    assert(pool);
+
+    void *ptr;
+    struct proxy_memory_pool *hPool = (struct proxy_memory_pool *)pool;
+
+    umf_result_t ret =
+        umfMemoryProviderAlloc(hPool->hProvider, size, alignment, &ptr);
+    if (ret != UMF_RESULT_SUCCESS) {
+        TLS_last_allocation_error = ret;
+        return NULL;
+    }
+
+    TLS_last_allocation_error = UMF_RESULT_SUCCESS;
+    return ptr;
+}
+
+static void *proxy_malloc(void *pool, size_t size) {
+    assert(pool);
+
+    return proxy_aligned_malloc(pool, size, 0);
+}
+
+static void *proxy_calloc(void *pool, size_t num, size_t size) {
+    assert(pool);
+
+    (void)pool;
+    (void)num;
+    (void)size;
+
+    // Currently we cannot implement calloc in a way that would
+    // work for memory that is inaccessible on the host
+    TLS_last_allocation_error = UMF_RESULT_ERROR_NOT_SUPPORTED;
+    return NULL;
+}
+
+static void *proxy_realloc(void *pool, void *ptr, size_t size) {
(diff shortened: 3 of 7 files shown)


igchor:
Ok, I was debugging the windows failure and it looks like there is a problem with exporting global variables (like OPS vars).

Accordingly to the [documentation](https://learn.microsoft.com/en-us/cpp/build/importing-into-an-application-using-declspec-dllimport?view=msvc-170): "Using __declspec(dllimport) is optional on function declarations, but the compiler produces more efficient code if you use this keyword. **However, you must use __declspec(dllimport) for the importing executable to access the DLL's public data symbols and objects**".

So, this means that we would need to define `UMF_EXPORT` like this:
```
#if defined(UMF_EXPORTS) && defined(UMF_SHARED_LIBRARY)
    #define UMF_EXPORT __declspec(dllexport)
#elif defined(UMF_SHARED_LIBRARY)
    #define UMF_EXPORT __declspec(dllimport)
#endif
```

but... then we have a problem when we build UMF as a static library because using __declspel(dllimport) on symbols from static library will result in linking failure (see [here](https://stackoverflow.com/questions/4841719/dllimport-dllexport-and-static-libraries-compilation-under-visual-c)). To solve this users would need to define `UMF_SHARED_LIBRARY` in their application whenever linking with libumf as a shared library (or the other way around, define something for static library). 

I don't think forcing users to define this is a good ideas to perhaps we could just avoid global variables altogether? Instead of global UMF_*_POOL_OPS variables we could just define functions that would return proper ops. @pbalcer @bratpiorka @vinser52  any thoughts?

lplewa:
> Ok, I was debugging the windows failure and it looks like there is a problem with exporting global variables (like OPS vars).
> 
> Accordingly to the [documentation](https://learn.microsoft.com/en-us/cpp/build/importing-into-an-application-using-declspec-dllimport?view=msvc-170): "Using __declspec(dllimport) is optional on function declarations, but the compiler produces more efficient code if you use this keyword. **However, you must use __declspec(dllimport) for the importing executable to access the DLL's public data symbols and objects**".
> 
> So, this means that we would need to define `UMF_EXPORT` like this:
> 
> ```
> #if defined(UMF_EXPORTS) && defined(UMF_SHARED_LIBRARY)
>     #define UMF_EXPORT __declspec(dllexport)
> #elif defined(UMF_SHARED_LIBRARY)
>     #define UMF_EXPORT __declspec(dllimport)
> #endif
> ```
> 
> but... then we have a problem when we build UMF as a static library because using __declspel(dllimport) on symbols from static library will result in linking failure (see [here](https://stackoverflow.com/questions/4841719/dllimport-dllexport-and-static-libraries-compilation-under-visual-c)). To solve this users would need to define `UMF_SHARED_LIBRARY` in their application whenever linking with libumf as a shared library (or the other way around, define something for static library).
> 
> I don't think forcing users to define this is a good ideas to perhaps we could just avoid global variables altogether? Instead of global UMF_*_POOL_OPS variables we could just define functions that would return proper ops. @pbalcer @bratpiorka @vinser52 any thoughts?

Imho solution is to use DEF file on Windows instead of dllexport. It's similar to linux map file, and it's hides windows mess from header.

igchor:
> Imho solution is to use DEF file on Windows instead of dllexport. It's similar to linux map file, and it's hides windows mess from header.

@lplewa +1 for using .def files but accordingly to https://learn.microsoft.com/en-us/cpp/build/reference/exports?view=msvc-170 this won't really solve the issue here:

"When you export a variable from a DLL by using a .DEF file, you do not have to specify __declspec(dllexport) on the variable. However, in any file that uses the DLL, you must still use __declspec(dllimport) on the declaration of data".

igchor:
> > > Imho solution is to use DEF file on Windows instead of dllexport. It's similar to linux map file, and it's hides windows mess from header.
> > 
> > 
> > @lplewa +1 for using .def files but accordingly to https://learn.microsoft.com/en-us/cpp/build/reference/exports?view=msvc-170 this won't really solve the issue here:
> > "When you export a variable from a DLL by using a .DEF file, you do not have to specify __declspec(dllexport) on the variable. However, in any file that uses the DLL, you must still use __declspec(dllimport) on the declaration of data".
> 
> Windows....
> 
> Can we replace exported global variable with an exported "get" function, which returns global "nonexported" variable from dll?

Yeah, that was my idea as well - I think this should work (although I haven't tried that yet, and it's windows so anything is possible). I just wanted to ask first, and see if there's a better way.

lplewa:
I will vote to switch to def files, and instead exporting variables, export getters.

pbalcer:
> I will vote to switch to def files, and instead exporting variables, export getters.

👍

igchor:
Ok, I'll make a PR to replace all variables with getters. This is not strictly necessary for every variable since some of them are only in static libraries, e.g. disjoint_pool but I think it's better to be consistent + we may choose to build those libs as DLLs in future.

ldorau (approved):

src/pool/pool_proxy.c
```diff
@@ -0,0 +1,125 @@
...
+
+static void *proxy_calloc(void *pool, size_t num, size_t size) {
+    assert(pool);
+
+    (void)pool;
+    (void)num;
+    (void)size;
+
+    // Currently we cannot implement calloc in a way that would
+    // work for memory that is inaccessible on the host
```
bratpiorka, line 73: this is specific for some providers - why can't we rely on providers for this? What if I want to use a proxy provider for an OS provider?

test/memoryPoolAPI.cpp
```diff
@@ -51,48 +64,40 @@ TEST_F(test, memoryPoolTrace) {
...
-
-    ASSERT_EQ(providerCalls["alloc"], 2);
-    ASSERT_EQ(providerCalls.size(), provider_call_count);
-
-    umfPoolRealloc(tracingPool.get(), nullptr, 0);
-    ASSERT_EQ(poolCalls["realloc"], 1);
-    ASSERT_EQ(poolCalls.size(), ++pool_call_count);
-
-    ASSERT_EQ(providerCalls.size(), provider_call_count);
+    // TODO: calloc is not supported by proxy pool
```
bratpiorka, line 80: see my previous comment - we should get the info if calloc or realloc is supported

bratpiorka (changes requested):

src/pool/pool_proxy.c
igchor: How do you propose to do that? Providers do not expose calloc or realloc functions.

test/memoryPoolAPI.cpp
igchor: What do you mean by "we should get the info"? You will get UNSUPPORTED error status here.

src/pool/pool_proxy.c
```diff
@@ -0,0 +1,125 @@
...
+struct proxy_memory_pool {
+    umf_memory_provider_handle_t hProvider;
+};
+
+static umf_result_t
+proxy_pool_initialize(umf_memory_provider_handle_t hProvider, void *params,
+                      void **ppPool) {
+    (void)params; // unused
+
+    struct proxy_memory_pool *pool = malloc(sizeof(struct proxy_memory_pool));
```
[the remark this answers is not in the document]
igchor: Done.

include/umf/pools/pool_proxy.h
```diff
@@ -0,0 +1,26 @@
...
+#define UMF_PROXY_MEMORY_POOL_H 1
+
+#include <umf/memory_pool.h>
+#include <umf/memory_provider.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+umf_memory_pool_ops_t *umfProxyPoolOps(void);
```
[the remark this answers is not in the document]
igchor: Done. I've updated the readme but we need to figure out how to document pools/providers in general.

src/pool/pool_proxy.c
bratpiorka: Yes, right. But I think this could be confusing for someone used to malloc API but not having the knowledge about the providers. I think we should add this to the documentation.

README.md
```diff
@@ -37,6 +37,11 @@ A memory provider that provides memory from an operating system.
 
 ## Memory pool managers
 
+### proxy_pool (part of libumf)
+
+This memory pool is distributed as part of libumf. It forwards all requests to the underlying
+memory provider.
```
bratpiorka, line 43: Could you add an info here what functions are not supported and why? This is not obvious without knowledge of memory provider API.

test/memoryPoolAPI.cpp
bratpiorka: Hm after thinking about it, I came to the conclusion that a simple proxy pool should not be used here to test the tracking provider, as it does not test all functions. Could we just use some "malloc-pool" here? You could just use part of the old code, but rename the pool.

bratpiorka (changes requested):

test/memoryPoolAPI.cpp
igchor: Actually, I think we can still use the proxy pool here but just don't check whether the provider was called or not. So I've restored some of the original code here. 

This test only verifies that a proper umfPool* function has been called anyway. For functional testing, we have different tests.

README.md
igchor: Done.

bratpiorka (approved):

PR #188, opened 2024-01-30
[CI] Add first dockers and trivy workflow

lukaszstolarczuk:
// tested on my fork: https://github.com/lukaszstolarczuk/unified-memory-framework/actions/runs/7886306110/job/21519277801

### DIFF
--- a/.github/docker/README.md
+++ b/.github/docker/README.md
@@ -1,0 +1,52 @@
+# Content
+
+Dockerfiles and scripts placed in this directory are intended to be used as
+development process vehicles and part of continuous integration process.
+
+Images built out of those recipes may be used with Docker or podman as
+development environment. If you want to use below instructions with `podman`,
+simply replace word `docker` with `podman`.
+
+# How to build docker image
+
+To build docker image on local machine enter the root dir of the repository and execute:
+
+```sh
+docker build -t ur:ubuntu-22.04 -f .github/docker/ubuntu-22.04.Dockerfile .
+```
+
+To set any build time variable (e.g., an optional ARG from docker recipe), add to the command (after `build`), e.g.:
+
+```sh
+ --build-arg TEST_DEPS=""
+```
+
+One other example of using these extra build arguments are proxy settings. They are required for accessing network
+(e.g., to download dependencies within docker), if a host is using a proxy server. Example usage:
+
+```sh
+ --build-arg https_proxy=http://proxy.com:port --build-arg http_proxy=http://proxy.com:port
+```
+
+# How to use docker image
+
+To run docker container (using the previously built image) execute:
+
+```sh
+docker run --shm-size=4G -v /your/workspace/path/:/opt/workspace:z -w /opt/workspace/ -it ur:ubuntu-22.04 /bin/bash
+```
+
+To set (or override) any docker environment variable, add to the command (after `run`):
+
+```sh
+ -e ENV_VARIABLE=VALUE
+```
+
+To start as a non-root user (created within our Dockerfiles), add to the command (after `run`):
+
+```sh
+ --user test_user
+```
+
+If you want to run a docker container using your specific user, please follow up, e.g.,
+with [this article](https://jtreminio.com/blog/running-docker-containers-as-current-host-user/).
--- a/.github/docker/ubuntu-20.04.Dockerfile
+++ b/.github/docker/ubuntu-20.04.Dockerfile
@@ -1,0 +1,63 @@
+# Copyright (C) 2024 Intel Corporation
+# Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#
+# Dockerfile - a 'recipe' for Docker to build an image of ubuntu-based
+#              environment for building the Unified Memory Framework project.
+#
+
+# Pull base image ("20.04")
+FROM registry.hub.docker.com/library/ubuntu@sha256:f2034e7195f61334e6caff6ecf2e965f92d11e888309065da85ff50c617732b8
+
+# Set environment variables
+ENV OS ubuntu
+ENV OS_VER 20.04
+ENV NOTTY 1
+ENV DEBIAN_FRONTEND noninteractive
+
+# Base development packages
+ARG BASE_DEPS="\
+	build-essential \
+	cmake \
+	git"
+
+# UMF's dependencies
+ARG UMF_DEPS="\
+	libjemalloc-dev \
+	libhwloc-dev \
+	libtbb-dev"
+
+# Dependencies for tests (optional)
+ARG TEST_DEPS="\
+	libnuma-dev"
+
+# Miscellaneous for our builds/CI (optional)
+ARG MISC_DEPS="\
+	clang \
+	g++-7 \
+	python3-pip \
+	sudo \
+	whois"
+
+# Update and install required packages
+RUN apt-get update \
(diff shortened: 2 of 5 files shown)


.github/docker/README.md
```diff
@@ -0,0 +1,43 @@
+# Content
+
+Dockerfiles and scripts placed in this directory are intended to be used as
+development process vehicles and part of continuous integration process.
+
+Images built out of those recipes may be used with Docker or podman as
+development environment.
```
PatKamin, line 7: Add a note that below are instructions for Docker users, or sth like that.

.github/docker/ubuntu-20.04.Dockerfile
```diff
@@ -0,0 +1,62 @@
...
+	cmake"
+
+# UMF's dependencies
+ARG UMF_DEPS="\
+	libjemalloc-dev \
+	libhwloc-dev \
+	libtbb-dev"
+
+# Dependencies for tests (optional)
+ARG TEST_DEPS="\
```
PatKamin, line 31: How to disable these optional installs?

lukaszstolarczuk: it's described in Readme, now

.github/docker/README.md
lukaszstolarczuk: do you mean I should point somehow this is for docker and not podman? if so, I think it's rather self explanatory 😉 esp. with `podman` being (from CLI perspective) just an alias for `docker`, so all command should work with just a command name change

PatKamin: Yes, I know that. But I doubt that someone that don't know hot to build a docker image will know it. Just add a mention that these are docker commands but should work for podman as well or sth like that. Or ignore my comment if you wish, it's not superimportant.

.github/workflows/trivy.yml
```diff
@@ -0,0 +1,39 @@
+# Runs linter for Docker files
+name: Trivy
```
PatKamin, line 2: This should be added to `pr_push.yml` as a requirement for the Build workflow, on the same level as Spellcheck.

.github/docker/README.md
lukaszstolarczuk: Done.

.github/workflows/trivy.yml
lukaszstolarczuk: actually we can't do this right now. We did the same with CodeQL, due to issue we encountered with scorecard and permissions. I copy-pasted the comment from CodeQL workflow.

PatKamin (approved):

.trivyignore
```diff
@@ -0,0 +1,10 @@
+# Docs: https://aquasecurity.github.io/trivy/latest/docs/configuration/filtering/#trivyignore
+
+# In docker files:
+# non-root user is always created within docker, but we switch it only in CI workflows;
+# not enforcing non-root user makes it easier for developers to use their own users in local container
+AVD-DS-0002
+
+# In docker files:
+# HEALTHCHECK is not required for development, nor in CI (failed docker = failed CI)
```
wlemkows, line 9: PSE approval is required for:
Hadolint YAML configuration files that disable any built-in rules. (I assume the same requirement will apply to Trivy)

wlemkows (changes requested):

.github/workflows/trivy.yml
```diff
@@ -0,0 +1,46 @@
...
+# Due to lower score on Scorecard we're running this separately from
+# "PR/push" workflow. For some reason permissions weren't properly set
+# or recognized (by Scorecard). If Scorecard changes its behavior we can
+# use 'workflow_call' trigger.
+on:
+  push:
+  pull_request:
+    paths:
+      - '.github/docker/*Dockerfile'
+      - '.github/workflows/trivy.yml'
```
KFilipek, line 13: Good idea +1

.github/docker/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+```sh
+ -e ENV_VARIABLE=VALUE
+```
+
+To start as a non-root user, add to the command (after `run`):
+
+```sh
+ --user test_user
+```
```
KFilipek, line 49: It's more complicated because Linux doesn't rely on user names at the filesystem level, you will encounter many problems with user permissions.
Look at the article:
https://jtreminio.com/blog/running-docker-containers-as-current-host-user/

.trivyignore
wlemkows: Since this task is blocking the other, let's just merge it as it is. There is still a lot of time left until the release, so we will certainly be able to address these problems later.

lukaszstolarczuk: I created an issue to follow this up: https://github.com/oneapi-src/unified-memory-framework/issues/262

.github/docker/README.md
lukaszstolarczuk: `test_user` is created within our dockers as test users 😉 

I've added a link to follow up, if someone needs it.

.github/workflows/trivy.yml
lukaszstolarczuk: 😺

.github/docker/README.md
```diff
@@ -0,0 +1,49 @@
...
+```sh
+ --build-arg TEST_DEPS=""
+```
+
+One other example of using these extra build arguments are proxy settings. They are required for accessing network
+(e.g., to download dependencies within docker), if a host is using a proxy server. Example usage:
+
+```sh
+ --build-arg https_proxy=http://proxy.com:port --build-arg http_proxy=http://proxy.com:port
+```
```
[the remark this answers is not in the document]
lukaszstolarczuk: I guess there are multiple ways to solve the proxy problem. This line is just an example of using build args.

.github/docker/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+# How to build docker image
+
+To build docker image on local machine enter the root dir of the repository and execute:
+
+```sh
+docker build -t ur:ubuntu-22.04 -f .github/docker/ubuntu-22.04.Dockerfile .
+```
+
+To set any build time variable (e.g., an optional ARG from docker recipe), add to the command (after `build`), e.g.:
```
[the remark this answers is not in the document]
lukaszstolarczuk: I believe this is proper English 😉 both of these `e.g.`s are proper examples

.github/docker/README.md
```diff
@@ -0,0 +1,49 @@
+# Content
+
+Dockerfiles and scripts placed in this directory are intended to be used as
+development process vehicles and part of continuous integration process.
+
+Images built out of those recipes may be used with Docker or podman as
+development environment. If you want to use below instructions with `podman`,
+simply replace word `docker` with `podman`.
```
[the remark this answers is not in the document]
lukaszstolarczuk: this little README is not a comprehensive solution to all problems 😉

.github/docker/README.md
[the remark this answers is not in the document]
KFilipek: Yes, there are. You've mentioned previously that "this little README is not a comprehensive solution to all problems", so there is no need to solve proxy issues here.

.trivyignore
bratpiorka: @wlemkows I am closing this conversation as we have agreed on the next steps

.github/docker/README.md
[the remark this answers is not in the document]
bratpiorka: @KFilipek If you think it is worth it, prepare a proposal on how exactly you would write this paragraph

KFilipek (approved):

PR #194, opened 2024-01-31
Fix the protection_flag_none test

ldorau:
The test:
```c
EXPECT_DEATH(write_memory("write string"), "");
```
was never executed, because it always died at the previous one:
```c
EXPECT_DEATH(read_memory(), "");
```

### DIFF
--- a/test/provider_os_memory_config.cpp
+++ b/test/provider_os_memory_config.cpp
@@ -66,7 +66,7 @@ struct providerConfigTest : testing::Test {
     }
 };
 
-TEST_F(providerConfigTest, protection_flag_none) {
+TEST_F(providerConfigTest, protection_flag_none_read) {
     // pages may not be accessed - PROT_NONE
     params.protection = UMF_PROTECTION_NONE;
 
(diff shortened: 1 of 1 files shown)


PatKamin:
I disagree. When debugging locally this test run, I see that both assertions are completed. The `EXPECT_DEATH` macro, according to the [documentation](https://google.github.io/googletest/reference/assertions.html#death) spawns a child process to verify death scenario. After that, the test case continues to the second `EXPECT_DEATH` test.

ldorau:
> I disagree. When debugging locally this test run, I see that both assertions are completed. The `EXPECT_DEATH` macro, according to the [documentation](https://google.github.io/googletest/reference/assertions.html#death) spawns a child process to verify death scenario. After that, the test case continues to the second `EXPECT_DEATH` test.
> 
> See in an [example](https://github.com/PatKamin/unified-memory-framework/actions/runs/7727566700/job/21066429733) branch with added a failing assertion at the end of this test fixture that the test case goes through both `EXPECT_DEATH` tests and fails in the added assertion `ASSERT_EQ(0, 1)`.

You are right. Thanks! Closing.

PR #197, opened 2024-01-31
Add linker map and .def file

igchor:


src/umf.def
lukaszstolarczuk, line 1: shouldn't it be called `libumf.def`? (and the same for `.map`)

cmake/helpers.cmake
```diff
@@ -94,14 +94,33 @@ endfunction()
 function(add_umf_library)
     # NAME - a name of the library
     # TYPE - type of the library (shared or static)
+    #        if shared library, LINUX_MAP_FILE and WINDOWS_DEF_FILE must also be specified
     # SRCS - source files
     # LIBS - libraries to be linked with
-    set(oneValueArgs NAME TYPE)
+    # LINUX_MAP_FILE - paths to linux linker map (.map) file
```
lukaszstolarczuk, line 100: `paths` -> `path`

.github/workflows/basic.yml
```diff
@@ -204,6 +204,7 @@ jobs:
         pool_tracking: ['ON', 'OFF']
         shared_library: ['OFF']
         sanitizers: [{asan: OFF}]
+        os_provider: ['ON']
```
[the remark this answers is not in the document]
ldorau: (it is used only in the matrix jobs, in the include jobs remains unset ... )

src/umf.def
igchor: Done. Although we can name it however we want, I think.

cmake/helpers.cmake
igchor: Done.

.github/workflows/basic.yml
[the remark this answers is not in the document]
igchor: Ok, I thought it worked differently, done.

ldorau (approved):

.github/workflows/basic.yml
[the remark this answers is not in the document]
lukaszstolarczuk: heh, no worries, we thought so too, at first

lukaszstolarczuk (commented):
👍

.github/workflows/basic.yml
```diff
@@ -204,28 +204,39 @@ jobs:
...
         include:
           - os: 'windows-2022'
             build_type: Release
             compiler: {c: clang-cl, cxx: clang-cl}
             pool_tracking: 'ON'
+            os_provider: ['ON']
```
lukaszstolarczuk, line 213: `['ON']` -> `'ON'` 😉

lukaszstolarczuk (changes requested):

src/libumf.def
DamianDuy, line 1: Shouldn't the `umfOsMemoryProviderOps` also be added to this file?

.github/workflows/basic.yml
igchor: Done.

src/libumf.def
igchor: No, at least not yet. Currently we do not build OS provider on windows and we cannot put non-existing symbols to libumf.def

src/libumf.map
```diff
@@ -0,0 +1,39 @@
...
+        umfPoolCalloc;
+        umfPoolCreate;
+        umfPoolDestroy;
+        umfPoolFree;
+        umfPoolGetLastAllocationError;
+        umfPoolGetMemoryProvider;
+        umfPoolMalloc;
+        umfPoolMallocUsableSize;
+        umfPoolRealloc;
+        umfOsMemoryProviderOps;
```
bratpiorka, line 32: what if we disable OS provider in cmake?

igchor: on linux, that's fine, if there's no such symbol in the .so linker will just ignore it

bratpiorka: ok but do we plan to generate these files in runtime in the future?

igchor: Iit depends on whether we will need to support enabling/disabling different features. For example, OS provider (but I think, eventually, we'll want OS provider to always be enabled.

bratpiorka:
it would be really great if you split this PR to smaller single-commit PRs

igchor:
> it would be really great if you split this PR to smaller single-commit PRs

Ok, done:
https://github.com/oneapi-src/unified-memory-framework/pull/200
https://github.com/oneapi-src/unified-memory-framework/pull/201

bratpiorka:
@igchor please rebase + change the name of this PR + could you also extract the changes in the github action file into a separate PR?

.github/workflows/basic.yml
lukaszstolarczuk: not quite - pls only update in `include`s, not in `matrix` (a list is required there)

igchor:
> @igchor please rebase + change the name of this PR + could you also extract the changes in the github action file into a separate PR?

I'll do that once https://github.com/oneapi-src/unified-memory-framework/pull/200 is merged

igchor:
> @igchor please rebase + change the name of this PR + could you also extract the changes in the github action file into a separate PR?

Done - rebased and changed the name of the PR. I've left the github action changes though since they are mostly to make sure that adding .def/.map file doesn't break anything.

bratpiorka:
@igor please resolve the issue pointed by @lukaszstolarczuk 
also, I think it would be good to generate def/map files in the future and add a test for that - I will create a JIRA task for this

ldorau (approved):

lukaszstolarczuk:
besides the issue (mentioned by Rafał) blocking CI, lgtm

.github/workflows/basic.yml
igchor: Ok, done.

igchor:
> @igor please resolve the issue pointed by @lukaszstolarczuk also, I think it would be good to generate def/map files in the future and add a test for that - I will create a JIRA task for this

Done.

igchor:
Hm, there's now another problem after we started using base alloc in jemalloc/scalable pool implementations. For dynamic library we should hide base_alloc symbols from libumf.so but that means we would need to build jemalloc/scalable pool with base_alloc sources (same as tests in: https://github.com/oneapi-src/unified-memory-framework/pull/200).

However, for static libumf, we do not control symbol visibility so if both libumf.a and jemalloc/scalable_pool.a are build with base_alloc we'll have conflicting symbols I think. Any ideas how we can solve this? Should we make base_alloc a separate dynamic library?

bratpiorka:
> Hm, there's now another problem after we started using base alloc in jemalloc/scalable pool implementations. For dynamic library we should hide base_alloc symbols from libumf.so but that means we would need to build jemalloc/scalable pool with base_alloc sources (same as tests in: #200).
> 
> However, for static libumf, we do not control symbol visibility so if both libumf.a and jemalloc/scalable_pool.a are build with base_alloc we'll have conflicting symbols I think. Any ideas how we can solve this? Should we make base_alloc a separate dynamic library?

can't we check for this case in cmake? and do not include base_alloc sources in static jemalloc/scallable pools? These libs will never be used without libumf.a

igchor:
> > Hm, there's now another problem after we started using base alloc in jemalloc/scalable pool implementations. For dynamic library we should hide base_alloc symbols from libumf.so but that means we would need to build jemalloc/scalable pool with base_alloc sources (same as tests in: #200).
> > However, for static libumf, we do not control symbol visibility so if both libumf.a and jemalloc/scalable_pool.a are build with base_alloc we'll have conflicting symbols I think. Any ideas how we can solve this? Should we make base_alloc a separate dynamic library?
> 
> can't we check for this case in cmake? and do not include base_alloc sources in static jemalloc/scallable pools? These libs will never be used without libumf.a

Ok, done. I've also had to add ba_alloc init code to jemalloc and scalable pools (see topmost commit). Still, this is not an ideal solution since our jemalloc/scalable pool libs will only work with libum that they were build with (.e.g. if jemalloc_pool.a was build with libumf.so it won't work with libumf.a).

bratpiorka:
> > > Hm, there's now another problem after we started using base alloc in jemalloc/scalable pool implementations. For dynamic library we should hide base_alloc symbols from libumf.so but that means we would need to build jemalloc/scalable pool with base_alloc sources (same as tests in: #200).
> > > However, for static libumf, we do not control symbol visibility so if both libumf.a and jemalloc/scalable_pool.a are build with base_alloc we'll have conflicting symbols I think. Any ideas how we can solve this? Should we make base_alloc a separate dynamic library?
> > 
> > 
> > can't we check for this case in cmake? and do not include base_alloc sources in static jemalloc/scallable pools? These libs will never be used without libumf.a
> 
> Ok, done. I've also had to add ba_alloc init code to jemalloc and scalable pools (see topmost commit). Still, this is not an ideal solution since our jemalloc/scalable pool libs will only work with libum that they were build with (.e.g. if jemalloc_pool.a was build with libumf.so it won't work with libumf.a).

ok, thanks. About the scnario you mentioned - currently there is no way to build only jemalloc_pool.a without building libumf. So I think the problem here is very limited

lukaszstolarczuk (approved):

PR #199, opened 2024-02-02
Fix the test of the linear base allocator (test_base_alloc_linear.c)

ldorau:


ldorau:
It is required by https://github.com/oneapi-src/unified-memory-framework/pull/167

test/test_base_alloc_linear.c
```diff
@@ -31,10 +31,18 @@ static void *start_routine(void *arg) {
     for (int i = 0; i < ITERATIONS; i++) {
         buffer[i].size = (rand() * MAX_ALLOCATION_SIZE) / RAND_MAX;
         buffer[i].ptr = umf_ba_linear_alloc(pool, buffer[i].size);
+        if (buffer[i].ptr == NULL) {
```
PatKamin, line 34: I'm not sure about that. This leaves the test successful even if all allocations will fail. I think that in this test case you should focus on having successful multiple allocations from several threads, not on filling the whole pool.

Won't just raising the size of the pool suffice?

bratpiorka: I agree - don't add this and add a check here for buffer[i].ptr != NULL. If we increased the pool size to the maximum possible size then we should never fail an allocation

bratpiorka (changes requested):

test/test_base_alloc_linear.c
ldorau: Replaced with assert.
Done.

bratpiorka (approved):

PR #200, opened 2024-02-02
Build base_alloc test with base_alloc sources

igchor:
base_alloc is an internal compontent of umf and it's symbols should not be visible outside.

This avoids linking error after introducing .def/.map files and limiting symbol visibility.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -118,9 +118,9 @@ endif()
 if(LINUX)
     # the base_alloc test uses linux pthreads
     add_umf_test(NAME base_alloc
-                SRCS test_base_alloc.c
+                SRCS ${BA_SOURCES} test_base_alloc.c
                 LIBS umf_utils)
     add_umf_test(NAME base_alloc_linear
-                SRCS test_base_alloc_linear.c
+                SRCS ${BA_SOURCES} test_base_alloc_linear.c
                 LIBS umf_utils)
 endif()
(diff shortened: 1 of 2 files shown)


test/CMakeLists.txt
```diff
@@ -115,9 +115,9 @@ endif()
 if(LINUX)
     # the base_alloc test uses linux pthreads
     add_umf_test(NAME base_alloc
-                SRCS test_base_alloc.c
+                SRCS ${BA_SOURCES} test_base_alloc.c
```
bratpiorka, line 118: I don't get it - you add these files both to libumf and tests?

bratpiorka (changes requested):

test/CMakeLists.txt
igchor: Yes, as I wrote in the commit description, BA was being compiled as part of libumf (not a separate library) but BA symbols should not be visible outside of the libumf.so/.a/.dll. On windows, since those symbols were not exported, the linking of this test would fail (same for linux after we introduce linker map). So, the solution is to either use those sources directly in tests (that's what I did) or make BA a static library (but I don't see a need for that right now).

ldorau (approved):

bratpiorka (approved):

ldorau:
@igchor rebase please

igchor:
> @igchor rebase please

Rebased.

PR #205, opened 2024-02-05
Add predefined HOST ALL memspace

kswiecicki:
- [x] Add tests
- [x] Add documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,6 +161,19 @@ The `UMF_BUILD_LIBUMF_POOL_SCALABLE` option has to be turned `ON` to build this 
 2) Required packages:
    - libtbb-dev (libraries: libtbbmalloc.so.2)
 
+### Memspaces (Linux-only)
+
+TODO: Add general information about memspaces.
+
+#### Host all memspace
+
+Memspace backed by all available NUMA nodes discovered on the platform. Can be retrieved
+using umfMemspaceHostAllGet.
+
+##### Requirements
+
+This feature relies on the OS memory provider.
+
 ## Contributions
 
 All contributions to the UMF project are most welcome! Before submitting
--- a/include/umf/memspace.h
+++ b/include/umf/memspace.h
@@ -44,6 +44,12 @@ umfMemoryProviderCreateFromMemspace(umf_memspace_handle_t hMemspace,
                                     umf_memspace_policy_handle_t hPolicy,
                                     umf_memory_provider_handle_t *hProvider);
 
+///
+/// \brief Retrieves predefined host all memspace.
+/// \return host all memspace handle on success or NULL on failure.
+///
+umf_memspace_handle_t umfMemspaceHostAllGet(void);
+
 #ifdef __cplusplus
 }
 #endif
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -52,12 +52,17 @@ set(UMF_SOURCES_MACOSX
     libumf_linux.c
 )
 
+# Compile definitions for UMF library
+# TODO: Cleanup the compile definitions across all the Cmake files
+set(UMF_COMPILE_DEFINITIONS "")
+
 if(UMF_BUILD_OS_MEMORY_PROVIDER)
     set(UMF_SOURCES_LINUX ${UMF_SOURCES_LINUX}
         provider/provider_os_memory.c
         provider/provider_os_memory_linux.c
         memory_targets/memory_target_numa.c
-        memspaces/memspace_numa.c)
+        memspaces/memspace_numa.c
+        memspaces/memspace_host_all.c)
 
     if(LINUX)
         if(PkgConfig_FOUND)
@@ -67,6 +72,8 @@ if(UMF_BUILD_OS_MEMORY_PROVIDER)
             find_package(LIBHWLOC REQUIRED hwloc)
         endif()
         set(UMF_LIBS ${UMF_LIBS} ${LIBHWLOC_LIBRARIES})
+        # Currently, OS provider is only built for Linux
+        set(UMF_COMPILE_DEFINITIONS ${UMF_COMPILE_DEFINITIONS} "UMF_BUILD_OS_MEMORY_PROVIDER=1")
     endif()
 endif()
 
@@ -85,15 +92,17 @@ if(UMF_BUILD_SHARED_LIBRARY)
                     LIBS ${UMF_LIBS}
                     LINUX_MAP_FILE ${CMAKE_CURRENT_SOURCE_DIR}/libumf.map
                     WINDOWS_DEF_FILE ${CMAKE_CURRENT_SOURCE_DIR}/libumf.def)
-    target_compile_definitions(umf PUBLIC UMF_SHARED_LIBRARY)
+    set(UMF_COMPILE_DEFINITIONS ${UMF_COMPILE_DEFINITIONS} "UMF_SHARED_LIBRARY")
     set_target_properties(umf PROPERTIES
         RUNTIME_OUTPUT_DIRECTORY ${CMAKE_UMF_OUTPUT_DIRECTORY})
 else()
     add_umf_library(NAME umf
-                    TYPE STATIC 
-                    SRCS ${UMF_SOURCES} 
+                    TYPE STATIC
+                    SRCS ${UMF_SOURCES}
                     LIBS ${UMF_LIBS})
 endif()
+
+target_compile_definitions(umf PUBLIC ${UMF_COMPILE_DEFINITIONS})
 
 if (UMF_ENABLE_POOL_TRACKING)
     target_sources(umf PRIVATE memory_pool_tracking.c)
@@ -103,10 +112,13 @@ endif()
 
 add_library(${PROJECT_NAME}::umf ALIAS umf)
 
-target_include_directories(umf PUBLIC 
+target_include_directories(umf PUBLIC
     $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/critnib>
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/provider>
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/memspaces>
+    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/memory_targets>
     $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
 )
 
--- a/src/libumf.def
+++ b/src/libumf.def
@@ -16,6 +16,7 @@ EXPORTS
     umfMemoryProviderAllocationMerge
     umfMemoryProviderAllocationSplit
     umfMemoryProviderCreate
+    umfMemoryProviderCreateFromMemspace
     umfMemoryProviderDestroy
     umfMemoryProviderFree
     umfMemoryProviderGetLastNativeError
@@ -24,10 +25,12 @@ EXPORTS
     umfMemoryProviderGetRecommendedPageSize
     umfMemoryProviderPurgeForce
     umfMemoryProviderPurgeLazy
+    umfMemspaceDestroy
     umfPoolAlignedMalloc
     umfPoolByPtr
     umfPoolCalloc
     umfPoolCreate
+    umfPoolCreateFromMemspace
     umfPoolDestroy
     umfPoolFree
     umfPoolGetLastAllocationError
@@ -35,7 +38,4 @@ EXPORTS
     umfPoolMalloc
     umfPoolMallocUsableSize
     umfPoolRealloc
-    umfPoolCreateFromMemspace
-    umfMemoryProviderCreateFromMemspace
-    umfMemspaceDestroy
     umfProxyPoolOps
--- a/src/libumf_linux.c
+++ b/src/libumf_linux.c
@@ -10,6 +10,7 @@
 #include <stddef.h>
 
 #include "base_alloc_global.h"
+#include "memspace_host_all_internal.h"
 #include "provider_tracking.h"
 
 umf_memory_tracker_handle_t TRACKER = NULL;
@@ -23,6 +24,10 @@ void __attribute__((destructor)) umfDestroy(void) {
     // make sure TRACKER is not used after being destroyed
     TRACKER = NULL;
     umfMemoryTrackerDestroy(t);
+
+#if defined(UMF_BUILD_OS_MEMORY_PROVIDER)
+    umfMemspaceHostAllDestroy();
+#endif
 }
 
 void libumfInit(void) {
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -50,7 +50,7 @@ function(add_umf_test)
     set_tests_properties(${TEST_NAME} PROPERTIES LABELS "umf")
 
     if(WIN32)
-        set_property(TEST ${TEST_NAME} PROPERTY ENVIRONMENT_MODIFICATION 
+        set_property(TEST ${TEST_NAME} PROPERTY ENVIRONMENT_MODIFICATION
             "PATH=path_list_append:../bin/$<CONFIG>")
     endif()
 
@@ -61,7 +61,7 @@ endfunction()
 
 add_subdirectory(common)
 
-add_umf_test(NAME base 
+add_umf_test(NAME base
              SRCS base.cpp)
 add_umf_test(NAME memoryPool
              SRCS memoryPoolAPI.cpp malloc_compliance_tests.cpp)
@@ -129,6 +129,9 @@ if(UMF_BUILD_OS_MEMORY_PROVIDER AND LINUX) # OS-specific functions are implement
     add_umf_test(NAME provider_os_memory_config
                  SRCS provider_os_memory_config.cpp
                  LIBS umf_utils ${LIBNUMA_LIBRARIES})
+    add_umf_test(NAME memspace_host_all
+                 SRCS memspace_host_all.cpp
+                 LIBS ${LIBNUMA_LIBRARIES})
 endif()
 
 if(UMF_BUILD_SHARED_LIBRARY)
--- a/src/libumf.map
+++ b/src/libumf.map
@@ -10,6 +10,7 @@ UMF_1.0 {
         umfMemoryProviderAllocationMerge;
         umfMemoryProviderAllocationSplit;
         umfMemoryProviderCreate;
+        umfMemoryProviderCreateFromMemspace;
         umfMemoryProviderDestroy;
         umfMemoryProviderFree;
         umfMemoryProviderGetLastNativeError;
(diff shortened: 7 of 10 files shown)


src/memspaces/memspace_numa.h
```diff
@@ -27,6 +27,8 @@ extern "C" {
 umf_result_t umfMemspaceCreateFromNumaArray(size_t *nodeIds, size_t numIds,
                                             umf_memspace_handle_t *hMemspace);
 
+extern umf_memspace_handle_t UMF_MEMSPACE_HOST;
```
igchor, line 30: Hm, I think for memspaces we'll need to follow the same approach as with ops - i.e. create a function that returns memspace handle instead of defining a global variable. See: https://github.com/oneapi-src/unified-memory-framework/pull/201

@vinser52 @bratpiorka @pbalcer are we OK with having predefined memspaces as functions? I'm just wondering if we should have a function per memspace or just a single function that accepts some enum (.e.g. `umfMemspaceGet(UMF_MEMSPACE_HOST)`)?

src/memspaces/memspace_host.c
```diff
@@ -0,0 +1,66 @@
...
+
+umf_result_t umfMemspaceHostCreate(umf_memspace_handle_t *hMemspace) {
+    if (!hMemspace) {
+        return UMF_RESULT_ERROR_INVALID_ARGUMENT;
+    }
+
+    umf_result_t umf_ret = UMF_RESULT_SUCCESS;
+
+    hwloc_topology_t topology;
+    if (hwloc_topology_init(&topology)) {
```
igchor, line 27: @pbalcer I know we wanted to strip libumf from malloc symbols (hence using base_alloc for internal allocations) but we have not control over what hwloc calls internally (I think they do call malloc). Will linking with hwloc be a problem for the 'proxy library'?

src/libumf_macosx.c
```diff
@@ -0,0 +1,29 @@
...
+ *
+ */
+
+#include "memspace_host_internal.h"
+#include "memspace_internal.h"
+#include "provider_tracking.h"
+
+#include <stddef.h>
+
+umf_memory_tracker_handle_t TRACKER = NULL;
```
igchor, line 16: If we really need a file for macosx I'm starting to think that it would be better to have all this initialization logic in a single file with ifdefs vs having 3 separate files for linux, windows and macosx.

kswiecicki: It was either a new file or checking for some compile option `MACOSX`, because we don't build the OS provider that the HOST memspace depends on.

src/memspaces/memspace_host.c
bratpiorka: I checked that they call malloc (https://github.com/open-mpi/hwloc/blob/eeeda4ad2d0511a2334a02141dab6138b1e0ce47/hwloc/topology.c). If we want to use HOST memspace in the proxy lib that exports malloc symbol, then we'll go into a recursion here

src/memspaces/memspace_host.c
```diff
@@ -0,0 +1,66 @@
...
+        umf_ret = UMF_RESULT_ERROR_UNKNOWN;
+        goto err_topology_destroy;
+    }
+
+    // Shouldn't return -1, as 'HWLOC_OBJ_NUMANODE' deosn't appear to be an
+    // object that can be present on multiple levels.
+    // Source: https://www.open-mpi.org/projects/hwloc/doc/hwloc-v2.10.0-letter.pdf
+    size_t nNodes =
+        (size_t)hwloc_get_nbobjs_by_type(topology, HWLOC_OBJ_NUMANODE);
+    size_t *nodeIds = malloc(nNodes * sizeof(size_t));
```
bratpiorka, line 42: please use ba_alloc here - check how this is done for pools

bratpiorka (changes requested):

src/libumf_linux.c
```diff
@@ -7,17 +7,30 @@
...
  */
 
+#include "memspace_host_internal.h"
+#include "memspace_internal.h"
 #include "provider_tracking.h"
 
 #include <stddef.h>
 
 umf_memory_tracker_handle_t TRACKER = NULL;
+umf_memspace_handle_t UMF_MEMSPACE_HOST = NULL;
```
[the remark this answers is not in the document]
kswiecicki: I'm thinking about how to free a memspace initialized that way. Would it be appropriate to free it in library destructor?

src/provider/provider_tracking_linux.c
```diff
@@ -0,0 +1,22 @@
...
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ *
+ */
+
+#include "provider_tracking.h"
+
+#include <stddef.h>
+
+extern umf_memory_tracker_handle_t TRACKER;
+
```
igchor, line 15: Why is this file added here?

src/libumf_linux.c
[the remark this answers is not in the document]
bratpiorka: Yes - these pre-defined namespaces are global, singular, and immutable. They can't be destroyed by the user.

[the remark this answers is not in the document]
kswiecicki: Done.

src/memspaces/memspace_host.c
```diff
@@ -0,0 +1,101 @@
...
+
+static bool UMF_MEMSPACE_HOST_INITIALIZED = false;
+static umf_memspace_handle_t UMF_MEMSPACE_HOST = NULL;
+
+umf_result_t umfMemspaceHostGet(umf_memspace_handle_t *hMemspace) {
+    if (!hMemspace) {
+        return UMF_RESULT_ERROR_INVALID_ARGUMENT;
+    }
+
+    if (!UMF_MEMSPACE_HOST_INITIALIZED) {
```
bratpiorka, line 83: can't you just check if UMF_MEMSPACE_HOST == NULL ?

bratpiorka:
btw - I added a separate JIRA task for extended test for HOST memspace (UMFW-246)

src/memspaces/memspace_host.c
kswiecicki: Done.

test/CMakeLists.txt
```diff
@@ -50,7 +50,7 @@ function(add_umf_test)
     set_tests_properties(${TEST_NAME} PROPERTIES LABELS "umf")
 
     if(WIN32)
-        set_property(TEST ${TEST_NAME} PROPERTY ENVIRONMENT_MODIFICATION 
+        set_property(TEST ${TEST_NAME} PROPERTY ENVIRONMENT_MODIFICATION
```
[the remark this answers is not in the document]
kswiecicki: I'm not sure if its capable of working in `CMakeLists.txt` files.

src/memspaces/memspace_host.c
```diff
@@ -0,0 +1,98 @@
...
+        // TODO: What would be an approrpiate err?
+        return UMF_RESULT_ERROR_UNKNOWN;
+    }
+
+    if (hwloc_topology_load(topology)) {
+        umf_ret = UMF_RESULT_ERROR_UNKNOWN;
+        goto err_topology_destroy;
+    }
+
+    // Shouldn't return -1, as 'HWLOC_OBJ_NUMANODE' deosn't appear to be an
```
igchor, line 37: pehaps add and assert for that?

src/memspaces/memspace_host.c
```diff
@@ -0,0 +1,98 @@
...
+}
+
+static umf_memspace_handle_t UMF_MEMSPACE_HOST = NULL;
+
+umf_result_t umfMemspaceHostGet(umf_memspace_handle_t *hMemspace) {
+    if (!hMemspace) {
+        return UMF_RESULT_ERROR_INVALID_ARGUMENT;
+    }
+
+    if (!UMF_MEMSPACE_HOST) {
```
igchor, line 82: This should be under pthread_once or in library ctor - we want to allow calling umfMemspaceHostGet simultaneously from multiple threads.

src/memspaces/memspace_host.c
kswiecicki: Done, but I've set the `UMF_RESULT_ERROR_UNKNOWN` return value instead of assert.

src/memspaces/memspace_host.c
kswiecicki: Done.

igchor (approved):

src/CMakeLists.txt
```diff
@@ -47,14 +47,20 @@ set(UMF_SOURCES_MACOSX
...
         provider/provider_os_memory.c
         provider/provider_os_memory_linux.c
         memory_targets/memory_target_numa.c
-        memspaces/memspace_numa.c)
+        memspaces/memspace_numa.c
+        memspaces/memspace_host.c)
     if(LINUX)
         set(UMF_LIBS ${UMF_LIBS} hwloc)
+        # Currently, OS provider is only built for Linux
+        set(UMF_COMPILE_DEFINITIONS ${UMF_COMPILE_DEFINITIONS} "UMF_BUILD_OS_MEMORY_PROVIDER")
```
bratpiorka, line 63: in "benchmark\CMakeLists.txt" there is a similar definition:

if (UMF_BUILD_OS_MEMORY_PROVIDER)
	target_compile_definitions(ubench PRIVATE UMF_BUILD_OS_MEMORY_PROVIDER=1)
endif()

could you add this definition the same way? In the future I think we should make this defines PUBLIC and cleanup cmake files but in the separate PR. You could even add a TODO for that

src/libumf.def
```diff
@@ -16,6 +16,7 @@ EXPORTS
     umfMemoryProviderAllocationMerge
     umfMemoryProviderAllocationSplit
     umfMemoryProviderCreate
+    umfMemoryProviderCreateFromMemspace
```
bratpiorka, line 19: this is one of the reasons why I would love this file to be generated :)

bratpiorka (changes requested):

include/umf/memspace.h
```diff
@@ -44,6 +44,12 @@ umfMemoryProviderCreateFromMemspace(umf_memspace_handle_t hMemspace,
                                     umf_memspace_policy_handle_t hPolicy,
                                     umf_memory_provider_handle_t *hProvider);
 
+///
+/// \brief Retrieves predefined host memspace.
+/// \return UMF_RESULT_SUCCESS on success or appropriate error code on failure.
+///
+umf_result_t umfMemspaceHostGet(umf_memspace_handle_t *hMemspace);
```
vinser52, line 51: Are we using getter function instead of predefined global memspace variable for the same reason (to be able to build UMF as a static lib) we replaced global OPS structures for pools/providers?

include/umf/memspace.h
```diff
@@ -44,6 +44,12 @@ umfMemoryProviderCreateFromMemspace(umf_memspace_handle_t hMemspace,
                                     umf_memspace_policy_handle_t hPolicy,
                                     umf_memory_provider_handle_t *hProvider);
 
+///
+/// \brief Retrieves predefined host memspace.
+/// \return UMF_RESULT_SUCCESS on success or appropriate error code on failure.
```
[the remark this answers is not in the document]
kswiecicki: Done.

src/CMakeLists.txt
kswiecicki: Done.

include/umf/memspace.h
```diff
@@ -44,6 +44,13 @@ umfMemoryProviderCreateFromMemspace(umf_memspace_handle_t hMemspace,
                                     umf_memspace_policy_handle_t hPolicy,
                                     umf_memory_provider_handle_t *hProvider);
 
+///
+/// \brief Retrieves predefined host memspace.
```
vinser52, line 48: What does "host" memspace mean? Is it something similar to the default kind in Memkind? Anyway, I think we need a more descriptive name.

kswiecicki: It's a memspace backed by numa nodes. Perhaps `UMF_MEMSPACE_NUMA` would be better?

include/umf/memspace.h
kswiecicki: Yeah, @igchor mentioned that it would be for the best if host memspace followed the same approach as with ops.

src/CMakeLists.txt
```diff
@@ -47,14 +47,21 @@ set(UMF_SOURCES_MACOSX
     libumf_linux.c
 )
 
+# Compile definitions for UMF library
+# TODO: Cleanup the compile definitions across all the Cmake files
+set(UMF_COMPILE_DEFINITIONS "")
```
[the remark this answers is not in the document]
kswiecicki: I can't use `target_compile_definitions()` on a target that was not yet created.

kswiecicki:
I've seen the changes in https://github.com/oneapi-src/unified-memory-framework/pull/229/ and adopted the same approach to host memspace.

include/umf/memspace.h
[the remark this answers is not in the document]
kswiecicki: Actually, I've changed the signature of `umfMemspaceHostGet` to no longer accept a param.

include/umf/memspace.h
bratpiorka: I think we should call it the same as in memkind - "defaut". In memkind, all NUMA nodes available for allocation (so also, for example, PMEM nodes set to KMEM_DAX mode) created a "default" namespace. In addition, NUMA nodes that contain the CPU (= DRAM memory) created the "regular" kind.

bratpiorka (approved):

test/memspace_host.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+#define SIZE_4K (4096)
+
+struct predefinedMemspaceTest : test {
+
+    void SetUp() override {
+        ::test::SetUp();
+
+        hMemspace = umfMemspaceHostGet();
+        ASSERT_NE(hMemspace, nullptr);
```
lukaszstolarczuk, line 19: perhaps include `test_helpers.h` and use our `UT_ASSERT*` macros

src/memspaces/memspace_host.c
lukaszstolarczuk, line 1: could you please provide at lest a little description in the commit message (to the commit "Add predefined HOST memspace")

test/memspace_host.cpp
kswiecicki: Done.

src/memspaces/memspace_host.c
kswiecicki: I've added a more descriptive commit message.

lukaszstolarczuk: thx!

lukaszstolarczuk (commented):
👍

include/umf/memspace.h
kswiecicki: I've changed the memspace name to `default`.

README.md
```diff
@@ -161,6 +161,18 @@ The `UMF_BUILD_LIBUMF_POOL_SCALABLE` option has to be turned `ON` to build this
 2) Required packages:
    - libtbb-dev (libraries: libtbbmalloc.so.2)
 
+### Memspaces (Linux-only)
+
+TODO: Add general information about memspaces.
+
+#### Default memspace
+
+Memspace backed by NUMA node discovery. Can be retrieved using umfMemspaceDefaultGet.
```
vinser52, line 170: Can we describe more precisely what “dafault” means (e.g. all available NUMA nodes)? Or we intentionaly do not specify this for some abstraction purposes.

test/memspace_default.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+    umf_memory_provider_handle_t provider = nullptr;
+};
+
+TEST_F(test, memspaceGet) {
+    umf_memspace_handle_t hMemspace = umfMemspaceDefaultGet();
+    UT_ASSERTne(hMemspace, nullptr);
+}
+
+TEST_F(predefinedMemspaceTest, providerFromMemspace) {
```
vinser52, line 51: This test looks like a duplication. The same flow is done and checked in the `SetUp` function.

README.md
kswiecicki: I've changed the brief about `default` memspace. The info that it's backed by all available NUMA nodes is now transparent.

include/umf/memspace.h
```diff
@@ -44,6 +44,12 @@ umfMemoryProviderCreateFromMemspace(umf_memspace_handle_t hMemspace,
                                     umf_memspace_policy_handle_t hPolicy,
                                     umf_memory_provider_handle_t *hProvider);
 
+///
+/// \brief Retrieves predefined default memspace.
+/// \return default memspace handle on success or NULL on failure.
+///
+umf_memspace_handle_t umfMemspaceDefaultGet(void);
```
vinser52, line 51: An alternative API could be a single entry function which accept some enum to specify which predefined memspace to return. For example:
```
enum predefined_memspaces {
    DEFAULT,
    ...
};

umf_memspace_handle_t umfPredefinedMemspaceGet(enum predefined_memspaces) {
    ...
}
```

@igchor, @bratpiorka, @pbalcer what do you think?

src/memspaces/memspace_default.c
```diff
@@ -0,0 +1,91 @@
...
+    umfMemspaceDestroy(UMF_MEMSPACE_DEFAULT);
+    UMF_MEMSPACE_DEFAULT = NULL;
+}
+
+static void umfMemspaceDefaultInit(void) {
+    umf_result_t ret = umfMemspaceDefaultCreate(&UMF_MEMSPACE_DEFAULT);
+    assert(ret == UMF_RESULT_SUCCESS);
+    (void)ret;
+
+    atexit(umfMemspaceDefaultDestroy);
```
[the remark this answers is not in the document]
igchor: The problem is that there's no library destructor on windows when we build as a static library. I think we can do the same as for base_alloc: https://github.com/oneapi-src/unified-memory-framework/blob/main/src/base_alloc/base_alloc_global.c#L63 We use atexit when building as static lib on windows and call umf_ba_destroy_global in library dtor otherwise.

include/umf/memspace.h
igchor: Yes, this is also possible. I don't really have a preference here.

bratpiorka: Currently, we use umfXXXPool/ProviderOps() for pools and providers. We should be consistent with that or change this everywhere. But please note that the enum is more problematic when you want to build this list at cmake configuration (enable/disable some compontents).

vinser52: > Currently, we use umfXXXPool/ProviderOps() for pools and providers. We should be consistent with that or change this everywhere.

I do not think the same rules apply everywhere.

> But please note that the enum is more problematic when you want to build this list at cmake configuration (enable/disable some components).

Ok, it makes sense if we want to configure the available predefined memspaces at build-time.

src/memspaces/memspace_default.c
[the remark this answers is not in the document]
vinser52: In general, what Igor suggested makes sense if we want to support dynamic and static versions of UMF. 
But eventually building UMF core as a static library does not make much sense - tracking would not work (as expected).

So I agree with what @igchor suggested.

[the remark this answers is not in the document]
kswiecicki: Done. I've followed the example that @igchor gave. I've removed the `atexit` altogether since DEFAULT memspace depends on OS provider that is unsupported on Windows for now.

PatKamin (approved):

ldorau (changes requested):
Please rebase on top of the current main branch where the proxy library is present.

ldorau (approved):

PR #206, opened 2024-02-05
Do not fail linear allocator with OOM

ldorau:
The linear base allocator fails with OOM now,
when it runs out of memory requested in `umf_ba_linear_create()`.
The linear base allocator is used to allocate structures
which size depends on number of nodes for example,
so it is not known before calling `umf_ba_linear_create()`.

This patch adds a feature that the pool of linear allocator can "grow"
(a new memory region is allocated from OS in fact)
when more space is needed.

### DIFF
--- a/src/base_alloc/base_alloc_linear.c
+++ b/src/base_alloc/base_alloc_linear.c
@@ -19,24 +19,59 @@
 // alignment of the linear base allocator
 #define MEMORY_ALIGNMENT (sizeof(uintptr_t))
 
-// metadata of the linear base allocator
-typedef struct {
-    size_t pool_size;
+typedef struct umf_ba_next_linear_pool_t umf_ba_next_linear_pool_t;
+
+// metadata is set and used only in the main (the first) pool
+typedef struct umf_ba_main_linear_pool_meta_t {
+    size_t pool_size; // size of each pool (argument of each ba_os_alloc() call)
     os_mutex_t lock;
     char *data_ptr;
     size_t size_left;
+#ifndef NDEBUG
+    size_t n_pools;
+#endif /* NDEBUG */
 } umf_ba_main_linear_pool_meta_t;
 
-// pool of the linear base allocator
+// the main pool of the linear base allocator (there is only one such pool)
 struct umf_ba_linear_pool {
+    // address of the beginning of the next pool (a list of allocated pools
+    // to be freed in umf_ba_linear_destroy())
+    umf_ba_next_linear_pool_t *next_pool;
+
+    // metadata is set and used only in the main (the first) pool
     umf_ba_main_linear_pool_meta_t metadata;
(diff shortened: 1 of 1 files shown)


src/base_alloc/base_alloc_linear.c
```diff
@@ -69,26 +105,59 @@ umf_ba_linear_pool_t *umf_ba_linear_create(size_t pool_size) {
...
     util_mutex_lock(&pool->metadata.lock);
     if (pool->metadata.size_left < aligned_size) {
-        fprintf(stderr,
-                "error: umf_ba_linear_alloc() failed (requested size: %zu > "
-                "space left: %zu)\n",
-                aligned_size, pool->metadata.size_left);
-        util_mutex_unlock(&pool->metadata.lock);
-        return NULL; // out of memory
+        // assert in Debug build (OOM)
+        assert(pool->metadata.size_left >= aligned_size);
```
igchor, line 111: Why are we failing in Debug and succeeding in Release? Why can't we have a consistent behavior?

src/base_alloc/base_alloc_linear.c
```diff
@@ -69,26 +105,59 @@ umf_ba_linear_pool_t *umf_ba_linear_create(size_t pool_size) {
...
+        align_ptr_size(&data_ptr, &size_left, MEMORY_ALIGNMENT);
+
+        pool->metadata.data_ptr = data_ptr;
+        pool->metadata.size_left = size_left;
+
+        // add the new pool to the list of pools
+        new_pool->next_pool = pool->next_pool;
+        pool->next_pool = new_pool;
+#ifndef NDEBUG
+        pool->metadata.n_pools++;
```
igchor, line 133: Will this ever happen? We are failing in Debug on line 111 if there's not enough memory in the pool, right?

src/base_alloc/base_alloc_linear.c
```diff
@@ -19,24 +19,56 @@
...
     size_t size_left;
+#ifndef NDEBUG
+    size_t n_pools;
+#endif /* NDEBUG */
 } umf_ba_main_linear_pool_meta_t;
 
-// pool of the linear base allocator
+// the main pool of the linear base allocator (there is only one such pool)
 struct umf_ba_linear_pool {
+    // address of the beginning of the next pool (a list of allocated pools to be freed in umf_ba_linear_destroy())
```
bratpiorka, line 37: please break long comments manually - this is disabled in clang-format

bratpiorka (approved):

src/base_alloc/base_alloc_linear.c
ldorau: Done

ldorau:
> can you please explain a little more in the commit msg...?

@lukaszstolarczuk Done

PatKamin (approved):

PR #220, opened 2024-02-07
Don't use CodeQL in reusable workflow

lukaszstolarczuk:
Due to lower score on Scorecard we'll be running this separately from "PR/push" workflow. For some reason permissions weren't properly set or recognized (by Scorecard). If Scorecard changes its behavior we can go back to use 'workflow_call' trigger.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,6 +1,10 @@
 name: CodeQL
 
-on: workflow_call
+# Due to lower score on Scorecard we're running this separately from
+# "PR/push" workflow. For some reason permissions weren't properly set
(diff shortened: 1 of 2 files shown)


PatKamin (approved):

bratpiorka (approved):

bratpiorka:
so this is all about setting securityEvents permissions to write?

wlemkows:
> so this is all about setting securityEvents permissions to write?

Yes, this change is necessary to achieve the required scorecard result in Q1. In the case of using reusable workflows scorecard did not allow setting "write" at the job level. It is hard for me to say whether it is a bug or a feature :) We decided to make the simplest and fastest change that will improve our score.

PR #221, opened 2024-02-07
Fix: make it possible to allocate more than initial pool size

ldorau:
Make it possible to allocate more than the initial pool size of the linear allocator.
It fixes a serious bug leading to a segfault in the linear base allocator.

PR #218, opened 2024-02-06
Add extra debug check for base_alloc

igchor:
that verifies that ptr belongs to a specific pool. This will be especially useful after https://github.com/oneapi-src/unified-memory-framework/pull/203 is merged.

PR #222, opened 2024-02-07
Add umf_ba_linear_pool_contains_pointer

ldorau:
Add `umf_ba_linear_pool_contains_pointer()`.
It returns:
- 0 if `ptr` does not belong to the pool or
- size (> 0) of the memory region from `ptr` to the end of the pool if `ptr` belongs to the pool.

It will be useful to implement realloc() in the proxy library.

Requires:
- [ ] #221

### DIFF
--- a/src/base_alloc/base_alloc_linear.c
+++ b/src/base_alloc/base_alloc_linear.c
@@ -23,7 +23,7 @@ typedef struct umf_ba_next_linear_pool_t umf_ba_next_linear_pool_t;
 
 // metadata is set and used only in the main (the first) pool
 typedef struct umf_ba_main_linear_pool_meta_t {
-    size_t pool_size; // size of each pool (argument of each ba_os_alloc() call)
+    size_t pool_size; // size of this pool (argument of ba_os_alloc() call)
     os_mutex_t lock;
     char *data_ptr;
     size_t size_left;
@@ -52,6 +52,8 @@ struct umf_ba_next_linear_pool_t {
     // to be freed in umf_ba_linear_destroy())
     umf_ba_next_linear_pool_t *next_pool;
 
+    size_t pool_size; // size of this pool (argument of ba_os_alloc() call)
+
     // data area of all pools except of the main (the first one) starts here
     char data[];
 };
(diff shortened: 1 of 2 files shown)


ldorau:
BTW: the code was copied from #218 by @igchor

ldorau:
> Looks good. But can you explain why this will be needed for proxy pool? I still don't understand why we can't just detect recursion in malloc/free (globally, not per pointer) and then call appropriate function (base_alloc or umf*)

If `free()` is called from "outside world" (not from inside of `malloc()`), it does not know if the pointer it gets comes from the base allocator or from the pool manager, so it has to check that.

igchor:
> > Looks good. But can you explain why this will be needed for proxy pool? I still don't understand why we can't just detect recursion in malloc/free (globally, not per pointer) and then call appropriate function (base_alloc or umf*)
> 
> If `free()` is called from "outside world" (not from inside of `malloc()`), it does not know if the pointer it gets comes from the base allocator or from the pool manager, so it has to check that.

How can a pointer from the 'outside world' come from the base allocator?

ldorau:
> > > Looks good. But can you explain why this will be needed for proxy pool? I still don't understand why we can't just detect recursion in malloc/free (globally, not per pointer) and then call appropriate function (base_alloc or umf*)
> > 
> > 
> > If `free()` is called from "outside world" (not from inside of `malloc()`), it does not know if the pointer it gets comes from the base allocator or from the pool manager, so it has to check that.
> 
> How can a pointer from the 'outside world' come from the base allocator? If a user calls 'free', the pointer has to come from a 'malloc' which means it has to be part of the umf pool, right?

Hundreds of pointers can come from the base allocator:
1) An app starts with LD_PRELOAD=proxy_lib
2) libc calls malloc() ... (they come **from the base allocator**):
```
>>> malloc(72704)
>>> realloc(0x0, 416)
>>> malloc(20)
>>> malloc(24)
```
3) The constructor of proxy_lib starts ...
4) The constructor of proxy_lib calls `umfMemoryProviderCreate(umfOsMemoryProviderOps())` and `hwloc` calls hundreds of malloc()s (they come **from the base allocator**):
```
>>> malloc(896)
>>> malloc(32)
>>> malloc(32816)
>>> malloc(100)
>>> malloc(44)
...
```
5) The constructor of proxy_lib calls `umfPoolCreate(umfJemallocPoolOps())`
6) The UMF pool is created. All allocations starting from this point come from the UMF pool.
...
7) An app calls `free(ptr)` - does the ptr come from the base allocator or from the UMF pool ???

ldorau:
Closing for now - it will be reopened as a part of the pull request with the proxy library ...

PR #223, opened 2024-02-08
Add tests with multiple pools

igchor:
To verify that:
- tracking works across different type of pools
- we can succesfully link multiple pool libs to a single app

### DIFF
--- a/test/c_api/multi_pool.c
+++ b/test/c_api/multi_pool.c
@@ -1,0 +1,82 @@
+// Copyright (C) 2023-2024 Intel Corporation
+// Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include <stdlib.h>
+
+#include <umf/memory_pool.h>
+#include <umf/pools/pool_disjoint.h>
+#include <umf/pools/pool_jemalloc.h>
+#include <umf/pools/pool_proxy.h>
+#include <umf/providers/provider_os_memory.h>
+
+#include "test_helpers.h"
+
+umf_memory_pool_handle_t
+createDisjointPool(umf_memory_provider_handle_t provider) {
+    umf_memory_pool_handle_t pool = NULL;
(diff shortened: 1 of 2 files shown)


test/c_api/multi_pool.c
```diff
@@ -0,0 +1,82 @@
...
+    umf_memory_pool_handle_t pool = NULL;
+    umf_result_t ret =
+        umfPoolCreate(umfJemallocPoolOps(), provider, NULL, 0, &pool);
+    UT_ASSERTeq(ret, UMF_RESULT_SUCCESS);
+    return pool;
+}
+
+#define ALLOC_SIZE 64
+
+int main(void) {
```
bratpiorka, line 45: could you use gtest?

bratpiorka (changes requested):

lukaszstolarczuk (commented):
👍

test/c_api/multi_pool.c
igchor: No, it's a C-only test.

bratpiorka: OK so could you move this test to the common test folder? I think we should use gtest wherever this is possible. This would also make it easier to extend this test in the future. BTW - grep for "test_disjoint_pool_shared_limits()" - this is one of the reason why we should use gtest :)

igchor: Right, but I specifically wanted this test to be in C so that we can verify our API is C-compliant. We don't have any other C-only test for pools other than disjoint_pool (btw, what's wrong with 'test_disjoint_pool_shared_limits'?). If you want I can rewrite this test in C++ and use gtest but then we should plan to write extra C tests in the future.

bratpiorka: ok I understand. I think we could have this test for the C API but if we want to add more complex muli-pools scenarios we should use gtest

bratpiorka:
@igchor please request someone to be a 2nd approver - maybe ping @lukaszstolarczuk for an approval?

lukaszstolarczuk (approved):

PR #224, opened 2024-02-09
[RFC] Add umfPoolFreeSized

igchor:
a version of umfPoolFree that accepts allocation size as arg.

Rationale:
It's often possible to greatly optimize free implementation when size is known. To support such optimizations we should expose umfPoolFreeSized an the let each pool decide whether it needs size argument or not.

Very often size of the allocation is known: e.g. especially when allocating metadata or fixed-sized structs.

### DIFF
--- a/src/memory_pool.c
+++ b/src/memory_pool.c
@@ -58,9 +58,15 @@ size_t umfPoolMallocUsableSize(umf_memory_pool_handle_t hPool, void *ptr) {
     return hPool->ops.malloc_usable_size(hPool->pool_priv, ptr);
 }
 
+umf_result_t umfPoolFreeSized(umf_memory_pool_handle_t hPool, void *ptr,
+                              size_t size) {
+    UMF_CHECK((hPool != NULL), UMF_RESULT_ERROR_INVALID_ARGUMENT);
+    return hPool->ops.free(hPool->pool_priv, ptr, size);
+}
+
 umf_result_t umfPoolFree(umf_memory_pool_handle_t hPool, void *ptr) {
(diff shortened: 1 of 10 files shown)


src/memory_pool.c
```diff
@@ -58,9 +58,15 @@ size_t umfPoolMallocUsableSize(umf_memory_pool_handle_t hPool, void *ptr) {
...
+umf_result_t umfPoolFreeSized(umf_memory_pool_handle_t hPool, void *ptr,
+                              size_t size) {
+    UMF_CHECK((hPool != NULL), UMF_RESULT_ERROR_INVALID_ARGUMENT);
+    return hPool->ops.free(hPool->pool_priv, ptr, size);
+}
+
 umf_result_t umfPoolFree(umf_memory_pool_handle_t hPool, void *ptr) {
     UMF_CHECK((hPool != NULL), UMF_RESULT_ERROR_INVALID_ARGUMENT);
-    return hPool->ops.free(hPool->pool_priv, ptr);
+    return hPool->ops.free(hPool->pool_priv, ptr, 0);
```
bratpiorka, line 69: As this is non-default API I think this should not be required to have this function in ops. So I would not change the regular free signature, but add an extra function to the ops. Note that the optimizations you mentioned have to be implemented in pools, so it makes sense to the pool to expose this function only if using it makes sense. Also, as a fallback (if there is no such function in the ops) we could just call free(ptr) and ignore the size

bratpiorka (changes requested):

bratpiorka:
@igchor are you planning to add the optimizations you mentioned for DisjointPool or another pool to the UMF?

igchor:
Closing for now. Will reopen if any pool needs this

PR #226, opened 2024-02-12
Add proxy library (for Linux only now)

ldorau:
Ref: https://github.com/oneapi-src/unified-memory-framework/issues/240

### DIFF
--- a/.github/workflows/proxy_lib.yml
+++ b/.github/workflows/proxy_lib.yml
@@ -1,0 +1,62 @@
+name: Proxy library
+
+on: workflow_call
+
+permissions:
+  contents: read
+
+jobs:
+  ubuntu-build:
+    name: Build - Ubuntu
+
+    strategy:
+      matrix:
+        os: ['ubuntu-22.04']
+        build_type: [Release, Debug]
+        compiler: [{c: gcc, cxx: g++}]
+    runs-on: ${{matrix.os}}
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+
+      - name: Install apt packages
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y cmake libhwloc-dev libjemalloc-dev libtbb-dev
+
+      - name: Configure build
+        run: >
+          cmake
+          -B ${{github.workspace}}/build
+          -DCMAKE_BUILD_TYPE=${{matrix.build_type}}
+          -DCMAKE_C_COMPILER=${{matrix.compiler.c}}
+          -DCMAKE_CXX_COMPILER=${{matrix.compiler.cxx}}
+          -DUMF_BUILD_SHARED_LIBRARY=ON
+          -DUMF_BUILD_BENCHMARKS=ON
+          -DUMF_BUILD_TESTS=ON
+          -DUMF_FORMAT_CODE_STYLE=OFF
+          -DUMF_DEVELOPER_MODE=OFF
+          -DUMF_BUILD_LIBUMF_POOL_JEMALLOC=ON
+          -DUMF_BUILD_LIBUMF_POOL_DISJOINT=ON
+          -DUMF_BUILD_LIBUMF_POOL_SCALABLE=ON
+          -DUMF_ENABLE_POOL_TRACKING=OFF
+
+      - name: Build UMF
+        run: cmake --build ${{github.workspace}}/build -j $(nproc)
+
+      - name: Run "ctest --output-on-failure" with proxy library
+        working-directory: ${{github.workspace}}/build
+        run: LD_PRELOAD=./lib/libumf_proxy.so ctest --output-on-failure
+
+      - name: Run "./test/umf_test-memoryPool" with proxy library
+        working-directory: ${{github.workspace}}/build
+        run: LD_PRELOAD=./lib/libumf_proxy.so ./test/umf_test-memoryPool
+
+      - name: Run "/usr/bin/ls" with proxy library
+        working-directory: ${{github.workspace}}/build
+        run: LD_PRELOAD=./lib/libumf_proxy.so /usr/bin/ls
+
+      - name: Run "/usr/bin/date" with proxy library
+        working-directory: ${{github.workspace}}/build
+        run: LD_PRELOAD=./lib/libumf_proxy.so /usr/bin/date
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -4,6 +4,11 @@
 
 include(${UMF_CMAKE_SOURCE_DIR}/cmake/helpers.cmake)
 
+# set UMF_PROXY_LIB_BASED_ON_POOL to one of:
+# - SCALABLE
+# - JEMALLOC
+set(UMF_PROXY_LIB_BASED_ON_POOL SCALABLE CACHE STRING "A UMF pool the proxy library is based on (SCALABLE or JEMALLOC)" FORCE)
+
 add_subdirectory(utils)
 
 set(UMF_LIBS umf_utils)
@@ -103,3 +108,26 @@ install(TARGETS umf
 )
 
 add_subdirectory(pool)
+
+# TODO: enable proxy_lib on Windows
+if(LINUX)
+    if(UMF_PROXY_LIB_BASED_ON_POOL STREQUAL SCALABLE)
+        set(PROXY_LIB_USES_SCALABLE_POOL ON)
+        set(PROXY_LIBS umf scalable_pool)
+        if(UMF_BUILD_LIBUMF_POOL_SCALABLE)
+            add_subdirectory(proxy_lib)
+        else()
+            message(STATUS "Disabling the proxy library, because UMF_PROXY_LIB_BASED_ON_POOL==SCALABLE but UMF_BUILD_LIBUMF_POOL_SCALABLE is OFF")
+        endif()
+    elseif(UMF_PROXY_LIB_BASED_ON_POOL STREQUAL JEMALLOC)
+        set(PROXY_LIB_USES_JEMALLOC_POOL ON)
+        set(PROXY_LIBS umf jemalloc_pool)
+        if(UMF_BUILD_LIBUMF_POOL_JEMALLOC)
+            add_subdirectory(proxy_lib)
+        else()
+            message(STATUS "Disabling the proxy library, because UMF_PROXY_LIB_BASED_ON_POOL==JEMALLOC but UMF_BUILD_LIBUMF_POOL_JEMALLOC is OFF")
+        endif()
+    else()
+        message(FATAL_ERROR "Proxy library: pool manager not chosen or set to a non-supported one (see UMF_PROXY_LIB_BASED_ON_POOL)")
+    endif()
+endif()
--- a/src/base_alloc/base_alloc_linear.c
+++ b/src/base_alloc/base_alloc_linear.c
@@ -175,3 +175,33 @@ void umf_ba_linear_destroy(umf_ba_linear_pool_t *pool) {
     util_mutex_destroy_not_free(&pool->metadata.lock);
     ba_os_free(pool, pool->metadata.pool_size);
 }
+
+// umf_ba_linear_pool_contains_pointer() returns:
+// - 0 if ptr does not belong to the pool or
+// - size (> 0) of the memory region from ptr
+//   to the end of the pool if ptr belongs to the pool
+size_t umf_ba_linear_pool_contains_pointer(umf_ba_linear_pool_t *pool,
+                                           void *ptr) {
+    util_mutex_lock(&pool->metadata.lock);
+    char *cptr = (char *)ptr;
+    if (cptr >= pool->data &&
+        cptr < ((char *)(pool)) + pool->metadata.pool_size) {
+        size_t size = ((char *)(pool)) + pool->metadata.pool_size - cptr;
+        util_mutex_unlock(&pool->metadata.lock);
+        return size;
+    }
+
+    umf_ba_next_linear_pool_t *next_pool = pool->next_pool;
+    while (next_pool) {
+        if (cptr >= next_pool->data &&
+            cptr < ((char *)(next_pool)) + next_pool->pool_size) {
+            size_t size = ((char *)(next_pool)) + next_pool->pool_size - cptr;
+            util_mutex_unlock(&pool->metadata.lock);
+            return size;
+        }
+        next_pool = next_pool->next_pool;
+    }
+
+    util_mutex_unlock(&pool->metadata.lock);
+    return 0;
+}
--- a/src/base_alloc/base_alloc_linear.c
+++ b/src/base_alloc/base_alloc_linear.c
@@ -161,9 +161,17 @@ void *umf_ba_linear_alloc(umf_ba_linear_pool_t *pool, size_t size) {
 }
 
 void umf_ba_linear_destroy(umf_ba_linear_pool_t *pool) {
+    // Do not destroy if we are running in the proxy library,
+    // because it may need those resources till
+    // the very end of exiting the application.
+    if (is_running_in_proxy_lib()) {
+        return;
+    }
+
 #ifndef NDEBUG
     ba_debug_checks(pool);
 #endif /* NDEBUG */
+
     umf_ba_next_linear_pool_t *current_pool;
     umf_ba_next_linear_pool_t *next_pool = pool->next_pool;
     while (next_pool) {
--- a/src/proxy_lib/proxy_lib.c
+++ b/src/proxy_lib/proxy_lib.c
@@ -1,0 +1,261 @@
+/*
+ * Copyright (C) 2024 Intel Corporation
+ *
+ * Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+*/
+
+/*
+ * UMF proxy library - a library for intercepting user allocation requests
+ *
+ * It intercepts following APIs:
+ * - aligned_alloc()
+ * - calloc()
+ * - free()
+ * - malloc()
+ * - malloc_usable_size()
+ * - realloc()
+ */
+
+#if (defined PROXY_LIB_USES_JEMALLOC_POOL)
+#include <umf/pools/pool_jemalloc.h>
+#define umfPoolManagerOps umfJemallocPoolOps
+#elif (defined PROXY_LIB_USES_SCALABLE_POOL)
+#include <umf/pools/pool_scalable.h>
+#define umfPoolManagerOps umfScalablePoolOps
+#else
+#error Pool manager not defined
+#endif
+
+#include <assert.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include <umf/memory_pool.h>
+#include <umf/memory_provider.h>
+#include <umf/providers/provider_os_memory.h>
+
+#include "base_alloc_linear.h"
+#include "proxy_lib.h"
+#include "utils_common.h"
+#include "utils_concurrency.h"
+
+/*
+ * The UMF proxy library uses two memory allocators:
+ * 1) the "LEAK" internal linear base allocator based on the anonymous mapped
+ *    memory that will NOT be destroyed (with API ba_leak_*()).
+ * 2) the main one - UMF pool allocator.
+ *
+ * Ad 1)
+ * The "LEAK" internal linear base allocator is used from the very beginning
+ * to the creation of a UMF pool in the constructor of the proxy library.
+ * It is used to allocate memory for OS specific data used during loading and unloading
+ * applications (for example _dl_init() and _dl_fini() on Linux storing data of all
+ * constructors and destructors that have to be called) and also memory needed
+ * by umfMemoryProviderCreate() and umfPoolCreate().
+ * That memory will be leaked on purpose (OS will have to free it during destroying
+ * the process), because we cannot free the memory containing data of destructors
+ * that have to be called at the end (for example memory allocated by _dl_init()
+ * and used internally by _dl_fini() on Linux).
+ * The "LEAK" internal linear base allocator uses about 900 kB on Linux.
+ *
+ * Ad 2)
+ * The UMF pool allocator (the main one) is used from the creation to the destruction
+ * of a UMF pool to allocate memory needed by an application. It should be freed
+ * by an application.
+ */
+
+static UTIL_ONCE_FLAG Base_alloc_leak_initialized = UTIL_ONCE_FLAG_INIT;
+static umf_ba_linear_pool_t *Base_alloc_leak = NULL;
+static umf_memory_provider_handle_t OS_memory_provider = NULL;
+static umf_memory_pool_handle_t Proxy_pool = NULL;
+
+// it protects us from recursion in umfPool*()
+static __TLS int was_called_from_umfPool = 0;
+
+/*****************************************************************************/
+/*** The constructor and destructor of the proxy library *********************/
+/*****************************************************************************/
+
+void proxy_lib_create_common(void) {
+    umf_os_memory_provider_params_t os_params =
+        umfOsMemoryProviderParamsDefault();
+    enum umf_result_t umf_result;
+
+    umf_result = umfMemoryProviderCreate(umfOsMemoryProviderOps(), &os_params,
+                                         &OS_memory_provider);
+    if (umf_result != UMF_RESULT_SUCCESS) {
+        fprintf(stderr, "error: creating OS memory provider failed\n");
+        exit(-1);
+    }
+
+    umf_result = umfPoolCreate(umfPoolManagerOps(), OS_memory_provider, NULL, 0,
+                               &Proxy_pool);
+    if (umf_result != UMF_RESULT_SUCCESS) {
+        fprintf(stderr, "error: creating UMF pool manager failed\n");
+        exit(-1);
+    }
+    // The UMF pool has just been created (Proxy_pool != NULL). Stop using
+    // the linear allocator and start using the UMF pool allocator from now on.
+}
+
+void proxy_lib_destroy_common(void) {
+    // We cannot destroy 'Base_alloc_leak' nor 'Proxy_pool' nor 'OS_memory_provider',
+    // because it could lead to use-after-free in the program's unloader
+    // (for example _dl_fini() on Linux).
+}
+
+/*****************************************************************************/
+/*** Generic version of realloc() of linear base allocator *******************/
+/*****************************************************************************/
+
+static inline void *ba_generic_realloc(umf_ba_linear_pool_t *pool, void *ptr,
+                                       size_t new_size, size_t max_size) {
+    assert(ptr);      // it should be verified in the main realloc()
+    assert(new_size); // it should be verified in the main realloc()
+    assert(max_size); // max_size should be set in the main realloc()
+
+    void *new_ptr = umf_ba_linear_alloc(pool, new_size);
+    if (!new_ptr) {
+        return NULL;
+    }
+
+    if (new_size > max_size) {
+        new_size = max_size;
+    }
+
+    memcpy(new_ptr, ptr, new_size);
+
+    return new_ptr;
+}
+
+/*****************************************************************************/
+/*** The "LEAK" linear base allocator functions ******************************/
+/*****************************************************************************/
+
+static void ba_leak_create(void) { Base_alloc_leak = umf_ba_linear_create(0); }
+
+// it does not implement destroy(), because it will not free memory at all
+
+static inline void *ba_leak_malloc(size_t size) {
+    util_init_once(&Base_alloc_leak_initialized, ba_leak_create);
+    return umf_ba_linear_alloc(Base_alloc_leak, size);
+}
+
+static inline void *ba_leak_calloc(size_t nmemb, size_t size) {
+    util_init_once(&Base_alloc_leak_initialized, ba_leak_create);
+    // umf_ba_linear_alloc() returns zeroed memory
+    return umf_ba_linear_alloc(Base_alloc_leak, nmemb * size);
+}
+
+static inline void *ba_leak_realloc(void *ptr, size_t size, size_t max_size) {
+    util_init_once(&Base_alloc_leak_initialized, ba_leak_create);
+    return ba_generic_realloc(Base_alloc_leak, ptr, size, max_size);
+}
+
+static inline void *ba_leak_aligned_alloc(size_t alignment, size_t size) {
+    util_init_once(&Base_alloc_leak_initialized, ba_leak_create);
+    void *ptr = umf_ba_linear_alloc(Base_alloc_leak, size + alignment);
+    return (void *)ALIGN_UP((uintptr_t)ptr, alignment);
+}
+
+static inline size_t ba_leak_pool_contains_pointer(void *ptr) {
+    return umf_ba_linear_pool_contains_pointer(Base_alloc_leak, ptr);
+}
+
+/*****************************************************************************/
+/*** The UMF pool allocator functions (the public API) ***********************/
+/*****************************************************************************/
+
+void *malloc(size_t size) {
+    if (!was_called_from_umfPool && Proxy_pool) {
+        was_called_from_umfPool = 1;
+        void *ptr = umfPoolMalloc(Proxy_pool, size);
+        was_called_from_umfPool = 0;
+        return ptr;
+    }
+
+    return ba_leak_malloc(size);
+}
+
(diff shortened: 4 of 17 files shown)


src/proxy_lib/proxy_lib_linux.c
```diff
@@ -0,0 +1,41 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+
+#include "proxy_lib.h"
+
+void __attribute__((constructor(101))) proxy_lib_create(void) {
+    proxy_lib_create_common();
+}
+
+void __attribute__((destructor(101))) proxy_lib_destroy(void) {
```
igchor, line 22: Is the destructor priority needed after all? If yes, please add some comment.

src/proxy_lib/proxy_lib.h
```diff
@@ -0,0 +1,43 @@
...
+#define UMF_PROXY_LIB_H 1
+
+#include <stddef.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#ifdef _WIN32
+#define WIN32_dllexport __declspec(dllexport)
```
igchor, line 18: This is not needed since we use .def file

src/proxy_lib/proxy_lib.c
```diff
@@ -0,0 +1,175 @@
...
+    umf_memory_pool_handle_t pool = Proxy_pool;
+    Proxy_pool = NULL;
+    umfPoolDestroy(pool);
+    umfMemoryProviderDestroy(OS_memory_provider);
+}
+
+/******************************************************************************/
+
+static inline void *system_malloc(size_t size) {
+    if (!System_malloc_func) {
```
igchor, line 65: I think there's a race possible here. Shouldn;t proxy_lib_system_alloc_init be called using pthread_init_once?

ldorau: Yes. Fixed.
Done.

src/proxy_lib/proxy_lib.h
ldorau: OK. There is an error during linking on Windows. I thought it was needed to fix it, but it didn't help.
@igchor Do you know maybe how fix this error?
Removed.
Done.

src/proxy_lib/proxy_lib.c
```diff
@@ -0,0 +1,168 @@
...
+
+static UTIL_ONCE_FLAG alloc_hooks_are_initialized = UTIL_ONCE_FLAG_INIT;
+
+static umf_memory_provider_handle_t OS_memory_provider = NULL;
+static umf_memory_pool_handle_t Proxy_pool = NULL;
+
+// it protects us from recursion in umfPool*()
+static __TLS int was_called_from_umfPool = 0;
+
+malloc_func_t System_malloc_func;
```
igchor, line 30: Why do we use original malloc functions instead of just using base_alloc or linear alloc? WIth one of those, we wouldn't even need to load malloc* symbols, right?

ldorau: See the comment added just now:
```c
/*
 * The UMF proxy library uses three memory allocators:
 * 1) the internal linear base allocator (LBA) based on the anonymous mapped memory
 * 2) the system allocator (libc malloc on Linux for example)
 * 3) the main one - UMF pool allocator.
 *
 * Ad 1)
 * The internal linear base allocator (LBA) is used from the very beginning
 * to the start of the constructor of the proxy library to allocate memory
 * for OS specific structures used during loading and starting applications
 * (for example _dl_init() and _dl_fini() on Linux storing arrays of all
 * constructors and destructors that have to be called). That memory will be leaked
 * on purpose (we believe OS will free it), because we cannot free the memory
 * containing array of destructors that have to be called at the end.
 * There are only 4 such allocations on Linux now:
 * - realloc(0x0, 480)
 * - malloc(20)
 * - malloc(24)
 * - malloc(72704)
 *
 * Ad 2)
 * The system allocator is used from the start of the constructor of the proxy library
 * to the creation of a UMF pool to allocate memory needed by umfMemoryProviderCreate()
 * and umfPoolCreate(). It will be freed by the system allocator.
 *
 * Ad 3)
 * The UMF pool allocator (the main one) is used from the creation to the destruction
 * of a UMF pool to allocate memory needed by an application. It should be freed
 * by an application.
 */
```

ldorau:
@igchor There is an issue with #229. I run ubench with proxy library in LD_PRELOAD:
```
$ LD_PRELOAD=./lib/libumf_proxy.so ./benchmark/ubench
```
1) The constructor of libumf_proxy.so is called.
2) The constructor runs `umfMemoryProviderCreate(umfOsMemoryProviderOps())` that creates the global base allocator no. 1
3) The constructor runs `umfPoolCreate(umfJemallocPoolOps())` that creates the global base allocator no. 2 and allocates je_malloc pool inside it.
4) The benchmark runs and ends.
5) `atexit(umf_ba_destroy_global);` is called and the global base allocator no. 2 is destroyed.
6) The destructor of libumf_proxy.so is called.
7) The destructor of libumf_proxy.so runs `umfPoolDestroy(pool)` and it blows up, because the memory it uses has already been freed in 5)

src/proxy_lib/proxy_lib.c
```diff
@@ -0,0 +1,291 @@
+/*
+ * Copyright (C) 2024 Intel Corporation
+ *
+ * Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+*/
+
+// UMF proxy library
```
lukaszstolarczuk, line 8: can you extend this description a little - what is a proxy lib?

src/proxy_lib/proxy_lib.c
lukaszstolarczuk, line 1: Please add any description to the commit "Add proxy library"

src/CMakeLists.txt
```diff
@@ -103,3 +103,6 @@ install(TARGETS umf
 )
 
 add_subdirectory(pool)
+if(LINUX) # TODO: enable it on Windows
```
lukaszstolarczuk, line 106: if LINUX and UMF_BUILD_LIBUMF_POOL_JEMALLOC

ldorau:
Build issue on Windows:
https://github.com/oneapi-src/unified-memory-framework/actions/runs/7877470520/job/21493692037?pr=226

src/proxy_lib/proxy_lib.c
ldorau: Done

src/proxy_lib/proxy_lib.c
ldorau: Done

src/CMakeLists.txt
ldorau: Done

src/proxy_lib/proxy_lib.c
igchor: > See the comment added just now:
> 
> ```c
> /*
>  * The UMF proxy library uses three memory allocators:
>  * 1) the internal linear base allocator (LBA) based on the anonymous mapped memory
>  * 2) the system allocator (libc malloc on Linux for example)
>  * 3) the main one - UMF pool allocator.
>  *
>  * Ad 1)
>  * The internal linear base allocator (LBA) is used from the very beginning
>  * to the start of the constructor of the proxy library to allocate memory
>  * for OS specific structures used during loading and starting applications
>  * (for example _dl_init() and _dl_fini() on Linux storing arrays of all
>  * constructors and destructors that have to be called). That memory will be leaked
>  * on purpose (we believe OS will free it), because we cannot free the memory
>  * containing array of destructors that have to be called at the end.
>  * There are only 4 such allocations on Linux now:
>  * - realloc(0x0, 480)
>  * - malloc(20)
>  * - malloc(24)
>  * - malloc(72704)
>  *
>  * Ad 2)
>  * The system allocator is used from the start of the constructor of the proxy library
>  * to the creation of a UMF pool to allocate memory needed by umfMemoryProviderCreate()
>  * and umfPoolCreate(). It will be freed by the system allocator.
>  *
>  * Ad 3)
>  * The UMF pool allocator (the main one) is used from the creation to the destruction
>  * of a UMF pool to allocate memory needed by an application. It should be freed
>  * by an application.
>  */
> ```
> 
> So the short answer is: to not leak too much memory.

I still don't quite understand why we would leak memory when using base_alloc. Are you worried that there are multiple different allocation sizes and creating a single base_alloc pool would waste too much memory? If so, we can probably just use the version from here: https://github.com/oneapi-src/unified-memory-framework/pull/203 right?

igchor:
> Build issue on Windows: https://github.com/oneapi-src/unified-memory-framework/actions/runs/7877470520/job/21493692037?pr=226

This seems to be caused by the fact that we include stdlib.h where malloc and others are defined (probably with dllimport). The fix is to not include stdlib.h in the file that implements proxy_lib

src/utils/utils_common.h
```diff
@@ -96,6 +96,17 @@ static inline size_t align_size(size_t size, size_t alignment) {
     }
 }
 
+static inline void *align_ptr(void *ptr, size_t alignment) {
+    // align pointer to 'alignment' bytes
+    uintptr_t uptr = (uintptr_t)ptr;
+    uintptr_t rest = uptr & (alignment - 1);
+    if (rest) {
```
igchor, line 103: I think you can avoid this check, you could probably copy the implementation from pmdk: https://github.com/pmem/pmdk/blob/e8d67f4b0d8c47d4e1d36cfcbd71726638b1f1e1/src/core/util.h#L73

src/proxy_lib/proxy_lib.c
ldorau: @igchor On the other hand - what is wrong in using original malloc functions and loading malloc* symbols?

src/utils/utils_common.h
```diff
@@ -96,6 +96,17 @@ static inline size_t align_size(size_t size, size_t alignment) {
     }
 }
 
+static inline void *align_ptr(void *ptr, size_t alignment) {
```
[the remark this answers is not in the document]
ldorau: Done

src/utils/utils_common.h
ldorau: Done

src/proxy_lib/proxy_lib.c
ldorau: > I still don't quite understand why we would leak memory when using base_alloc. Are you worried that there are multiple different allocation sizes and creating a single base_alloc pool would waste too much memory? If so, we can probably just use the version from here: #203 right? (we might need to just tweak the allocation class sizes)

We cannot use the version from #203 since `free()` requires the size of the allocation in this version, but we do not know that (allocations are done inside hwloc for example).
The only solution I see now is to use the second instance of the linear base allocator instead of the system one.

ldorau: @igchor I have replaced the system allocator with the second instance of the linear base allocator - please review now.

igchor: Thanks, you are right about #203, we would need to pass size. 

> @igchor On the other hand - what is wrong in using original malloc functions and loading malloc* symbols?

I think loading malloc symbols just adds complexity (especially since the code is different for Windows) and we already have allocators (linear_alloc or even other memory pools that we could use with a simplified memory provider: one that doesn't need to allocate anything and just calls mmap/VirtualReserve).

ldorau: @igchor is it resolved now?

igchor:
@ldorau Will you add some tests to this PR?

ldorau:
> @igchor There is an issue with #229. I run ubench with proxy library in LD_PRELOAD:
> 
> ```
> $ LD_PRELOAD=./lib/libumf_proxy.so ./benchmark/ubench
> ```
> 
> 1. The constructor of libumf_proxy.so is called.
> 2. The constructor runs `umfMemoryProviderCreate(umfOsMemoryProviderOps())` that creates the global base allocator no. 1
> 3. The constructor runs `umfPoolCreate(umfJemallocPoolOps())` that creates the global base allocator no. 2 and allocates je_malloc pool inside it.
> 4. The benchmark runs and ends.
> 5. `atexit(umf_ba_destroy_global);` is called and the global base allocator no. 2 is destroyed.
> 6. The destructor of libumf_proxy.so is called.
> 7. The destructor of libumf_proxy.so runs `umfPoolDestroy(pool)` and it blows up, because the memory it uses has already been freed in 5)
> 
> See: https://github.com/oneapi-src/unified-memory-framework/actions/runs/7887458469/job/21523176454?pr=226

This issue has been fixed by #235

src/base_alloc/base_alloc_linear.c
```diff
@@ -175,3 +175,33 @@ void umf_ba_linear_destroy(umf_ba_linear_pool_t *pool) {
     util_mutex_destroy_not_free(&pool->metadata.lock);
     ba_os_free(pool, pool->metadata.pool_size);
 }
+
+// umf_ba_linear_pool_contains_pointer() returns:
+// - 0 if ptr does not belong to the pool or
+// - size (> 0) of the memory region from ptr
+//   to the end of the pool if ptr belongs to the pool
+size_t umf_ba_linear_pool_contains_pointer(umf_ba_linear_pool_t *pool,
```
lukaszstolarczuk, line 183: can you pls add a test for this new function?

.github/workflows/proxy_lib.yml
```diff
@@ -0,0 +1,52 @@
...
+    runs-on: ${{matrix.os}}
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+
+      - name: Install apt packages
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y cmake libhwloc-dev libnuma-dev libjemalloc-dev libtbb-dev
```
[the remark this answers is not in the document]
ldorau: Removed

.github/workflows/proxy_lib.yml
```diff
@@ -0,0 +1,52 @@
...
+          -DUMF_ENABLE_POOL_TRACKING=OFF
+
+      - name: Build UMF
+        run: cmake --build ${{github.workspace}}/build -j $(nproc)
+
+      - name: Run benchmarks
+        working-directory: ${{github.workspace}}/build
+        run: |
+          LD_DEBUG=all LD_PRELOAD=./lib/libumf_proxy.so ./benchmark/ubench
+          echo $?
```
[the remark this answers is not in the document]
ldorau: Removed.

src/base_alloc/base_alloc_linear.c
ldorau: OK. I will.

ldorau:
> @ldorau Will you add some tests to this PR?

I have added a separate workflow running `LD_PRELOAD=./lib/libumf_proxy.so ./benchmark/ubench`.
I will add more later.

ldorau:
**REMARK**
The proxy library works correctly now if the app it is loaded for **does NOT link with `libumf.so`,**
otherwise it can work incorrectly, because both the proxy library and the app use one and the same `TRACKER` and the one and the same global base allocator.

src/base_alloc/base_alloc_linear.c
ldorau: @lukaszstolarczuk test added.
Done

ldorau:
> @ldorau Will you add some tests to this PR?

I have added a separate workflow running:
```sh
          LD_PRELOAD=./lib/libumf_proxy.so /usr/bin/date
          LD_PRELOAD=./lib/libumf_proxy.so /usr/bin/ls
          LD_PRELOAD=./lib/libumf_proxy.so ./benchmark/ubench
```

@igchor is it enough?

bratpiorka:
IMO the scenario where someone uses both the proxy and the UMF API is invalid. I don't think we should support this, or at least it shouldn't be our priority right now. LGTM

bratpiorka (approved):

src/proxy_lib/proxy_lib.c
```diff
@@ -0,0 +1,253 @@
...
+        // it means free(ptr), but linear base allocator does not implement free()
+        return NULL;
+    }
+
+    assert(ptr); // it is verified in the main realloc()
+
+    size_t max_size = umf_ba_linear_pool_contains_pointer(pool, ptr);
+    assert(max_size > 0); // assert that pool contains the pointer
+
+    void *new_ptr = umf_ba_linear_alloc(pool, size);
```
igchor, line 117: There's a possible race here if some other thread calls `umf_ba_linear_pool_contains_pointer` after this thread calls `umf_ba_linear_pool_contains_pointer` and before `umf_ba_linear_alloc`.

src/proxy_lib/proxy_lib.c
```diff
@@ -0,0 +1,253 @@
...
+    if (!new_ptr) {
+        return NULL;
+    }
+
+    if (size > max_size) {
+        size = max_size;
+    }
+
+    memcpy(new_ptr, ptr, size);
```
igchor, line 126: I don;t know if this is a big problem but you shouldn't call `memcpy(new_ptr, ptr, size);` - you should call memcpy with OLD size, not the new one I think. As it is now, you are potentially reading memory that belong to another allocation which can lead to races (and possibly be UB).

src/proxy_lib/proxy_lib.c
```diff
@@ -0,0 +1,253 @@
...
+ * It is used to allocate memory for OS specific structures used during loading and starting
+ * applications (for example _dl_init() and _dl_fini() on Linux storing arrays of all
+ * constructors and destructors that have to be called) and to allocate memory needed
+ * by umfMemoryProviderCreate() and umfPoolCreate().
+ * That memory will be leaked on purpose (we believe OS will free it),
+ * because we cannot free the memory containing array of destructors
+ * that have to be called at the end.
+ * It uses 892 kB on Linux.
+ *
+ * Ad 2)
```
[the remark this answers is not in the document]
ldorau: "... because we cannot free the memory containing data of destructors that have to be called at the end (for example memory allocated by _dl_init() and used internally by _dl_fini() on Linux)." - see the comment 6 lines above:

```c
 * It is used to allocate memory for OS specific data used during loading and unloading
 * applications (for example _dl_init() and _dl_fini() on Linux storing data of all
 * constructors and destructors that have to be called) and also memory needed
 * by umfMemoryProviderCreate() and umfPoolCreate().
 * That memory will be leaked on purpose (OS will have to free it during destroying
 * the process), because we cannot free the memory containing data of destructors
 * that have to be called at the end (for example memory allocated by _dl_init()
 * and used internally by _dl_fini() on Linux).
```

src/proxy_lib/proxy_lib.c
ldorau: 1) The linear allocator is used mainly in the phase of loading the application in the single-threaded environment.
2) It can be used in the MT environment only in the case of "recursive" calls when memory is allocated inside of the external libraries (hwloc for example). In such case I do not see anything dangerous in this race, because the linear allocator has an internal lock. If you see a dangerous scenario, explain it in details, please.

src/proxy_lib/proxy_lib.c
ldorau: 1) We do NOT know the OLD size (storing it would complicate the allocator too much)
2) Reading the extra memory is not a problem, because we do not use it.

src/proxy_lib/proxy_lib.c
```diff
@@ -0,0 +1,253 @@
...
+
+    size_t max_size = umf_ba_linear_pool_contains_pointer(pool, ptr);
+    assert(max_size > 0); // assert that pool contains the pointer
+
+    void *new_ptr = umf_ba_linear_alloc(pool, size);
+    if (!new_ptr) {
+        return NULL;
+    }
+
+    if (size > max_size) {
```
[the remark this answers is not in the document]
ldorau: because always `max_size >= old_size` so it is OK

ldorau:
> > IMO the scenario where someone uses both the proxy and the UMF API is invalid. I don't think we should support this, or at least it shouldn't be our priority right now. LGTM
> 
> To be honest, I think it SHOULD work in this scenario. If both proxy_lib and app uses libumf.so (dynamic lib) then there should be only one base_alloc and tracker. I think there is some other issue that we should solve here.

The occurring issues can be caused by the fact that there is only one base_alloc and one tracker in case of dynamic linking. I think that the **proxy library should use a separate base_alloc and a separate tracker** than an app.

vinser52:
> IMO the scenario where someone uses both the proxy and the UMF API is invalid. I don't think we should support this, or at least it shouldn't be our priority right now. LGTM

Did we discussed the scenarios that we are going to cover with this proxy library at our Memory WG? We also have not discussed the architecture and potential issues. Probably TBB malloc already solves such issues somehow, we have not investigated this. Furthermore, if we consider migrating TBB malloc and building it on top of UMF then there is a TBB malloc proxy that intercepts regular malloc/free flow- do we need to duplicate such functionality with umf_proxy? Why not to discuss this first and get feedback from our Memory WG attendees? Do we have any urgency with this proxy library?
Also, as I remember CAL needs some functionality to intercept malloc/free for GPU use cases.

ldorau:
> > IMO the scenario where someone uses both the proxy and the UMF API is invalid. I don't think we should support this, or at least it shouldn't be our priority right now. LGTM
> 
> To be honest, I think it SHOULD work in this scenario. If both proxy_lib and app uses libumf.so (dynamic lib) then there should be only one base_alloc and tracker. I think there is some other issue that we should solve here.

@bratpiorka @igchor 
It DOES WORK in this scenario too with scalable pool. It turned out that those issues occur **only with jemalloc pool**, so there can be an issue with jemalloc pool. 

The proxy library with scalable pool works well.

src/proxy_lib/proxy_lib.c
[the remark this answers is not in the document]
igchor: Right, I get that you cannot free memory for the constructors/destructors but you're also saying this allocator is used for memory needed by umfMemoryProviderCreate() and umfPoolCreate(). Those allocations can be freed safely, right?

I remember you had a 3rd allocator type (linear alloc) that we would release in the proxy_lib dtor. Is something different in this version that you removed it?

src/proxy_lib/proxy_lib.c
igchor: Ok, let's keep it as is then. Just FYI, data races are always UB so even if we don't do anything with the data it might, in theory, lead to some problems (although in this scenario it's probably very unlikely).

src/proxy_lib/proxy_lib.c
igchor: Ok, you're right - I misunderstood how this code is used.

igchor:
> > @ldorau Will you add some tests to this PR?
> 
> I have added a separate workflow running:
> 
> ```shell
>           LD_PRELOAD=./lib/libumf_proxy.so /usr/bin/date
>           LD_PRELOAD=./lib/libumf_proxy.so /usr/bin/ls
>           LD_PRELOAD=./lib/libumf_proxy.so ./benchmark/ubench
> ```
> 
> @igchor is it enough?

For now that's fine I think. But we should add some extra tests with a lot of mallocs/callocs/reallocs (even from multiple threads) as a next step.

src/proxy_lib/proxy_lib.c
```diff
@@ -0,0 +1,263 @@
...
+    void *ptr = umf_ba_linear_alloc(Base_alloc_leak, size + alignment);
+    return (void *)ALIGN_UP((uintptr_t)ptr, alignment);
+}
+
+static inline size_t ba_leak_pool_contains_pointer(void *ptr) {
+    return umf_ba_linear_pool_contains_pointer(Base_alloc_leak, ptr);
+}
+
+/*****************************************************************************/
+/*** The UMF pool allocator functions (the public API) ***********************/
```
igchor, line 176: Should we support other functions, like posix_memaling first? Perhaps we should look at what's supported in https://github.com/pmem/vmem/blob/master/src/libvmmalloc/libvmmalloc.c

src/proxy_lib/proxy_lib.c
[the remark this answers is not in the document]
ldorau: The linear base allocator does not implement free(), so we can only destroy the whole allocator and free all allocations or - the second option - free nothing. We do not know which allocations come from _dl_init() and which come from hwloc for example. So the only possible option is to free nothing. When there were three allocators, two of them were linear base allocators. The first was a "leak" allocator and the second was destroyed, but this was wrong. It turned out that none of them can be destroyed, so only one "leak" linear base allocator is enough.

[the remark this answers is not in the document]
igchor: Ok, I understand it now. But still, eventually, it would be good to fix this (perhaps similarly as in libvmmaloc where they just use jemalloc with a default hooks or by extending our base_alloc). Could you create an issue/task for that?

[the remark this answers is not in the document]
ldorau: > The linear base allocator does not implement free(), so we can only destroy the whole allocator and free all allocations or - the second option - free nothing. We do not know which allocations come from _dl_init() and which come from hwloc for example. So the only possible option is to free nothing. When there were three allocators, two of them were linear base allocators. The first was a "leak" allocator and the second was destroyed, but this was wrong. It turned out that none of them can be destroyed, so only one "leak" linear base allocator is enough.

@bratpiorka @igchor Moreover, for the same reason we cannot destroy the UMF pool either! It turned out that in the phase of loading libraries, when allocations come from the UMF pool, the program loader (maybe _dl_init()) allocates new memory for the app's environment that is read at the very end after unloading all libraries and after all destructors were finished - `getenv()` is called and there is a segfault caused by use-after-free, because this part of environment was freed earlier during destroying the UMF pool in the destructor of the proxy library ...

ldorau:
> The linear base allocator does not implement free(), so we can only destroy the whole allocator and free all allocations or - the second option - free nothing. We do not know which allocations come from _dl_init() and which come from hwloc for example. So the only possible option is to free nothing. When there were three allocators, two of them were linear base allocators. The first was a "leak" allocator and the second was destroyed, but this was wrong. It turned out that none of them can be destroyed, so only one "leak" linear base allocator is enough.

@bratpiorka @igchor Moreover, for the same reason we cannot destroy the UMF pool either! It turned out that in the phase of loading libraries, when allocations come from the UMF pool, the program loader (maybe _dl_init()) allocates new memory for the app's environment that is read at the very end after unloading all libraries and after all destructors were finished - `getenv()` is called and there is a segfault caused by use-after-free, because this part of environment was freed earlier during destroying the UMF pool in the destructor of the proxy library ...

src/proxy_lib/proxy_lib.c
ldorau: The UMF pool API (`umf_memory_pool_ops_t`) supports only the following ones:
```c
    void *(*malloc)(void *pool, size_t size);
    void *(*calloc)(void *pool, size_t num, size_t size);
    void *(*realloc)(void *pool, void *ptr, size_t size);
    void *(*aligned_malloc)(void *pool, size_t size, size_t alignment);
    size_t (*malloc_usable_size)(void *pool, void *ptr);
    umf_result_t (*free)(void *pool, void *);
```

ldorau:
The proxy library works correctly with all following applications now:
```sh
$ LD_PRELOAD=./lib/libumf_proxy.so ctest --output-on-failure
$ LD_PRELOAD=./lib/libumf_proxy.so ./benchmark/ubench
$ LD_PRELOAD=./lib/libumf_proxy.so ./test/umf_test-memoryPool
$ LD_PRELOAD=./lib/libumf_proxy.so /usr/bin/ls
$ LD_PRELOAD=./lib/libumf_proxy.so /usr/bin/date
```

src/proxy_lib/proxy_lib.c
ldorau: We can add more functions when we all agree what API exactly should be supported. It can be added in a separate PR too.

bratpiorka:
With the latest changes LGTM. We should make more improvements in separate PRs.

src/CMakeLists.txt
```diff
@@ -103,3 +103,12 @@ install(TARGETS umf
 )
 
 add_subdirectory(pool)
+
+# TODO: enable proxy_lib on Windows
+if(LINUX)
+    if(UMF_BUILD_LIBUMF_POOL_JEMALLOC)
```
lukaszstolarczuk, line 109: how does it work with `PROXY_LIB_USES_SCALABLE_POOL` variable set in proxy_lib/CMakeLists.txt?

src/utils/utils_common.h
```diff
@@ -19,11 +19,45 @@
...
+
+static inline char *os_getenv(const char *name) { return getenv(name); }
+static inline void os_free_getenv(const char *val) {
+    (void)val; // unused
+}
+
+#endif /* _WIN32 */
+
+// check if we are running in the proxy library
+static inline int is_running_in_proxy_lib(void) {
```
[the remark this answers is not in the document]
igchor: Btw, it would be nice to have a test for such scenario.

igchor (approved):

src/utils/utils_common.h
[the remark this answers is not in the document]
ldorau: > Will this work if someone links with proxy_lib? (I think we should support this scenario - from what I understand, this is how it will work on windows) Or is there no problem in that case?

Do you mean the dynamic linking instead of LD_PRELOD?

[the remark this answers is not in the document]
igchor: Yes

src/CMakeLists.txt
ldorau: Done

src/CMakeLists.txt
```diff
@@ -103,3 +108,26 @@ install(TARGETS umf
...
+
+# TODO: enable proxy_lib on Windows
+if(LINUX)
+    if(UMF_PROXY_LIB_BASED_ON_POOL STREQUAL SCALABLE)
+        set(PROXY_LIB_USES_SCALABLE_POOL ON)
+        set(PROXY_LIBS umf scalable_pool)
+        if(UMF_BUILD_LIBUMF_POOL_SCALABLE)
+            add_subdirectory(proxy_lib)
+        else()
+            message(WARNING "Disabling the proxy library, because UMF_PROXY_LIB_BASED_ON_POOL==SCALABLE but UMF_BUILD_LIBUMF_POOL_SCALABLE is OFF")
```
lukaszstolarczuk, line 120: I don't think it should be a WARNING. We use default value ("scalable" in this case), so if I decide to set `UMF_BUILD_LIBUMF_POOL_SCALABLE=OFF` I'll get a warning in each build - I don't think it's desired.

If we have an option to enable/disable proxy_lib, that would be a different story, but I wouldn't add it, I'd just change these WARNINGs to STATUSes.

src/CMakeLists.txt
```diff
@@ -103,3 +108,26 @@ install(TARGETS umf
...
+    elseif(UMF_PROXY_LIB_BASED_ON_POOL STREQUAL JEMALLOC)
+        set(PROXY_LIB_USES_JEMALLOC_POOL ON)
+        set(PROXY_LIBS umf jemalloc_pool)
+        if(UMF_BUILD_LIBUMF_POOL_JEMALLOC)
+            add_subdirectory(proxy_lib)
+        else()
+            message(WARNING "Disabling the proxy library, because UMF_PROXY_LIB_BASED_ON_POOL==JEMALLOC but UMF_BUILD_LIBUMF_POOL_JEMALLOC is OFF")
+        endif()
+    else()
+        message(FATAL_ERROR "Proxy library: pool manager not chosen")
```
lukaszstolarczuk, line 131: pls add info that it should be set via `UMF_PROXY_LIB_BASED_ON_POOL`.

lukaszstolarczuk, line 131: `...not chosen or set to a non-supported one`

src/CMakeLists.txt
```diff
@@ -4,6 +4,11 @@
 
 include(${UMF_CMAKE_SOURCE_DIR}/cmake/helpers.cmake)
 
+# set UMF_PROXY_LIB_BASED_ON_POOL to one of:
+# - SCALABLE
+# - JEMALLOC
+set(UMF_PROXY_LIB_BASED_ON_POOL SCALABLE)
```
lukaszstolarczuk, line 10: Do you want to enable setting this from the CLI level? I think you'd have to add `CACHE` (ref. https://cmake.org/cmake/help/latest/command/set.html#set-cache-entry)

src/CMakeLists.txt
ldorau: Done

src/CMakeLists.txt
ldorau: Done

ldorau: Done

src/CMakeLists.txt
ldorau: Done

lukaszstolarczuk:
Thx! LGTM

PR #227, opened 2024-02-12
Test benchmark also with UMF_BUILD_SHARED_LIBRARY=ON

ldorau:

PR #229, opened 2024-02-12
Unify global umf_ba_pool_t initialization

igchor:
Initializing the global pool in libumf and in each pool separately led to problems with double initialization when multiple pools were used in a single application (and static libumf).

To avoid this problem, use util_init_once inside umf_ba_get_pool instead of having explicit constructor/destructor.

This fixes an issue exposed by:  0c55d8f55d583e87af4f08157c0ad1c828a02ac6

### DIFF
--- a/src/base_alloc/base_alloc_global.c
+++ b/src/base_alloc/base_alloc_global.c
@@ -6,34 +6,39 @@
 */
 
 #include <assert.h>
+#include <stdlib.h>
 
 #include "base_alloc.h"
+#include "utils_concurrency.h"
 
 #define SIZE_BA_POOL_CHUNK 128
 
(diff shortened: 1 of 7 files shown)


igchor:
This should have been tested by https://github.com/oneapi-src/unified-memory-framework/pull/223 but scalable pool was missing from the test. It's fixed here: https://github.com/oneapi-src/unified-memory-framework/pull/228

ldorau (approved):

ldorau (changes requested):
There is an issue here. I am describing it just now ...

ldorau (approved):
However I think it should be merged as is, because it fixes the failure of #227 and refined later.

bratpiorka (approved):

PR #232, opened 2024-02-13
Add documentation to OS provider

igchor:
and make it more OS-agnostic

### DIFF
--- a/include/umf/providers/provider_os_memory.h
+++ b/include/umf/providers/provider_os_memory.h
@@ -16,47 +16,56 @@ extern "C" {
 
 #define UMF_OS_RESULTS_START_FROM 1000
 
+/// @brief Protection of the memory allocations
 typedef enum umf_mem_protection_flags_t {
-    UMF_PROTECTION_NONE = (1 << 0),
-    UMF_PROTECTION_READ = (1 << 1),
-    UMF_PROTECTION_WRITE = (1 << 2),
-    UMF_PROTECTION_EXEC = (1 << 3),
+    UMF_PROTECTION_NONE = (1 << 0),  ///< Memory allocations can not be accessed
+    UMF_PROTECTION_READ = (1 << 1),  ///< Memory allocations can be read.
+    UMF_PROTECTION_WRITE = (1 << 2), ///< Memory allocations can be written.
+    UMF_PROTECTION_EXEC = (1 << 3),  ///< Memory allocations can be executed.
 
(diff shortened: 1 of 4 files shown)


include/umf/providers/provider_os_memory.h
```diff
@@ -16,47 +16,56 @@ extern "C" {
...
 
     UMF_PROTECTION_MAX // must be the last one
 } umf_mem_protection_flags_t;
 
+/// @brief Visibility of the memory allocations
 typedef enum umf_mem_visibility_t {
-    UMF_VISIBILITY_SHARED,
-    UMF_VISIBILITY_PRIVATE,
+    UMF_VISIBILITY_SHARED, ///< Updates to the memory allocated using OS provider are visible to other processes.
+    /// TODO: need to expose functionality to share open the mapping in other process and explicit sync?
```
igchor, line 32: @vinser52 @pbalcer do we even want to support `UMF_VISIBILITY_SHARED` for OS provider?  I don't think we have a use case for this yet.

PatKamin (approved):

include/umf/providers/provider_os_memory.h
vinser52: MPI mentioned such a use case if I remember correctly. The IPC mechanism is supposed to be used. BUt need to investigate deeply such use case

igchor: I've created to remove this param for now: https://github.com/oneapi-src/unified-memory-framework/pull/236 It should be fairly easy to restore if we find that it's needed

src/provider/provider_os_memory.c
```diff
@@ -146,8 +146,16 @@ static hwloc_membind_policy_t translate_numa_mode(umf_numa_mode_t mode,
...
     case UMF_NUMA_MODE_BIND:
+        if (nodemaskEmpty) {
+            // nodeset must not be empty
+            return -1;
+        }
         return HWLOC_MEMBIND_BIND;
     case UMF_NUMA_MODE_INTERLEAVE:
+        if (nodemaskEmpty) {
+            // nodeset must not be empty
+            return -1;
```
bratpiorka, line 157: could you add tests for this case?

igchor: Done. I've extended the tests to cover all negative cases.

bratpiorka (approved):

PR #235, opened 2024-02-14
Destroy global base allocator in the library destructor

ldorau:
The global base allocator should be destroyed in the library destructor in order to avoid use-after-free.

### DIFF
--- a/src/base_alloc/base_alloc_global.c
+++ b/src/base_alloc/base_alloc_global.c
@@ -9,6 +9,7 @@
 #include <stdlib.h>
 
 #include "base_alloc.h"
+#include "base_alloc_global.h"
 #include "utils_concurrency.h"
 
 #define SIZE_BA_POOL_CHUNK 128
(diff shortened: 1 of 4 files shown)


bratpiorka (approved):

src/base_alloc/base_alloc.h
```diff
@@ -21,7 +21,9 @@ typedef struct umf_ba_pool_t umf_ba_pool_t;
 umf_ba_pool_t *umf_ba_create(size_t size);
 void *umf_ba_alloc(umf_ba_pool_t *pool);
 void umf_ba_free(umf_ba_pool_t *pool, void *ptr);
+void umf_ba_init(void);
 void umf_ba_destroy(umf_ba_pool_t *pool);
+void umf_ba_destroy_global(void);
```
igchor, line 26: nit: should;t this be in base_alloc_global.h? (umf_ba_init should also be there and be named umf_ba_global_init probably)

src/base_alloc/base_alloc_windows.c
```diff
@@ -5,8 +5,38 @@
...
+}
+
+#else /* UMF_SHARED_LIBRARY */
+
+INIT_ONCE BA_init_once_flag = INIT_ONCE_STATIC_INIT;
+
+BOOL CALLBACK BAinitOnceCb(PINIT_ONCE InitOnce, PVOID Parameter,
+                           PVOID *lpContext) {
+    umf_ba_constructor();
+    atexit(umf_ba_destructor);
```
igchor, line 30: won't we have the same problem with proxy lib as on linux when using atexit?

src/base_alloc/base_alloc_windows.c
```diff
@@ -5,8 +5,38 @@
  * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 */
 
+#include "base_alloc.h"
 #include <windows.h>
 
+static void umf_ba_constructor(void) {}
```
igchor, line 11: You can avoid all the changes in this file by just adding:
```
#if defined(_WIN32) && defined(UMF_SHARED_LIBRARY)
atexit(umf_ba_destroy_global);
#endif
```

to `umf_ba_create_global` in base_alloc_global.c but see my comment below about the atexit.

src/base_alloc/base_alloc.h
ldorau: Done. `umf_ba_init()` is OK.

src/base_alloc/base_alloc_windows.c
ldorau: I do not know yet. I have done that exactly like in `libumf_windows.c`.

src/base_alloc/base_alloc.h
ldorau: `umf_ba_init()` is removed.
Done.

igchor (approved):

src/base_alloc/base_alloc_windows.c
igchor: Right, the proxy lib does not support windows yet anyway,

Issue #225, opened 2024-02-09
USM support


igchor:
UMF should support managing USM (unified shared memory) by:
- exposing memory pools capable of handling memory that might not be accessible on the host:
    - [x] disjoint_pool
    - [x] jemalloc_pool
       - note: there is a known issue with jemalloc: if compiled with default flags it cannot be dlopened (neither can a library that links with jemalloc) which makes it problematic to use in Unified Runtime which manages adapters using dlopen
- exposing memory providers capable of allocating memory on a device:
   - [x] L0 provider (currently implemented in Unified Runtime)
   - [ ] CUDA provider (currently implemented in Unified Runtime)
   - [ ] Unified Runtime (?)

It should be possible to build L0 and CUDA providers as static libraries so that they can be used by Unified Runtime.

We should also consider:
- adding memory copy / memset operations for providers (to implement calloc/realloc on device-only memory)
- async memory management (transfers and/or allocation)

vinser52:
Another requirement from MPI/oneCCL. They do not have hard dependency to L0 today the same MPI/oneCCL binaries should work on the system without L0. They resolve dependency to L0 at runtime using dlopen/dlsym to look for `libze_loader.so`.
We should consider this case when developing L0 memory provider.

PR #237, opened 2024-02-14
add Level Zero memory provider

bratpiorka:
This PR adds a Level Zero memory provider that supports #225 enhancement. TODO:
- [x] Make Level Zero provider a separate lib that dynamically loads Level Zero functions from ze_loader
- [x] Create a separate workflow for this with required deps (Level Zero) that will run on GPU
- [x] Add basic test with simple allocation/deallocation
- [x] Update README and docs

Please note that shared memory would be supported in separate PR

### DIFF
--- a/.github/workflows/gpu.yml
+++ b/.github/workflows/gpu.yml
@@ -1,0 +1,49 @@
+# This workflow builds and tests providers using GPU memory. It requires 
+# "level_zero" labeled self-hosted runners installed on systems with the 
+# appropriate GPU and drivers.
+name: GPU
+
+on: [workflow_call]
+
+permissions:
+  contents: read
+
+jobs:
+  # TODO: add support for Windows
+  ubuntu-build:
+    name: Build - Ubuntu
+
+    strategy:
+      matrix:
+        os: ['ubuntu-22.04']
+        build_type: [Release]
+        compiler: [{c: gcc, cxx: g++}]
+        shared_library: ['ON', 'OFF']
+    runs-on: level_zero
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+
+      - name: Configure build
+        run: >
+          cmake
+          -B ${{github.workspace}}/build
+          -DCMAKE_BUILD_TYPE=${{matrix.build_type}}
+          -DCMAKE_C_COMPILER=${{matrix.compiler.c}}
+          -DCMAKE_CXX_COMPILER=${{matrix.compiler.cxx}}
+          -DUMF_BUILD_SHARED_LIBRARY=${{matrix.shared_library}}
+          -DUMF_BUILD_BENCHMARKS=ON
+          -DUMF_BUILD_TESTS=ON
+          -DUMF_FORMAT_CODE_STYLE=ON
+          -DUMF_DEVELOPER_MODE=ON
+          -DUMF_BUILD_LEVEL_ZERO_PROVIDER=ON
+          -DUMF_BUILD_LIBUMF_POOL_DISJOINT=ON
+          -DUMF_ENABLE_POOL_TRACKING=ON
+
+      - name: Build UMF
+        run: cmake --build ${{github.workspace}}/build -j $(nproc)
+
+      - name: Run tests
+        working-directory: ${{github.workspace}}/build
+        run: ctest --output-on-failure --test-dir test
--- a/.github/workflows/pr_push.yml
+++ b/.github/workflows/pr_push.yml
@@ -65,6 +65,7 @@ jobs:
         -DUMF_BUILD_LIBUMF_POOL_JEMALLOC=${{matrix.jemalloc}}
         -DUMF_BUILD_TESTS=ON
         -DUMF_BUILD_EXAMPLES=ON
+        -DUMF_BUILD_LEVEL_ZERO_PROVIDER=OFF
         ${{matrix.extra_build_options}}
 
     - name: Build
@@ -94,6 +95,7 @@ jobs:
         -DUMF_FORMAT_CODE_STYLE=ON
         -DUMF_BUILD_OS_MEMORY_PROVIDER=OFF
         -DUMF_BUILD_TESTS=OFF
+        -DUMF_BUILD_LEVEL_ZERO_PROVIDER=OFF
 
     - name: Check clang-format
       run: cmake --build build --target clang-format-check
@@ -113,3 +115,6 @@ jobs:
   ProxyLib:
     needs: [Build]
     uses: ./.github/workflows/proxy_lib.yml
+  GPU:
+    needs: [Build]
+    uses: ./.github/workflows/gpu.yml
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,6 +8,7 @@ project(unified-memory-framework VERSION 0.1.0 LANGUAGES C)
 # Build Options
 option(UMF_BUILD_SHARED_LIBRARY "Build UMF as shared library" OFF)
 option(UMF_BUILD_OS_MEMORY_PROVIDER "Build OS memory provider" ON)
+option(UMF_BUILD_LEVEL_ZERO_PROVIDER "Build Level Zero memory provider" ON)
 option(UMF_BUILD_LIBUMF_POOL_DISJOINT "Build the libumf_pool_disjoint static library" OFF)
 option(UMF_BUILD_LIBUMF_POOL_JEMALLOC "Build the libumf_pool_jemalloc static library" OFF)
 option(UMF_BUILD_LIBUMF_POOL_SCALABLE "Build the libumf_pool_scalable static library" OFF)
--- a/README.md
+++ b/README.md
@@ -5,6 +5,7 @@
 [![SpellCheck](https://github.com/oneapi-src/unified-memory-framework/actions/workflows/spellcheck.yml/badge.svg?branch=main)](https://github.com/oneapi-src/unified-memory-framework/actions/workflows/spellcheck.yml)
 [![GitHubPages](https://github.com/oneapi-src/unified-memory-framework/actions/workflows/docs.yml/badge.svg?branch=main)](https://github.com/oneapi-src/unified-memory-framework/actions/workflows/docs.yml)
 [![Benchmarks](https://github.com/oneapi-src/unified-memory-framework/actions/workflows/benchmarks.yml/badge.svg?branch=main)](https://github.com/oneapi-src/unified-memory-framework/actions/workflows/benchmarks.yml)
+[![GPU](https://github.com/oneapi-src/unified-memory-framework/actions/workflows/gpu.yml/badge.svg?branch=main)](https://github.com/oneapi-src/unified-memory-framework/actions/workflows/gpu.yml)
 [![Nightly](https://github.com/oneapi-src/unified-memory-framework/actions/workflows/nightly.yml/badge.svg?branch=main)](https://github.com/oneapi-src/unified-memory-framework/actions/workflows/nightly.yml)
 [![OpenSSF Scorecard](https://api.securityscorecards.dev/projects/github.com/oneapi-src/unified-memory-framework/badge)](https://securityscorecards.dev/viewer/?uri=github.com/oneapi-src/unified-memory-framework)
 
@@ -30,6 +31,10 @@ For development and contributions:
 
 For building tests, multithreaded benchmarks and Disjoint Pool:
 - C++ compiler with C++17 support
+
+For Level Zero memory provider and its tests:
+- Level Zero headers and libraries
+- compatible GPU with installed driver
 
 ### Linux
 
@@ -84,6 +89,7 @@ List of options provided by CMake:
 | - | - | - | - |
 | UMF_BUILD_SHARED_LIBRARY | Build UMF as shared library | ON/OFF | OFF |
 | UMF_BUILD_OS_MEMORY_PROVIDER | Build OS memory provider | ON/OFF | ON |
+| UMF_BUILD_LIBUMF_PROVIDER_LEVEL_ZERO | Build Level Zero memory provider | ON/OFF | OFF |
 | UMF_BUILD_LIBUMF_POOL_DISJOINT | Build the libumf_pool_disjoint static library | ON/OFF | OFF |
 | UMF_BUILD_LIBUMF_POOL_JEMALLOC | Build the libumf_pool_jemalloc static library | ON/OFF | OFF |
 | UMF_BUILD_LIBUMF_POOL_SCALABLE | Build the libumf_pool_scalable static library | ON/OFF | OFF |
@@ -122,6 +128,18 @@ A memory provider that provides memory from an operating system.
    - libhwloc-dev
 4) Required packages for tests:
    - libnuma-dev
+
+#### Level Zero memory provider (Linux-only yet)
+
+A memory provider that provides memory from L0 device.
+
+##### Requirements
+
+1) System with Level Zero compatible GPU
+2) Linux OS
+3) The `UMF_BUILD_LIBUMF_PROVIDER_LEVEL_ZERO` option turned `ON`
+4) Required packages:
+   - level-zero-dev (TODO: check this)
 
 ### Memory pool managers
 
--- a/include/umf/providers/provider_level_zero.h
+++ b/include/umf/providers/provider_level_zero.h
@@ -1,0 +1,41 @@
+/*
+ * Copyright (C) 2024 Intel Corporation
+ *
+ * Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+*/
+
+#ifndef UMF_PROVIDER_LEVEL_ZERO_H
+#define UMF_PROVIDER_LEVEL_ZERO_H
+
+#include "umf/memory_provider.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/// @brief USM memory allocation type
+typedef enum umf_usm_memory_type_t {
+    UMF_MEMORY_TYPE_UNKNOWN = 0,
+    UMF_MEMORY_TYPE_HOST,
+    UMF_MEMORY_TYPE_DEVICE,
+    UMF_MEMORY_TYPE_SHARED,
+} umf_usm_memory_type_t;
+
+/// @brief Level Zero Memory Provider settings struct
+typedef struct level_zero_memory_provider_params_t {
+    void *level_zero_context_handle;
+    void *level_zero_device_handle;
+    umf_usm_memory_type_t memory_type;
+    uint32_t level_zero_host_mem_alloc_flags;
+    uint32_t level_zero_device_mem_alloc_flags;
+    uint32_t level_zero_device_local_mem_ordinal;
+} level_zero_memory_provider_params_t;
+
+umf_memory_provider_ops_t *umfLevelZeroMemoryProviderOps(void);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* UMF_PROVIDER_LEVEL_ZERO_H */
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -77,6 +77,18 @@ if(UMF_BUILD_OS_MEMORY_PROVIDER)
     endif()
 endif()
 
+if(UMF_BUILD_LEVEL_ZERO_PROVIDER)
+    # Currently, Level Zero provider is only built for Linux
+    # TODO add fetch content or PkgConfig/find_package for Level Zero
+    set(UMF_SOURCES_LINUX ${UMF_SOURCES_LINUX}
+        provider/provider_level_zero.c)
+
+    if(LINUX)
+        set(UMF_COMPILE_DEFINITIONS ${UMF_COMPILE_DEFINITIONS} 
+            "UMF_BUILD_LEVEL_ZERO_PROVIDER=1")
+    endif()
+endif()
+
 if(LINUX)
     set(UMF_SOURCES ${UMF_SOURCES} ${UMF_SOURCES_LINUX})
 elseif(WINDOWS)
--- a/src/libumf.map
+++ b/src/libumf.map
@@ -6,6 +6,7 @@ UMF_1.0 {
     global:
         umfFree;
         umfGetLastFailedMemoryProvider;
+        umfLevelZeroMemoryProviderOps;
         umfMemoryProviderAlloc;
         umfMemoryProviderAllocationMerge;
         umfMemoryProviderAllocationSplit;
--- a/src/provider/provider_level_zero.c
+++ b/src/provider/provider_level_zero.c
@@ -1,0 +1,274 @@
+/*
+ * Copyright (C) 2024 Intel Corporation
+ *
+ * Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+*/
+
+#include <assert.h>
+#include <dlfcn.h>
+#include <pthread.h>
+#include <stdbool.h>
+#include <stddef.h>
+#include <unistd.h>
+
+#include <level_zero/ze_api.h>
+
+#include <umf.h>
+#include <umf/memory_provider_ops.h>
+#include <umf/providers/provider_level_zero.h>
+
+#include "base_alloc_global.h"
+#include "utils_common.h"
+#include "utils_concurrency.h"
+#include "utils_sanitizers.h"
+
+typedef struct ze_memory_provider_t {
+    ze_context_handle_t context;
+    ze_device_handle_t device;
+    ze_memory_type_t memory_type;
+} ze_memory_provider_t;
+
+typedef struct ze_ops_t {
+    ze_result_t (*zeMemAllocHost)(ze_context_handle_t,
+                                  const ze_host_mem_alloc_desc_t *, size_t,
+                                  size_t, void *);
+    ze_result_t (*zeMemAllocDevice)(ze_context_handle_t,
+                                    const ze_device_mem_alloc_desc_t *, size_t,
+                                    size_t, ze_device_handle_t, void *);
+    ze_result_t (*zeMemAllocShared)(ze_context_handle_t,
+                                    const ze_device_mem_alloc_desc_t *,
+                                    const ze_host_mem_alloc_desc_t *, size_t,
+                                    size_t, ze_device_handle_t, void *);
+    ze_result_t (*zeMemFree)(ze_context_handle_t, void *);
+} ze_ops_t;
+
+static ze_ops_t g_ze_ops;
+static UTIL_ONCE_FLAG ze_is_initialized = UTIL_ONCE_FLAG_INIT;
+static bool Init_ze_global_state_failed;
+
+static void init_ze_global_state(void) {
+    // check if Level Zero shared library is already loaded
+    // we pass 0 as a handle to search the global symbol table
+    *(void **)&g_ze_ops.zeMemAllocHost = dlsym(0, "zeMemAllocHost");
+    *(void **)&g_ze_ops.zeMemAllocDevice = dlsym(0, "zeMemAllocDevice");
+    *(void **)&g_ze_ops.zeMemAllocShared = dlsym(0, "zeMemAllocShared");
+    *(void **)&g_ze_ops.zeMemFree = dlsym(0, "zeMemFree");
+
+    if (!g_ze_ops.zeMemAllocHost || !g_ze_ops.zeMemAllocDevice ||
+        !g_ze_ops.zeMemAllocShared || !g_ze_ops.zeMemFree) {
+        fprintf(stderr, "Required Level Zero symbols not found.\n");
+        Init_ze_global_state_failed = true;
+    }
+}
+
+enum umf_result_t ze_memory_provider_initialize(void *params, void **provider) {
+    if (provider == NULL || params == NULL) {
+        return UMF_RESULT_ERROR_INVALID_ARGUMENT;
+    }
+
+    level_zero_memory_provider_params_t *ze_params =
+        (level_zero_memory_provider_params_t *)params;
+
+    util_init_once(&ze_is_initialized, init_ze_global_state);
+    if (Init_ze_global_state_failed) {
+        fprintf(stderr, "Loading Level Zero symbols failed\n");
+        return UMF_RESULT_ERROR_UNKNOWN;
+    }
+
+    ze_memory_provider_t *ze_provider =
+        umf_ba_global_alloc(sizeof(ze_memory_provider_t));
+    if (!ze_provider) {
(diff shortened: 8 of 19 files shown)


igchor:
@bratpiorka can you describe what's your plan regarding integration with UR? Do we want to have a duplicate provider in UMF for now or are you aiming to use this provider in UR? If it's the latter, there are more features you need to support: see the original implementation here https://github.com/oneapi-src/unified-runtime/blob/main/source/adapters/level_zero/usm.cpp#L170

igchor:
Also, the L0 provider should be a separate static library: see https://github.com/oneapi-src/unified-memory-framework/issues/225

bratpiorka:
@igchor thanks for the links. What do you think about creating a basic functionality first and then adding more features (in separate PRs) so it could finally replace the providers used in UR?

igchor:
> @igchor thanks for the links. What do you think about creating a basic functionality first and then adding more features (in separate PRs) so it could finally replace the providers used in UR?

Sure, that sounds reasonable, I just wanted to highlight that there might be some gaps with the current implementation.

include/umf/providers/provider_ze_memory.h
```diff
@@ -0,0 +1,32 @@
...
+#endif
+
+/// @brief ze memory provider settings struct
+typedef struct ze_memory_provider_params_t {
+    ze_context_handle_t context;
+    ze_device_handle_t device;
+    ze_device_mem_alloc_desc_t device_mem_alloc_desc;
+} ze_memory_provider_params_t;
+
+umf_memory_provider_ops_t *umfZeMemoryProviderOps(void);
```
igchor, line 26: I think it should be called umfLevelZeroMemoryProvider (file as well). It would be more consistent with UR naming.

src/libumf.map
```diff
@@ -35,6 +35,7 @@ UMF_1.0 {
         umfMemoryProviderCreateFromMemspace;
         umfMemspaceCreateFromNumaArray;
         umfMemspaceDestroy;
+        umfZeMemoryProviderOps;
```
igchor, line 38: What about windows .map?

src/CMakeLists.txt
```diff
@@ -58,6 +58,14 @@ if(UMF_BUILD_OS_MEMORY_PROVIDER)
     endif()
 endif()
 
+if(UMF_BUILD_ZE_MEMORY_PROVIDER)
+    set(UMF_SOURCES_LINUX ${UMF_SOURCES_LINUX}
+        provider/provider_ze_memory.c)
+    if(LINUX)
+        set(UMF_LIBS ${UMF_LIBS} ze_loader)
```
igchor, line 65: We should either:
- look for L0 using find_package() or similar
- use FetchContent to download L0 header (UR does that)

src/CMakeLists.txt
```diff
@@ -58,6 +58,14 @@ if(UMF_BUILD_OS_MEMORY_PROVIDER)
     endif()
 endif()
 
+if(UMF_BUILD_ZE_MEMORY_PROVIDER)
+    set(UMF_SOURCES_LINUX ${UMF_SOURCES_LINUX}
+        provider/provider_ze_memory.c)
+    if(LINUX)
```
igchor, line 64: Why is this linux-specific? L0 also works on windows

src/provider/provider_ze_memory.c
```diff
@@ -0,0 +1,140 @@
...
+        (ze_memory_provider_t *)malloc(sizeof(ze_memory_provider_t));
+    if (!ze_provider) {
+        return UMF_RESULT_ERROR_OUT_OF_HOST_MEMORY;
+    }
+
+    ze_provider->context = ze_params->context;
+    ze_provider->device = ze_params->device;
+    ze_provider->device_mem_alloc_desc = ze_params->device_mem_alloc_desc;
+
+    assert(ze_provider->device_mem_alloc_desc.stype ==
```
igchor, line 40: Return INVALID_ARGUMENT here?

CMakeLists.txt
```diff
@@ -8,6 +8,7 @@ project(unified-memory-framework VERSION 0.1.0 LANGUAGES C)
 # Build Options
 option(UMF_BUILD_SHARED_LIBRARY "Build UMF as shared library" OFF)
 option(UMF_BUILD_OS_MEMORY_PROVIDER "Build OS memory provider" ON)
+option(UMF_BUILD_ZE_MEMORY_PROVIDER "Build Level Zero memory provider" ON)
```
igchor, line 11: Default value (ON) does not match the README

vinser52:
> @igchor thanks for the links. What do you think about creating a basic functionality first and then adding more features (in separate PRs) so it could finally replace the providers used in UR?

But what is the issue of migrating the whole UR implementation as is? Probably with some minor tweaks

vinser52:
> Also, the L0 provider should be a separate static library: see #225

@igchor did you mean that L0 provider should be only static library or both: static and dynamic?

bratpiorka:
> > @igchor thanks for the links. What do you think about creating a basic functionality first and then adding more features (in separate PRs) so it could finally replace the providers used in UR?
> 
> But what is the issue of migrating the whole UR implementation as is? Probably with some minor tweaks

This version of the Level Zero provider is very simple and is intended to be used by oneCCL, but does not contain any oneCCL-specific code. On the other hand, the UR provider contains a lot of UR-specific code and is quite complex. Making it generic would require a lot of work. Also keep in mind that most of the contents of this PR are additional files needed by umf: workflows, cmake files, def/map etc. So even with this PR porting is still an option - it wouldn't change a lot of code.

igchor:
> > Also, the L0 provider should be a separate static library: see #225
> 
> @igchor did you mean that L0 provider should be only static library or both: static and dynamic?

I think we need the possibility to build it as a static library for UR. As for dynamic library, I don't know at this moment.

vinser52:
> > > @igchor thanks for the links. What do you think about creating a basic functionality first and then adding more features (in separate PRs) so it could finally replace the providers used in UR?
> > 
> > 
> > But what is the issue of migrating the whole UR implementation as is? Probably with some minor tweaks
> 
> This version of the Level Zero provider is very simple and is intended to be used by oneCCL, but does not contain any oneCCL-specific code. On the other hand, the UR provider contains a lot of UR-specific code and is quite complex. Making it generic would require a lot of work. Also keep in mind that most of the contents of this PR are additional files needed by umf: workflows, cmake files, def/map etc. So even with this PR porting is still an option - it wouldn't change a lot of code.

My point is since UR already uses UMF and we agreed to move L0 provider impl from UR repo to UMF we should migrate it first (of course UR specific code should be cleaned up) and make sure we are not breaking UR flow. And then we can test if it fits for oneCCL needs.

vinser52:
> > > Also, the L0 provider should be a separate static library: see #225
> > 
> > 
> > @igchor did you mean that L0 provider should be only static library or both: static and dynamic?
> 
> I think we need the possibility to build it as a static library for UR. As for dynamic library, I don't know at this moment.

Will it introduce any runtime dependencies to some L0 components? It is important in context of MPI, today they have no link dependency to L0 they try to load L0 at runtime (using dlopen) if I am correct. I will double-check.

bratpiorka:
> > > > @igchor thanks for the links. What do you think about creating a basic functionality first and then adding more features (in separate PRs) so it could finally replace the providers used in UR?
> > > 
> > > 
> > > But what is the issue of migrating the whole UR implementation as is? Probably with some minor tweaks
> > 
> > 
> > This version of the Level Zero provider is very simple and is intended to be used by oneCCL, but does not contain any oneCCL-specific code. On the other hand, the UR provider contains a lot of UR-specific code and is quite complex. Making it generic would require a lot of work. Also keep in mind that most of the contents of this PR are additional files needed by umf: workflows, cmake files, def/map etc. So even with this PR porting is still an option - it wouldn't change a lot of code.
> 
> My point is since UR already uses UMF and we agreed to move L0 provider impl from UR repo to UMF we should migrate it first (of course UR specific code should be cleaned up) and make sure we are not breaking UR flow. And then we can test if it fits for oneCCL needs.



> > > > Also, the L0 provider should be a separate static library: see #225
> > > 
> > > 
> > > @igchor did you mean that L0 provider should be only static library or both: static and dynamic?
> > 
> > 
> > I think we need the possibility to build it as a static library for UR. As for dynamic library, I don't know at this moment.
> 
> Will it introduce any runtime dependencies to some L0 components? It is important in context of MPI, today they have no link dependency to L0 they try to load L0 at runtime (using dlopen) if I am correct. I will double-check.

oneCCL also loads L0 at runtime

igchor:
> > > > Also, the L0 provider should be a separate static library: see #225
> > > 
> > > 
> > > @igchor did you mean that L0 provider should be only static library or both: static and dynamic?
> > 
> > 
> > I think we need the possibility to build it as a static library for UR. As for dynamic library, I don't know at this moment.
> 
> Will it introduce any runtime dependencies to some L0 components? It is important in context of MPI, today they have no link dependency to L0 they try to load L0 at runtime (using dlopen) if I am correct. I will double-check.

@vinser52  Okay, then in that case we'll need a dynamic version I guess. Can you add this info to the issue so that we don't miss it?

bratpiorka:
> > > > @igchor thanks for the links. What do you think about creating a basic functionality first and then adding more features (in separate PRs) so it could finally replace the providers used in UR?
> > > 
> > > 
> > > But what is the issue of migrating the whole UR implementation as is? Probably with some minor tweaks
> > 
> > 
> > This version of the Level Zero provider is very simple and is intended to be used by oneCCL, but does not contain any oneCCL-specific code. On the other hand, the UR provider contains a lot of UR-specific code and is quite complex. Making it generic would require a lot of work. Also keep in mind that most of the contents of this PR are additional files needed by umf: workflows, cmake files, def/map etc. So even with this PR porting is still an option - it wouldn't change a lot of code.
> 
> My point is since UR already uses UMF and we agreed to move L0 provider impl from UR repo to UMF we should migrate it first (of course UR specific code should be cleaned up) and make sure we are not breaking UR flow. And then we can test if it fits for oneCCL needs.

I looked into the code more deeply + consulted with @igchor and we could break the existing L0 provider in UR into two providers: a UR-specific top-level provider (implemented in UR) and a low-level generic L0 provider that just translates UMF provider API to ZE calls. With this approach, the provider implementation in this PR should be really simple and generic so it could be used by both UR and oneCCL

vinser52:
> > > > > Also, the L0 provider should be a separate static library: see #225
> > > > 
> > > > 
> > > > @igchor did you mean that L0 provider should be only static library or both: static and dynamic?
> > > 
> > > 
> > > I think we need the possibility to build it as a static library for UR. As for dynamic library, I don't know at this moment.
> > 
> > 
> > Will it introduce any runtime dependencies to some L0 components? It is important in context of MPI, today they have no link dependency to L0 they try to load L0 at runtime (using dlopen) if I am correct. I will double-check.
> 
> @vinser52 Okay, then in that case we'll need a dynamic version I guess. Can you add this info to the issue so that we don't miss it?

Sure, I will update the issue. But, the dynamic version can be omitted if our static version will try to dlopen/dlsym `libze_loader.so` - In that case we are not making any hard dependency.

include/umf/providers/provider_ze_memory.h
bratpiorka: changed

src/CMakeLists.txt
bratpiorka: I changed this and ze_loader is loaded in runtime using dlopen

src/provider/provider_ze_memory.c
bratpiorka: done

CMakeLists.txt
bratpiorka: done

.github/workflows/gpu.yml
```diff
@@ -0,0 +1,71 @@
...
+        with:
+          repository: oneapi-src/level-zero
+          path: level_zero
+  
+      - name: Build and Install Level Zero
+        run: |
+            cd level_zero
+            mkdir build
+            cd build
+            cmake .. -DCMAKE_INSTALL_PREFIX=/usr
```
lukaszstolarczuk, line 42: don't install in `/usr` if on self-hosted runner 😉 pls install locally, somewhere in workspace

.github/workflows/gpu.yml
```diff
@@ -0,0 +1,71 @@
...
+            sudo cmake --build . --config Release --target install
```
lukaszstolarczuk, line 43: sudo won't be needed

pbalcer:
> > 
> 
> Perhaps, but I think UR might prefer to have a static library that doesn't perform any dlopens - at least for the case where all dependencies are linked statically. @pbalcer is that true? or is there no difference for UR?

UR would require two variants of this provider: one that is fully dynamically linked that uses the exact same L0 symbols as the adapter (so not dlopen'ed), and one that uses static linking and then uses symbols that from the library that the adapter already dlopen'ed (so we'd probably need to dlopen the exact same library path).

bratpiorka:
> Let's discus it on our sync on Monday. I doubt if we need dlopen at all. maybe just dlsym is enough. My thoughts are the following:
> 
> * L0 memory provider is used only when client already uses L0.
> * So L0 should already be loaded into the process.
> * UMF should just find the corresponding  L0 symbols that are already loaded in the process somehow. It should be possible with `dlsym` and `RTLD_DEFAULT` (a special pseudo-handle)
> 
> With such schema, clients will use L0 in the same way like today (explicit linking, dlopen, etc.). UMF will not have dependency to `libze_loader`. L0 memory provider implementation will try to find corresponding symbols at runtime.

@vinser52 I made changes that follow this algorithm: first I try to get symbols using NULL handle and only if this fails I try to load libze_loader.so

bratpiorka:
@pbalcer I will resolve all UR-related issues in a separate PR when I also try to do a PR in UR repo that integrates with this provider

igchor:
> > > > 
> > > 
> > > 
> > > Perhaps, but I think UR might prefer to have a static library that doesn't perform any dlopens - at least for the case where all dependencies are linked statically. @pbalcer is that true? or is there no difference for UR?
> > 
> > 
> > UR would require two variants of this provider: one that is fully dynamically linked that uses the exact same L0 symbols as the adapter (so not dlopen'ed), and one that uses static linking and then uses symbols that from the library that the adapter already dlopen'ed (so we'd probably need to dlopen the exact same library path).
> 
> Let's discus it on our sync on Monday. I doubt if we need dlopen at all. maybe just dlsym is enough. My thoughts are the following:
> 
> * L0 memory provider is used only when client already uses L0.
> * So L0 should already be loaded into the process.
> * UMF should just find the corresponding  L0 symbols that are already loaded in the process somehow. It should be possible with `dlsym` and `RTLD_DEFAULT` (a special pseudo-handle)
> 
> With such schema, clients will use L0 in the same way like today (explicit linking, dlopen, etc.). UMF will not have dependency to `libze_loader`. L0 memory provider implementation will try to find corresponding symbols at runtime.

@vinser52 @pbalcer  But keep in mind that what Rafal implemented is using dlopen/dlsym on level zero **loader**. The loader itself will use dlopen to load appropriate drivers (from what I understand). So I don't think we need to worry about what symbols are being used, do we? Btw, assuming there would be a static version of L0 loader (could we ask L0 team?), could we just link with that and rely on L0 discovery instead of doing any dlopen/dlsym ourselves?

test/CMakeLists.txt
```diff
@@ -124,6 +125,12 @@ if(UMF_BUILD_OS_MEMORY_PROVIDER AND LINUX) # OS-specific functions are implement
                  LIBS umf_utils numa)
 endif()
 
+if(UMF_BUILD_LIBUMF_PROVIDER_LEVEL_ZERO AND LINUX) # TODO
```
igchor, line 128: Why TODO? Is there anything missing to support it on windows?

test/providers/provider_level_zero.cpp
```diff
@@ -0,0 +1,210 @@
...
+    umf_memory_provider_handle_t provider = nullptr;
+    umf_result_t umf_result = umfMemoryProviderCreate(
+        umfLevelZeroMemoryProviderOps(), &params, &provider);
+    ASSERT_EQ(umf_result, UMF_RESULT_SUCCESS);
+    ASSERT_NE(provider, nullptr);
+
+    const char *message;
+    int32_t error = 0;
+    umfMemoryProviderGetLastNativeError(provider, &message, &error);
+    ASSERT_EQ(error, 0);
```
igchor, line 178: I don't think you should test for that. The return value of umfMemoryProviderGetLastNativeError is unspecified if the previous call was successful.

test/providers/provider_level_zero.cpp
```diff
@@ -0,0 +1,210 @@
...
+
+        // Create context
+        ze_context_desc_t ctxtDesc = {ZE_STRUCTURE_TYPE_CONTEXT_DESC, nullptr,
+                                      0};
+        ze_result = zeContextCreate(hDriver, &ctxtDesc, &hContext);
+        ASSERT_EQ(ze_result, ZE_RESULT_SUCCESS);
+        ASSERT_NE(hContext, nullptr);
+    }
+
+    void TearDown() override { test::TearDown(); }
```
igchor, line 76: shouldn't you destroy the context, etc. here?

README.md
```diff
@@ -31,6 +31,10 @@ For development and contributions:
 For building tests, multithreaded benchmarks and Disjoint Pool:
 - C++ compiler with C++17 support
 
+For Level Zero memory provider and its tests:
+- Level Zero headers and libraries
```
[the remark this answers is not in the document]
bratpiorka: I've added this requirement one line down - if you have a better idea how to write it, let me know :)

src/libumf.map
bratpiorka: added

src/CMakeLists.txt
bratpiorka: Yes, but at the moment we don't have the internal infrastructure to test it

test/providers/provider_level_zero.cpp
bratpiorka: thanks, added

test/providers/provider_level_zero.cpp
bratpiorka: removed

test/CMakeLists.txt
bratpiorka: At the moment we don't have the internal infrastructure to test it. I will try to include Windows ASAP in separate PRs

src/provider/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
+# Copyright (C) 2024 Intel Corporation
+# Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# libumf_provider_level_zero
+if(UMF_BUILD_LIBUMF_PROVIDER_LEVEL_ZERO)
```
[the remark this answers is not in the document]
bratpiorka: Yes, but the Level Zero memory provider is a separate library, while the OS memory provider is not (there is only an option to enable/disable it by adding its sources to libumf). If we decide to make it a separate, static library (linked by default), it should be done in a separate PR.

igchor (approved):
I'd still like to discuss the linking (whether and how we use dlopen) on our sync but the rest looks good.

bratpiorka:
@igchor @vinser52 @pbalcer - please review. I removed dlopen() from the Level Zero provider at all. The test I added to the CI loads libze_loader dynamically using dlopen(RTLD_GLOBAL) and the Level Zero provider reuses these symbols without any problems.

src/libumf.map
```diff
@@ -35,6 +35,7 @@ UMF_1.0 {
         umfMemoryProviderCreateFromMemspace;
         umfMemspaceCreateFromNumaArray;
         umfMemspaceDestroy;
+        umfLevelZeroMemoryProviderOps;
```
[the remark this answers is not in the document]
bratpiorka: done

.github/workflows/gpu.yml
```diff
@@ -0,0 +1,47 @@
...
+          -DUMF_BUILD_LIBUMF_PROVIDER_LEVEL_ZERO=ON
+          -DUMF_BUILD_LIBUMF_POOL_DISJOINT=ON
+          -DUMF_ENABLE_POOL_TRACKING=ON
+
+      - name: Build UMF
+        run: cmake --build ${{github.workspace}}/build -j $(nproc)
+
+      - name: Run tests
+        working-directory: ${{github.workspace}}/build
+        run: |
```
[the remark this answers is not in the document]
bratpiorka: done

include/umf/providers/provider_level_zero.h
```diff
@@ -0,0 +1,34 @@
...
+ * Copyright (C) 2024 Intel Corporation
+ *
+ * Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+*/
+
+#ifndef UMF_PROVIDER_LEVEL_ZERO_H
+#define UMF_PROVIDER_LEVEL_ZERO_H
+
+#include <level_zero/ze_api.h>
```
igchor, line 11: I don't think we can include this header here. Our customers might not have the headers installed, if they use dlopen, right? @vinser52?

igchor (changes requested):

igchor:
> @igchor @vinser52 @pbalcer - please review. I removed dlopen() from the Level Zero provider at all. The test I added to the CI loads libze_loader dynamically using dlopen(RTLD_GLOBAL) and the Level Zero provider reuses these symbols without any problems.

Can we also have a similar test for when the application links with Level Zero normally (not using dlopen)? I.e. just have a second test that just assign all those functions pointers in libze_ops to &zeInit, &zeDriverGet, etc?

include/umf/providers/provider_level_zero.h
bratpiorka: but how they could pass ze_context_handle_t or ze_device_handle_t to the level zero provider params?

lukaszstolarczuk (commented):
👍

bratpiorka:
> Can we also have a similar test for when the application links with Level Zero normally (not using dlopen)? I.e. just have a second test that just assign all those functions pointers in libze_ops to &zeInit, &zeDriverGet, etc?

@igchor done

include/umf/providers/provider_level_zero.h
igchor: They might have a copy of the headers somewhere, but not necessarily in <level_zero/ze_api.h> I think.

bratpiorka: or I can check the #define and include the file only when the user hadn't done it yet + print the messafe

igchor: We can also just accept 'void*' for each param - I don't think there is a good solution for that if we want to use dlopen.

bratpiorka: Ok I removed this header from the interface

igchor (approved):

CMakeLists.txt
```diff
@@ -8,6 +8,7 @@ project(unified-memory-framework VERSION 0.1.0 LANGUAGES C)
 # Build Options
 option(UMF_BUILD_SHARED_LIBRARY "Build UMF as shared library" OFF)
 option(UMF_BUILD_OS_MEMORY_PROVIDER "Build OS memory provider" ON)
+option(UMF_BUILD_LIBUMF_PROVIDER_LEVEL_ZERO "Build Level Zero memory provider" OFF)
```
vinser52, line 11: if L0 provider implementation has no build-time dependencies why not ON it by default?

src/provider/provider_level_zero.c
```diff
@@ -0,0 +1,240 @@
...
+    ze_provider->memory_type = ze_params->memory_type;
+    ze_provider->host_mem_alloc_desc = (ze_host_mem_alloc_desc_t){
+        ZE_STRUCTURE_TYPE_HOST_MEM_ALLOC_DESC,
+        NULL, // pointer to an extension-specific structure
+        ze_params->level_zero_host_mem_alloc_flags};
+    ze_provider->device_mem_alloc_desc = (ze_device_mem_alloc_desc_t){
+        ZE_STRUCTURE_TYPE_DEVICE_MEM_ALLOC_DESC,
+        NULL, // pointer to an extension-specific structure
+        ze_params->level_zero_device_mem_alloc_flags,
+        ze_params->level_zero_device_local_mem_ordinal};
```
vinser52, line 93: I do not understand where it is used and why it is needed?

src/provider/provider_level_zero.c
```diff
@@ -0,0 +1,240 @@
...
+}
+
+static enum umf_result_t ze_memory_provider_alloc(void *provider, size_t size,
+                                                  size_t alignment,
+                                                  void **resultPtr) {
+    assert(provider);
+    assert(resultPtr);
+
+    ze_memory_provider_t *ze_provider = (ze_memory_provider_t *)provider;
+    ze_result_t ze_result = g_ze_ops.zeMemAllocDevice(
```
vinser52, line 114: Based on `ze_provider->memory_type` we should use `zeMemAllocDevice`, `zeMemAllocHost` or `zeMemAllocShared`.

src/provider/provider_level_zero.c
```diff
@@ -0,0 +1,240 @@
...
+}
+
+static enum umf_result_t
+ze_memory_provider_get_min_page_size(void *provider, void *ptr,
+                                     size_t *pageSize) {
+    (void)provider;
+    (void)ptr;
+
+    // TODO
+    *pageSize = 1024;
```
vinser52, line 156: as I know it is 64K

src/provider/provider_level_zero.c
```diff
@@ -0,0 +1,240 @@
...
+typedef struct ze_memory_provider_t {
+    ze_context_handle_t context;
+    ze_device_handle_t device;
+    ze_memory_type_t memory_type;
+    ze_host_mem_alloc_desc_t host_mem_alloc_desc;
+    ze_device_mem_alloc_desc_t device_mem_alloc_desc;
+} ze_memory_provider_t;
+
+typedef struct ze_ops_t {
+    ze_result_t (*zeMemAllocDevice)(ze_context_handle_t,
```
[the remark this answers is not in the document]
bratpiorka: I planned to add shared and host in the next PR which should contain all modes required by UR

CMakeLists.txt
bratpiorka: no, ze_api.h is still used in src/provider/provider_level_zero.c

README.md
```diff
@@ -123,6 +128,18 @@ A memory provider that provides memory from an operating system.
...
+
+A memory provider that provides memory from a graphics device.
+
+##### Requirements
+
+1) System with Level Zero compatible GPU
+2) Linux OS
+3) The `UMF_BUILD_LIBUMF_PROVIDER_LEVEL_ZERO` option turned `ON`
+4) Required packages:
+   - TODO
```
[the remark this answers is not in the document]
bratpiorka: is this a package? I rather thought of libze1 but I'm not sure

src/provider/CMakeLists.txt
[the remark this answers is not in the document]
bratpiorka: Right now I use ze_api.h header but do you suggest removing it completely? I'm not sure if I should re-define Level Zero structures inside UMF

src/provider/provider_level_zero.c
bratpiorka: you have to pass ze_device_mem_alloc_desc_t struct as an argument to zeMemAllocDevice

README.md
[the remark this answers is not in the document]
vinser52: Ok, I think I got the point. To build L0 provider we need `level-zero-dev` (on Ubuntu, maybe on RedHat it named differently) package. But for distribution, we do not need such dependency, right?

CMakeLists.txt
vinser52: Ok, there is a build-time dependency. Another question regarding distribution, since L0 provider implementation has no runtime dependency to L0 driver we can make L0 provider implementation to be part of core UMF library.

src/provider/CMakeLists.txt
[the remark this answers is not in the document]
vinser52: but how it relates to the fact that L0 provider is a separate static library?

src/provider/provider_level_zero.c
vinser52: right, but we can create it on stack inside `ze_memory_provider_alloc` function. Furthermore, `device_mem_alloc_desc` is needed only when `ze_provider->memory_type` equals to the device memory - it is another argument not to make it part of the `ze_params` object.

CMakeLists.txt
bratpiorka: I will make this option enabled by default (ON) + check for build deps in cmake + message to user

README.md
```diff
@@ -123,6 +128,18 @@ A memory provider that provides memory from an operating system.
 4) Required packages for tests:
    - libnuma-dev
 
+#### Level Zero memory provider (Linux-only yet)
+
+A memory provider that provides memory from a graphics device.
```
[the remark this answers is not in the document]
bratpiorka: done

README.md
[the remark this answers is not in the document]
bratpiorka: done

src/provider/provider_level_zero.c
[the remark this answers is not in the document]
bratpiorka: added

bratpiorka:
@vinser52 I have resolved the discussions started by you, as we agreed

src/provider/provider_level_zero.c
bratpiorka: done

.github/workflows/pr_push.yml
```diff
@@ -113,3 +115,6 @@ jobs:
   ProxyLib:
     needs: [Build]
     uses: ./.github/workflows/proxy_lib.yml
+  GPU:
+    needs: [Build]
+    uses: ./.github/workflows/gpu.yml
```
ldorau, line 120: We may add badges for new workflows later ...

ldorau (changes requested):
It does not build on WSL Ubuntu:
```
/home/ldorau/work/unified-memory-framework/src/provider/provider_level_zero.c: In function ‘init_ze_global_state’:
/home/ldorau/work/unified-memory-framework/src/provider/provider_level_zero.c:48:48: error: ‘RTLD_DEFAULT’ undeclared (first use in this function)
   48 |     *(void **)&g_ze_ops.zeMemAllocHost = dlsym(RTLD_DEFAULT, "zeMemAllocHost");
      |                                                ^~~~~~~~~~~~
/home/ldorau/work/unified-memory-framework/src/provider/provider_level_zero.c:48:48: note: each undeclared identifier is reported only once for each function it appears in
make[2]: *** [src/CMakeFiles/umf.dir/build.make:328: src/CMakeFiles/umf.dir/provider/provider_level_zero.c.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:1153: src/CMakeFiles/umf.dir/all] Error 2
make: *** [Makefile:146: all] Error 2
```

.github/workflows/pr_push.yml
bratpiorka: added

bratpiorka:
@ldorau please check now. BTW we definitely should add WSL to our CI

ldorau (approved):

vinser52 (approved):

PR #238, opened 2024-02-14
Make sure TRACKER is not used after being destroyed

ldorau:
Make sure TRACKER is not used after being destroyed.
Also add checks to umfMemoryTrackerGetPool()
to detect if TRACKER or its map is not initialized.

### DIFF
--- a/src/provider/provider_tracking.c
+++ b/src/provider/provider_tracking.c
@@ -81,6 +81,16 @@ static umf_result_t umfMemoryTrackerRemove(umf_memory_tracker_handle_t hTracker,
 umf_memory_pool_handle_t umfMemoryTrackerGetPool(const void *ptr) {
     assert(ptr);
 
+    if (TRACKER == NULL) {
+        fprintf(stderr, "tracker is not created\n");
+        return NULL;
+    }
+
+    if (TRACKER->map == NULL) {
+        fprintf(stderr, "tracker's map is not created\n");
+        return NULL;
+    }
+
     uintptr_t rkey;
     tracker_value_t *rvalue;
     int found = critnib_find(TRACKER->map, (uintptr_t)ptr, FIND_LE,
--- a/src/libumf_linux.c
+++ b/src/libumf_linux.c
@@ -19,7 +19,10 @@ void __attribute__((constructor)) umfCreate(void) {
 }
 
 void __attribute__((destructor)) umfDestroy(void) {
-    umfMemoryTrackerDestroy(TRACKER);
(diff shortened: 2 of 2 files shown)


igchor:
Can you add some more info what problem is this solving?

ldorau:
> Can you add some more info what problem is this solving?

Added a comment and commit description.

bratpiorka (approved):

lukaszstolarczuk (approved):

src/provider/provider_tracking.c
```diff
@@ -80,6 +80,8 @@ static umf_result_t umfMemoryTrackerRemove(umf_memory_tracker_handle_t hTracker,
 
 umf_memory_pool_handle_t umfMemoryTrackerGetPool(const void *ptr) {
     assert(ptr);
+    assert(TRACKER);
```
igchor, line 83: I don;t know if we should assert it. Creating memory tracker can fail for a few reasons, see here: https://github.com/oneapi-src/unified-memory-framework/blob/f2b784afcc54b360e46e0a098c7f3b2641b31625/src/provider/provider_tracking.c#L464. I think we should rather just log this.

ldorau: Done

ldorau: Replaced with:
```c
    if (TRACKER == NULL) {
        fprintf(stderr, "tracker is not created\n");
        return NULL;
    }

    if (TRACKER->map == NULL) {
        fprintf(stderr, "tracker's map is not created\n");
        return NULL;
    }
```

Done

src/provider/provider_tracking.c
```diff
@@ -81,6 +81,16 @@ static umf_result_t umfMemoryTrackerRemove(umf_memory_tracker_handle_t hTracker,
 umf_memory_pool_handle_t umfMemoryTrackerGetPool(const void *ptr) {
     assert(ptr);
 
+    if (TRACKER == NULL) {
+        fprintf(stderr, "tracker is not created\n");
+        return NULL;
+    }
+
+    if (TRACKER->map == NULL) {
```
vinser52, line 89: When this condition is possible?

ldorau: The same like in https://github.com/oneapi-src/unified-memory-framework/pull/238#discussion_r1491558174.
- when `critnib_new()` fails because `umf_ba_linear_create()` fails

vinser52 (approved):

igchor (approved):

PR #241, opened 2024-02-19
Add OS provider tests for different bind modes

DamianDuy:


test/provider_os_memory_multiple_numa_nodes.cpp
```diff
@@ -111,3 +136,102 @@ TEST_P(testNumaNodes, checkNumaNodesAllocations) {
...
+        UMF_OS_MEMORY_PROVIDER_PARAMS_TEST;
+    os_memory_provider_params.maxnode = numa_node_number + 1;
+    numa_bitmask_setbit(nodemask, numa_node_number);
+    os_memory_provider_params.nodemask = nodemask->maskp;
+    os_memory_provider_params.numa_mode = UMF_NUMA_MODE_PREFERRED;
+    initOsProvider(os_memory_provider_params);
+
+    umf_result_t umf_result;
+    umf_result =
+        umfMemoryProviderAlloc(os_memory_provider, alloc_size, 0, &ptr);
```
bratpiorka, line 154: you doing here only single small allocation, but please note that each of this modes differs in OOM scenario - you should test for that

test/provider_os_memory_multiple_numa_nodes.cpp
```diff
@@ -111,3 +136,102 @@ TEST_P(testNumaNodes, checkNumaNodesAllocations) {
...
+        UMF_OS_MEMORY_PROVIDER_PARAMS_TEST;
+    os_memory_provider_params.maxnode = numa_max_node() + 1;
+    set_all_available_nodemask_bits();
+    os_memory_provider_params.nodemask = nodemask->maskp;
+    os_memory_provider_params.numa_mode = UMF_NUMA_MODE_INTERLEAVE;
+    initOsProvider(os_memory_provider_params);
+
+    umf_result_t umf_result;
+    umf_result =
+        umfMemoryProviderAlloc(os_memory_provider, alloc_size, 0, &ptr);
```
bratpiorka, line 228: please make a large allocation here and test where each page will be placed

bratpiorka (changes requested):

test/provider_os_memory_multiple_numa_nodes.cpp
```diff
@@ -60,6 +61,30 @@ struct testNumaNodes : public testing::TestWithParam<int> {
         return numa_node;
     }
 
+    long unsigned int retrieve_nodemask(void *addr) {
+        struct bitmask *retrieved_nodemask = numa_allocate_nodemask();
+        EXPECT_NE(nodemask, nullptr);
```
kswiecicki, line 66: Use `UT_ASSERT...` macros for assertions.

igchor:
I don't know if this PR is supposed to cover entire https://github.com/oneapi-src/unified-memory-framework/issues/242 so feel free to address my comment in another PR:

We should check that policies like 'preferred' actually fall back to other nodes in case there is not enough memory on the specified node. Also, for modes that should fail if there is not enough memory we should test they actually do fail. One way to do that would be to add some kind of env variable which, if set, we could use to specify NUMA node on which we want to allocate (and use it for all binding tests). Then, we could run those tests in an environment with some custom NUMA nodes (e.g. we could create a very small NUMA node on pmem to make sure the oom is triggered quickly)

test/provider_os_memory_multiple_numa_nodes.cpp
```diff
@@ -111,3 +136,102 @@ TEST_P(testNumaNodes, checkNumaNodesAllocations) {
...
+    // This pointer must point to an initialized value before retrieving a number of
+    // the numa node that the pointer was allocated on (calling get_mempolicy).
+    memset(ptr, 0xFF, alloc_size);
+    int retrieved_numa_node_number = retrieve_numa_node_number(ptr);
+    ASSERT_EQ(retrieved_numa_node_number, numa_node_number);
+}
+
+// Test for allocations on numa nodes with interleave mode enabled.
+// The page allocations are interleaved across the set of nodes specified in nodemask.
+TEST_P(testNumaNodes, checkModeInterleave) {
```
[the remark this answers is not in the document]
DamianDuy: Done. It's not parametrized anymore.

test/provider_os_memory_multiple_numa_nodes.cpp
```diff
@@ -111,3 +136,102 @@ TEST_P(testNumaNodes, checkNumaNodesAllocations) {
...
+    umf_os_memory_provider_params_t os_memory_provider_params =
+        UMF_OS_MEMORY_PROVIDER_PARAMS_TEST;
+    os_memory_provider_params.maxnode = numa_max_node() + 1;
```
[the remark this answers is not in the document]
DamianDuy: Yeah, I wasn't sure what to pass there but `numa_max_node` makes more sense, I think. Done.

test/provider_os_memory_multiple_numa_nodes.cpp
```diff
@@ -60,6 +61,30 @@ struct testNumaNodes : public testing::TestWithParam<int> {
         return numa_node;
     }
 
+    long unsigned int retrieve_nodemask(void *addr) {
+        struct bitmask *retrieved_nodemask = numa_allocate_nodemask();
+        EXPECT_NE(nodemask, nullptr);
+        int ret = get_mempolicy(nullptr, retrieved_nodemask->maskp,
+                                numa_max_possible_node(), addr, MPOL_F_ADDR);
+        EXPECT_EQ(ret, 0);
+        long unsigned int retrieved_nodemask_value =
```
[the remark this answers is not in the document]
DamianDuy: Done.

test/provider_os_memory_multiple_numa_nodes.cpp
```diff
@@ -111,3 +136,102 @@ TEST_P(testNumaNodes, checkNumaNodesAllocations) {
...
+    // This pointer must point to an initialized value before retrieving a number of
+    // the numa node that the pointer was allocated on (calling get_mempolicy).
+    memset(ptr, 0xFF, alloc_size);
+    int retrieved_numa_node_number = retrieve_numa_node_number(ptr);
+    ASSERT_EQ(retrieved_numa_node_number, numa_node_number);
+}
+
+// Test for allocation on numa node with default mode enabled.
+// Since no policy is set by the set_mempolicy function, it should
+// default to The system-wide default policy, which allocates pages
```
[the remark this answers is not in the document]
DamianDuy: Done.

test/provider_os_memory_multiple_numa_nodes.cpp
DamianDuy: I was thinking of adding all the OOM tests in the next PR.

test/provider_os_memory_multiple_numa_nodes.cpp
DamianDuy: Done.

test/provider_os_memory_multiple_numa_nodes.cpp
DamianDuy: Done.

test/provider_os_memory_multiple_numa_nodes.cpp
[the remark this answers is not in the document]
bratpiorka: done? where?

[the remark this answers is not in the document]
DamianDuy: Sorry, I misunderstood. After testing it, I think long unsigned int won't be enough so I've changed this function to return bitmask itself and then used numa library function `numa_bitmask_equal` to compare the bitmasks.

bratpiorka (approved):

bratpiorka:
@igchor @PatKamin @kswiecicki please review. I think the rest tests required for #242 would be added in separate PRs

test/provider_os_memory_multiple_numa_nodes.cpp
```diff
@@ -111,3 +134,234 @@ TEST_P(testNumaNodes, checkNumaNodesAllocations) {
...
+    os_memory_provider_params.numa_mode = UMF_NUMA_MODE_PREFERRED;
+    initOsProvider(os_memory_provider_params);
+
+    umf_result_t umf_result;
+    umf_result =
+        umfMemoryProviderAlloc(os_memory_provider, alloc_size, 0, &ptr);
+    ASSERT_EQ(umf_result, UMF_RESULT_SUCCESS);
+    ASSERT_NE(ptr, nullptr);
+
+    int cpu = sched_getcpu();
```
PatKamin, line 178: I think that this test shouldn't be run on a randomized cpu. We don't have a guarantee that this test will be run on different numa nodes.

Maybe you could try parameterizing this test with all available cpus on which the process can be ran and using `sched_setaffinity()` or sth similar to run this test on each of available cpus. This way this test will cover all possible numa nodes with cpus in a multi-numa-node server.

test/provider_os_memory_multiple_numa_nodes.cpp
```diff
@@ -111,3 +134,234 @@ TEST_P(testNumaNodes, checkNumaNodesAllocations) {
...
+    // the numa node that the pointer was allocated on (calling get_mempolicy).
+    memset(ptr, 0xFF, alloc_size);
+    int retrieved_numa_node_number = retrieve_numa_node_number(ptr);
+    ASSERT_EQ(retrieved_numa_node_number, numa_node_number);
+}
+
+// Test for allocations on numa nodes with interleave mode enabled.
+// The page allocations are interleaved across the set of nodes specified in nodemask.
+TEST_F(testNuma, checkModeInterleave) {
+    int num_page = 1024;
```
PatKamin, line 265: nit: Make it a constexpr. Also, `pages_num` (a plural form) or sth like that would better describe what this var is for.

test/provider_os_memory_multiple_numa_nodes.cpp
```diff
@@ -28,7 +30,17 @@ std::vector<int> get_available_numa_nodes_numbers() {
     return available_numa_nodes_numbers;
```
PatKamin, line 30: This test is all about multiple numa nodes. I think there should be a check if the size of available numa nodes vector is greater than 1. Otherwise, we might think that we test umf on multiple numa nodes but actually testing just a single numa node.

test/provider_os_memory_multiple_numa_nodes.cpp
```diff
@@ -82,14 +103,16 @@ struct testNumaNodes : public testing::TestWithParam<int> {
...
 INSTANTIATE_TEST_SUITE_P(
-    testNumaNodesAllocations, testNumaNodes,
+    testNumaNodesAllocations, testNumaOnAllNodes,
     ::testing::ValuesIn(get_available_numa_nodes_numbers()));
 
 // Test for allocations on numa nodes. This test will be executed for all numa nodes
 // available on the system. The available nodes are returned in vector from the
 // get_available_numa_nodes_numbers() function and passed to test as parameters.
-TEST_P(testNumaNodes, checkNumaNodesAllocations) {
+TEST_P(testNumaOnAllNodes, checkNumaNodesAllocations) {
```
igchor, line 115: I think it would be good to have a test case for BIND where you set multiple NUMA nodes in node mask. This is allowed according to the documentation.

test/provider_os_memory_multiple_numa_nodes.cpp
```diff
@@ -111,3 +134,234 @@ TEST_P(testNumaNodes, checkNumaNodesAllocations) {
...
+    // This pointer must point to an initialized value before retrieving a number of
+    // the numa node that the pointer was allocated on (calling get_mempolicy).
+    memset(ptr, 0xFF, alloc_size);
+    int retrieved_numa_node_number = retrieve_numa_node_number(ptr);
+    ASSERT_EQ(retrieved_numa_node_number, numa_node_number);
+}
+
+// Test for allocations on numa nodes with interleave mode enabled.
+// The page allocations are interleaved across the set of nodes specified in nodemask.
+TEST_F(testNuma, checkModeInterleave) {
```
igchor, line 264: Please also add a test where nodemask does not contain all NUMA nodes (e.g. a test with just one node).

lukaszstolarczuk:
@PatKamin, @igchor, can you please review your comments here and move them (if valid) to the new PR (#415). I didn't do any changes in tests, just rebased it.

PR #245, opened 2024-02-20
Expand HOST ALL memspace tests

kswiecicki:
~~Contains https://github.com/oneapi-src/unified-memory-framework/pull/205.~~


igchor:
@kswiecicki could you move the files in a separate commit and then modify them? It's hard to see what changed without the diff.

kswiecicki:
> @kswiecicki could you move the files in a separate commit and then modify them? It's hard to see what changed without the diff.

Done.

igchor (approved):

test/memspaces/memspace_helpers.hpp
```diff
@@ -0,0 +1,111 @@
...
+    void SetUp() override {
+        ::umf_test::test::SetUp();
+
+        if (numa_available() == -1 || numa_all_nodes_ptr == nullptr) {
+            GTEST_SKIP() << "Failed to initialize libnuma";
+        }
+
+        int maxNode = numa_max_node();
+        if (maxNode < 0) {
+            GTEST_SKIP() << "No available numa nodes";
```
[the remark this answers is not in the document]
kswiecicki: I think this should be a skip, or should we expect every platform that runs those test to have visible numa nodes?

test/memspaces/memspace_helpers.hpp
```diff
@@ -0,0 +1,111 @@
...
+    }
+
+    std::vector<size_t> nodeIds;
+};
+
+struct memspaceNumaTest : ::numaNodesTest {
+    void SetUp() override {
+        ::numaNodesTest::SetUp();
+
+        if (nodeIds.size()) {
```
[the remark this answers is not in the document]
kswiecicki: Not really, and since this test case inherits from `numaNodesTest` which performs the necessary checks, I've removed this `if`.

test/memspaces/memspace_helpers.hpp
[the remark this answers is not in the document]
lukaszstolarczuk: @bratpiorka ? 😉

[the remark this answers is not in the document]
bratpiorka: no NUMA nodes == no DRAM? Please fail the test in this case, as this should never fail in our CI. Please note that tests needn't be super-generic - they should only help us to increase our confidence that the UMF is working as expected most important scenarios.

test/memspaces/memspace_host_all.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+TEST_F(numaNodesTest, memspaceGet) {
+    umf_memspace_handle_t hMemspace = umfMemspaceHostAllGet();
+    UT_ASSERTne(hMemspace, nullptr);
+
+    // Confirm that the HOST ALL memspace is composed of all available NUMA nodes.
+    UT_ASSERTeq(hMemspace->size, nodeIds.size());
+    for (size_t i = 0; i < hMemspace->size; i++) {
+        // NUMA memory target internally casts the config directly into priv.
+        struct umf_numa_memory_target_config_t *numaTargetCfg =
```
bratpiorka, line 24: please add a TODO here: "use the memory target API when it would be available"

test/memspaces/memspace_host_all.cpp
```diff
@@ -0,0 +1,91 @@
...
+        umfMemoryProviderAlloc(hProvider, size, alignment, &ptr);
+    UT_ASSERTeq(umf_ret, UMF_RESULT_SUCCESS);
+    UT_ASSERTne(ptr, nullptr);
+
+    auto it = std::max_element(nodeIds.begin(), nodeIds.end());
+    UT_ASSERT(it != nodeIds.end());
+
+    unsigned long maxNodeId = *it;
+
+    unsigned bitsPerUlong = sizeof(unsigned long) * 8;
```
bratpiorka, line 72: please add some comments here that would explain why we expect BIND mode etc

test/memspaces/memspace_numa.cpp
```diff
@@ -0,0 +1,67 @@
...
+    umf_memspace_handle_t hMemspace = nullptr;
+    enum umf_result_t ret = umfMemspaceCreateFromNumaArray(
+        nodeIds.data(), nodeIds.size(), &hMemspace);
+    ASSERT_EQ(ret, UMF_RESULT_SUCCESS);
+    ASSERT_NE(hMemspace, nullptr);
+
+    umfMemspaceDestroy(hMemspace);
+}
+
+TEST_F(numaNodesTest, createNullArray) {
```
bratpiorka, line 22: please rename negative tests eg.:
createNullArray -> create_INVALID_ARRAY
createZeroSize -> create_INVALID_SIZE
etc

please see how it is done in provider_os_memory.cpp

test/memspaces/memspace_host_all.cpp
```diff
@@ -0,0 +1,91 @@
...
+    for (auto &nodeId : nodeIds) {
+        size_t memNodeMaskIdx = ((nodeId + bitsPerUlong) / bitsPerUlong) - 1;
+        auto &memNodeMask = memNodeMasks.at(memNodeMaskIdx);
+
+        UT_ASSERT(memNodeMask && (1UL << (nodeId % bitsPerUlong)));
+    }
+
+    umf_ret = umfMemoryProviderFree(hProvider, ptr, size);
+    UT_ASSERTeq(umf_ret, UMF_RESULT_SUCCESS);
+}
```
bratpiorka, line 91: what I miss here is the test that allocs memory until we got OOM and checks if all nodes are used

bratpiorka (changes requested):

test/memspaces/memspace_helpers.hpp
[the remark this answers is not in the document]
kswiecicki: Done.

test/memspaces/memspace_host_all.cpp
kswiecicki: Done.

test/memspaces/memspace_host_all.cpp
kswiecicki: Done.

test/memspaces/memspace_numa.cpp
kswiecicki: I see that we have `TEST_F` and `TEST_P` test cases that use camel case and snake case in its names. What case should we stick to in UMF?

lukaszstolarczuk: In general, I believe we should unify the naming convention. I can see most of the test names use camelCase, so, perhaps, use that one as well.

That said, I think Rafał meant to change words null/zero to `invalid`...? and probably add more cases, as in provider_os_memory test

test/memspaces/memspace_host_all.cpp
kswiecicki: Do we really need OOM test for such case? From my understanding, the provider created from memspace should always return memory that's bound to every node present on such memspace, and the `memoryPolicy` test case tests that. Would you like me to change this test to create multiple allocations?

I could create a test where memory is obtained until `UMF_OS_RESULT_ERROR_ALLOC_FAILED` OS provider native error (there's no clear OOM error) if it's necessary.

bratpiorka: yes, "should" :) yes, please add this case - simple loop + check if we allocate at least one allocation on each node would be enough (checking the total size of allcaction vs total size of NUMA nodes could be tricky because of swap file)

test/memspaces/memspace_numa.cpp
kswiecicki: I've marked the negative test cases with `Invalid` in the test name.

lukaszstolarczuk (approved):

bratpiorka (approved):

PR #246, opened 2024-02-21
Do not call ba_os_get_page_size() to often

ldorau:


src/base_alloc/base_alloc_linear.c
```diff
@@ -13,8 +13,11 @@
 #include "utils_common.h"
 #include "utils_concurrency.h"
 
+// will be initialized to ba_os_get_page_size() in umf_ba_linear_create()
+static __TLS size_t Page_size;
```
bratpiorka, line 17: why this is in TLS? I think it would be simpler to put this into umf_ba_main_linear_pool_meta_t

ldorau: Replaced with https://github.com/oneapi-src/unified-memory-framework/pull/249

ldorau:
Replaced with https://github.com/oneapi-src/unified-memory-framework/pull/249

PR #248, opened 2024-02-21
Implement and use umf_ba_linear_free() in the proxy library

ldorau:
Implement and use `umf_ba_linear_free()` in the proxy library.
`umf_ba_linear_free()` really frees memory only if all allocations from an inactive pool were freed.

### DIFF
--- a/src/base_alloc/base_alloc_linear.c
+++ b/src/base_alloc/base_alloc_linear.c
@@ -17,10 +17,12 @@
 #define DEBUG_RUN_CHECKS(pool) ba_debug_checks(pool)
 #define DEBUG_SET_VAR(var, value) DO_WHILE_EXPRS((var = value))
 #define DEBUG_INC_VAR(var) DO_WHILE_EXPRS((var++))
+#define DEBUG_DEC_VAR(var) DO_WHILE_EXPRS((var--))
 #else
 #define DEBUG_RUN_CHECKS(pool) DO_WHILE_EMPTY
 #define DEBUG_SET_VAR(var, value) DO_WHILE_EMPTY
 #define DEBUG_INC_VAR(var) DO_WHILE_EMPTY
+#define DEBUG_DEC_VAR(var) DO_WHILE_EMPTY
 #endif /* NDEBUG */
 
 // minimum size of a single pool of the linear base allocator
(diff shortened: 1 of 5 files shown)


test/test_base_alloc_linear.cpp
```diff
@@ -77,6 +81,10 @@ TEST_F(test, baseAllocLinearMultiThreadedAllocMemset) {
                 UT_ASSERTeq(*(buffer[i].ptr + k), (i + TID) & 0xFF);
             }
         }
+
+        for (int i = 0; i < ITERATIONS; i++) {
```
igchor, line 85: nit: it would be good to also have a test that frees the pointers in a different order (e.g random + from last to first).

src/base_alloc/base_alloc_linear.c
```diff
@@ -27,6 +27,8 @@ typedef struct umf_ba_main_linear_pool_meta_t {
     os_mutex_t lock;
     char *data_ptr;
     size_t size_left;
+    size_t global_n_allocs; // global number of allocations in all pools
```
igchor, line 30: shouldn't this be under debug only?

src/base_alloc/base_alloc_linear.c
```diff
@@ -160,6 +172,65 @@ void *umf_ba_linear_alloc(umf_ba_linear_pool_t *pool, size_t size) {
...
+// umf_ba_linear_free() really frees memory only if all allocations from an inactive pool were freed
+void umf_ba_linear_free(umf_ba_linear_pool_t *pool, void *ptr) {
+    util_mutex_lock(&pool->metadata.lock);
+#ifndef NDEBUG
+    ba_debug_checks(pool);
+#endif /* NDEBUG */
+    // check if ptr belongs to pool
+    char *cptr = (char *)ptr;
+    if (cptr >= pool->data &&
+        cptr < ((char *)(pool)) + pool->metadata.pool_size) {
```
igchor, line 184: nit: could you extract this check to a function like 'contains_ptr'?

ldorau: Done

test/test_base_alloc_linear.cpp
ldorau: Yes, it would be good. BTW random and real-life scenario is already tested by the proxy library.

ldorau: @bratpiorka I think it can be a separate task in Jira and we can give it someone else.

bratpiorka: done

bratpiorka (approved):

PR #249, opened 2024-02-21
Cache page size in ba_os_get_page_size()

ldorau:


src/base_alloc/base_alloc_linux.c
```diff
@@ -33,4 +33,12 @@ void ba_os_free(void *ptr, size_t size) {
     (void)ret; // unused
 }
 
-size_t ba_os_get_page_size(void) { return sysconf(_SC_PAGE_SIZE); }
+size_t ba_os_get_page_size(void) {
+    static size_t Page_size = 0;
+
+    if (!Page_size) {
```
igchor, line 39: Shouldn't we use util_init_once or something like that? This function can be called from multiple threads, right?

ldorau: Right. So it is better to make it a TLS variable - no MT synchronization is needed then.
Done.

pbalcer: using tls isn't free, especially for a library. For something this simple, util_init_once is the right thing to use...

ldorau: OK, done

bratpiorka (approved):

kswiecicki (approved):

igchor (approved):

PR #251, opened 2024-02-22
Build libumf_pool_scalable on Windows

ldorau:


src/pool/pool_scalable.c
```diff
@@ -62,11 +65,20 @@ static struct tbb_callbacks g_tbb_ops;
 static UTIL_ONCE_FLAG tbb_is_initialized = UTIL_ONCE_FLAG_INIT;
 static bool Init_tbb_global_state_failed;
 
+#ifdef _WIN32
```
vinser52, line 68: I would move this somewhere to utils and implement OS abstraction layer for such functionality.

bratpiorka: totally agree

bratpiorka (changes requested):

cmake/FindLIBTBB.cmake
```diff
@@ -0,0 +1,44 @@
+# Copyright (C) 2024 Intel Corporation
+# Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+message(STATUS "Checking for module 'libtbb' using find_library()")
```
igchor, line 5: Why do we need this file? Can't we just call `find_package(TBB components tbb)`?

examples/basic/basic.c
```diff
@@ -13,6 +13,8 @@
 #include <stdio.h>
 #include <string.h>
 
+#include "utils_common.h"
```
igchor, line 16: should we really use utils in examples? it will be harder to make them standalone

cmake/FindLIBTBB.cmake
ldorau: `find_package(TBB)` just calls the `FindLIBTBB.cmake` script

examples/basic/basic.c
ldorau: I can define `util_strncpy()`  inside `basic.c`

src/pool/pool_scalable.c
ldorau: @vinser52 is it OK now?

cmake/FindLIBTBB.cmake
igchor: Right, but I thought TBB already provides TBBconfig.cmake. Or do we use too old TBB version on our CI?

ldorau: Hmm, but to use `TBBconfig.cmake` we have to find the TBB package first, so to run `find_package(TBB)` we have to have it found already ... ;-)

igchor: Fom what I understand find_package will look for the config files in the system just like find_library that you're using. The difference is only in case of non-standard location I think - you need to set `CMAKE_MODULE_PATH` instead of `CMAKE_PREFIX_PATH`.

igchor: BTW, vcpkg docs recommend seeting CMAKE_TOOLCHAIN_FILE: https://github.com/microsoft/vcpkg?tab=readme-ov-file#using-vcpkg-with-cmake

igchor: When you install tbb using vcpkg it says:
```
tbb provides CMake targets:

  # this is heuristically generated, and may not be correct
  find_package(TBB CONFIG REQUIRED)
  target_link_libraries(main PRIVATE TBB::tbb TBB::tbbmalloc TBB::tbbmalloc_proxy)
```

I tested it locally and it works fine without findLIBTBB.cmake You just need to pass this CMAKE_TOOLCHAIN_FILE flag to cmake and change `find_package(LIBTBB REQUIRED tbb)` to `find_package(TBB CONFIG REQUIRED)` . Also, to get include dirs for TBB here: https://github.com/oneapi-src/unified-memory-framework/blob/c5785409841028022492f2c153b7689a7e920707/src/pool/CMakeLists.txt#L68 you need to do `get_target_property(LIBTBB_INCLUDE_DIRS TBB::tbbmalloc INTERFACE_INCLUDE_DIRECTORIES)`

igchor: Same is true for findLIBHWLOC.cmake I think, we can probably remove that too.

ldorau: OK, I will try.

.github/workflows/pr_push.yml
```diff
@@ -14,17 +14,17 @@ jobs:
...
     name: Fast build
     env:
-      HWLOC_PATH: "${{github.workspace}}/build/vcpkg/packages/hwloc_x64-windows"
+      VCPKG_PATH: "${{github.workspace}}/build/vcpkg/packages/hwloc_x64-windows;${{github.workspace}}/build/vcpkg/packages/tbb_x64-windows"
     strategy:
       matrix:
         include:
           - os: windows-latest
-            pool_scalable: 'OFF'
+            pool_scalable: 'ON'
```
bratpiorka, line 22: you enable pool_scalable in each matrix entry - could you just enable it in cmake options? This is a fast build so we don't need to disable it here

.github/workflows/codeql.yml
```diff
@@ -29,9 +29,9 @@ jobs:
           - os: ubuntu-latest
             pool_scalable: 'ON'
             # Windows doesn't recognize 'CMAKE_BUILD_TYPE', it uses '--config' param in build command
-            extra_build_option: ['-DCMAKE_BUILD_TYPE=Release']
+            extra_build_option: '-DCMAKE_BUILD_TYPE=Release'
           - os: windows-latest
-            pool_scalable: 'OFF'
+            pool_scalable: 'ON'
```
bratpiorka, line 34: just enable it in cmake options and remove this from matrix

bratpiorka (changes requested):

cmake/FindLIBTBB.cmake
ldorau: @igchor It does not work:
```
CMake Error at CMakeLists.txt:134 (find_package):
  Found package configuration file:

    C:/Users/ldorau/work/vcpkg/packages/tbb_x64-windows/share/tbb/TBBConfig.cmake

  but it set TBB_FOUND to FALSE so package "TBB" is considered to be NOT
  FOUND.
```

.github/workflows/basic.yml
```diff
@@ -178,7 +178,7 @@ jobs:
   windows-build:
     name: Windows
     env:
-      HWLOC_PATH: "${{github.workspace}}/build/vcpkg/packages/hwloc_x64-windows"
+      VCPKG_PATH: "${{github.workspace}}/build/vcpkg/packages/hwloc_x64-windows;${{github.workspace}}/build/vcpkg/packages/tbb_x64-windows"
```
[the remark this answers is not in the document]
ldorau: Done

.github/workflows/pr_push.yml
ldorau: Done

.github/workflows/codeql.yml
ldorau: Done

ldorau:
@bratpiorka Done

CMakeLists.txt
```diff
@@ -113,8 +113,33 @@ install(
...
+        find_package(TBB REQUIRED tbb)
+    endif()
+    message(STATUS ">>>>> TBB_LIBRARIES = ${TBB_LIBRARIES}")
+    message(STATUS ">>>>> TBB_LINK_LIBRARIES = ${TBB_LINK_LIBRARIES}")
+    message(STATUS ">>>>> TBB_LIBRARY_DIRS = ${TBB_LIBRARY_DIRS}")
+    message(STATUS ">>>>> TBB_LDFLAGS = ${TBB_LDFLAGS}")
+    message(STATUS ">>>>> TBB_LDFLAGS_OTHER = ${TBB_LDFLAGS_OTHER}")
+    message(STATUS ">>>>> TBB_INCLUDE_DIRS = ${TBB_INCLUDE_DIRS}")
+    message(STATUS ">>>>> TBB_CFLAGS = ${TBB_CFLAGS}")
+    message(STATUS ">>>>> TBB_CFLAGS_OTHER = ${TBB_CFLAGS_OTHER}")
```
KFilipek, line 142: Is it only for debugging purposes? I'd prefer to build it with a verbose param or something like that instead of leaving it here.

examples/basic/basic.c
```diff
@@ -45,9 +47,11 @@ int main(void) {
         goto memory_provider_destroy;
     }
 
+    const char *strSource = "Allocated memory at";
+
     // Write to the allocated memory
     memset(ptr_provider, '\0', alloc_size);
-    strcpy(ptr_provider, "Allocated memory at");
+    util_strncpy(ptr_provider, alloc_size, strSource, strlen(strSource) + 1);
```
bratpiorka, line 54: maybe you could use simple memcpy here? then basic.c needn't to use utils

bratpiorka (approved):

KFilipek (approved):

igchor (approved):

src/pool/pool_scalable.c
ldorau: Fixed.
Done.

PR #253, opened 2024-02-23
Fix computing ptr and update pool_size

ldorau:
The first (main) pool has just been freed except of the first page containing the metadata. The pool size has to be updated and set to the size of one page.

### DIFF
--- a/src/base_alloc/base_alloc_linear.c
+++ b/src/base_alloc/base_alloc_linear.c
@@ -199,9 +199,11 @@ int umf_ba_linear_free(umf_ba_linear_pool_t *pool, void *ptr) {
         if ((pool->metadata.pool_n_allocs == 0) && pool->next_pool &&
             (pool->metadata.pool_size > page_size)) {
             // we can free the first (main) pool except of the first page containing the metadata
-            void *ptr = pool + page_size;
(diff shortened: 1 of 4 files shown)


src/base_alloc/base_alloc_linear.c
```diff
@@ -202,6 +202,8 @@ int umf_ba_linear_free(umf_ba_linear_pool_t *pool, void *ptr) {
             void *ptr = pool + page_size;
             size_t size = pool->metadata.pool_size - page_size;
             ba_os_free(ptr, size);
+            // update pool_size
+            pool->metadata.pool_size = page_size;
```
bratpiorka, line 206: as this is a fix could you add test for that? Where did you find this?

ldorau: I'm afraid it is impossible to add a test for that, because `ba_os_free()` does not return a value. 
It is a very simple and obvious fix. I found it in my head thinking about the code under shower in the evening.
I just got a kind of "Eureka" ;-)

ldorau: @bratpiorka Done

kswiecicki (approved):

PR #257, opened 2024-02-23
Add searching for libnuma in CMake

ldorau:


cmake/FindLIBNUMA.cmake
```diff
@@ -0,0 +1,19 @@
+# Copyright (C) 2024 Intel Corporation
+# Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+message(STATUS "Checking for module 'libnuma' w/o PkgConfig")
+
+find_library(LIBNUMA_LIBRARY NAMES libnuma numa)
+set(LIBNUMA_LIBRARIES ${LIBNUMA_LIBRARY})
```
bratpiorka, line 8: do we need to introduce this new variable? LIBNUMA_LIBRARY isn't enough?

cmake/FindLIBNUMA.cmake
```diff
@@ -0,0 +1,19 @@
+# Copyright (C) 2024 Intel Corporation
+# Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+message(STATUS "Checking for module 'libnuma' w/o PkgConfig")
```
bratpiorka, line 5: "w/o PkgConfig" -> "using find_library()" sounds better IMO

.github/workflows/pr_push.yml
```diff
@@ -71,7 +71,7 @@ jobs:
       - name: Install apt packages
         run: |
           sudo apt-get update
-          sudo apt-get install -y cmake clang-format-15
+          sudo apt-get install -y cmake clang-format-15 libnuma-dev
```
lukaszstolarczuk, line 74: I don't think it's required in coding style - perhaps just disable building of the tests

ldorau: Done.

cmake/FindLIBNUMA.cmake
ldorau: So what do you suggest? It works as is now.

lukaszstolarczuk (approved):

bratpiorka (approved):

Issue #260, opened 2024-02-26
The umf-base_alloc_linear test fails sporadically on the assert on Windows


ldorau:
The umf-base_alloc_linear test fails sporadically on Windows on the following assert:
https://github.com/oneapi-src/unified-memory-framework/blob/bd135261817b5546ce4ad07931287a6e0b8de345/src/base_alloc/base_alloc_linear.c#L251

Failed build:
https://github.com/oneapi-src/unified-memory-framework/actions/runs/8020268125/job/21911573318?pr=255

## Environment Information

- UMF version (hash commit or a tag): bd135261817b5546ce4ad07931287a6e0b8de345
- OS(es) version(s): Windows

## Please provide a reproduction of the bug:

Run the umf-base_alloc_linear test.

## How often bug is revealed:

sporadically - often

PR #261, opened 2024-02-26
Skip temporarily the assert in umf_ba_linear_destroy()

ldorau:
This assert fails sporadically on Windows only.
It is safe to comment it out temporarily,
because the linear base allocator is used
in the proxy library only but `umf_ba_linear_destroy()`
is skipped, so this code is never run in real life.

Ref: #260


DamianDuy:
LGTM

bratpiorka (approved):

PR #263, opened 2024-02-26
Use global base_alloc in proxy lib

igchor:
Replace linear allocator with global base alloc. This PR also makes linear allocator obsolete and hence fixes: https://github.com/oneapi-src/unified-memory-framework/issues/260

After these changes, it is possible to destroy Proxy_pool and OS provider in proxy lib.

During debugging I found that there is still one or two allocations coming from dlopen that are not free, but they are bigger than any allocation class size that we use hence it doesn't trigger any asserts.


src/base_alloc/base_alloc_global.c
```diff
@@ -82,41 +83,114 @@ static int size_to_idx(size_t size) {
     return index;
 }
 
-void *umf_ba_global_alloc(size_t size) {
+static void *transform_ptr(void *ptr, size_t size, size_t alignment) {
+    assert(ptr);
+
+    void *user_ptr;
+    if (alignment <= sizeof(size_t)) {
+        user_ptr = (void *)((uintptr_t)ptr + sizeof(size_t));
```
ldorau, line 91: Add a comment why it is aligned to `alignment`, please.

src/proxy_lib/proxy_lib.c
```diff
@@ -36,38 +36,30 @@
...
+#include "base_alloc_global.h"
 #include "proxy_lib.h"
 #include "utils_common.h"
 #include "utils_concurrency.h"
 
 /*
  * The UMF proxy library uses two memory allocators:
- * 1) the "LEAK" internal linear base allocator based on the anonymous mapped
- *    memory that will NOT be destroyed (with API ba_leak_*()).
+ * 1) base_alloc for bootstrapping
```
ldorau, line 46: base allocator

src/proxy_lib/proxy_lib.c
```diff
@@ -97,80 +89,46 @@ void proxy_lib_create_common(void) {
...
-    // We cannot destroy 'Base_alloc_leak' nor 'Proxy_pool' nor 'OS_memory_provider',
-    // because it could lead to use-after-free in the program's unloader
-    // (for example _dl_fini() on Linux).
+    umfPoolDestroy(Proxy_pool);
+    Proxy_pool = NULL;
+
+    umfMemoryProviderDestroy(OS_memory_provider);
+    OS_memory_provider = NULL;
+
+    umf_ba_destroy_global();
```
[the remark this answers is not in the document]
ldorau: I mean `umf_ba_destroy_global();`

src/proxy_lib/proxy_lib.c
```diff
@@ -207,41 +169,32 @@ void *realloc(void *ptr, size_t size) {
         return NULL;
     }
 
-    size_t leak_pool_contains_pointer = ba_leak_pool_contains_pointer(ptr);
-    if (leak_pool_contains_pointer) {
-        return ba_leak_realloc(ptr, size, leak_pool_contains_pointer);
-    }
-
-    if (Proxy_pool) {
+    if (!was_called_from_umfPool && Proxy_pool) {
```
ldorau, line 171: **(Important)** What if an app wants to realloc a pointer from the base allocator when `Proxy_pool` already exists?
How can you distinguish which allocator a pointer comes from?

src/proxy_lib/proxy_lib.c
```diff
@@ -207,41 +169,32 @@ void *realloc(void *ptr, size_t size) {
...
     if (ptr == NULL) {
         return;
     }
 
-    if (ba_leak_free(ptr) == 0) {
-        return;
-    }
-
-    if (Proxy_pool) {
+    if (!was_called_from_umfPool && Proxy_pool) {
```
ldorau, line 186: **(Important)** What if an app wants to free a pointer from the base allocator when `Proxy_pool` already exists?
How can you distinguish which allocator a pointer comes from?

ldorau (changes requested):

ldorau (changes requested):
General comment: 
This PR merges 4 independent changes:
1) Fixing a problem with freeing memory in proxy lib (make sure base_alloc dtor is only called once by adding a macro to skip the dtor (it's set in proxy lib).)
2) Improvements in the global base allocator (adding metadata with stored size, `alloc_usable_size()` and removing size from `free()`)
3) Replacing the linear base allocator with the global base allocator in the proxy library
4) Removing the linear base allocator from the repo. 

IMHO this PR should be split into 4 separate smaller pull requests.

ldorau (changes requested):
```sh
$ LD_PRELOAD=./lib/libumf_proxy.so ctest --output-on-failure
```
ends with Segmentation fault:
```
Program received signal SIGSEGV, Segmentation fault.
0x00007ffff783dbf5 in __GI_getenv (name=0x7ffff6958edc "GNUTLS_NO_IMPLICIT_INIT") at ./stdlib/getenv.c:75
75      in ./stdlib/getenv.c
```

igchor:
> @igchor please split this change into multiple PRs

Ok, here's the first one: https://github.com/oneapi-src/unified-memory-framework/pull/272

PR #268, opened 2024-02-27
Enable OS memory provider on Windows

ldorau:


src/provider/provider_os_memory_windows.c
```diff
@@ -0,0 +1,64 @@
...
+    }
+    fprintf(stderr,
+            "os_translate_mem_protection_flags(): unsupported protection flag: "
+            "%u\n",
+            protection);
+    assert(0);
+    return -1;
+}
+
+void *os_mmap(void *hint_addr, size_t length, int prot) {
```
igchor, line 42: I think we'll need to modify the abstractions a bit (instead of os_mmap, have os_allocate that would do allocation and binding). As far as I understand there is no 'mbind' equivalent on windows. I only know of VirtualAllocExNuma which combines allocation and binding to a specific node (and even then, it's only a hint).

ldorau: Why? We use `hwloc_set_area_membind()`.

src/CMakeLists.txt
```diff
@@ -102,7 +100,15 @@ else()
...
 
-target_compile_definitions(umf PUBLIC ${UMF_COMPILE_DEFINITIONS})
+target_compile_definitions(umf PUBLIC ${UMF_PUBLIC_COMPILE_DEFINITIONS})
+
+if(UMF_BUILD_OS_MEMORY_PROVIDER AND WINDOWS)
+    # Disable the warning:
+    # hwloc-win64-build-2.10.0\include\hwloc/helper.h(1266):
+    # warning C4996: 'sscanf': This function or variable may be unsafe.
+    # Consider using sscanf_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS.
+    target_compile_definitions(umf PRIVATE _CRT_SECURE_NO_WARNINGS)
```
bratpiorka, line 116: I think this is too strong because it disables all warning for the whole lib - I would prefer to use 

#ifdef windows
#pragma warning( push )
#pragma warning( disable : 4101)
#include <hwloc.h>
#pragma warning( pop )

src/CMakeLists.txt
```diff
@@ -86,13 +84,19 @@ elseif(MACOSX)
 endif()
 
 if(UMF_BUILD_SHARED_LIBRARY)
+    if(WINDOWS AND (NOT UMF_BUILD_OS_MEMORY_PROVIDER))
+        # UMF_BUILD_SHARED_LIBRARY=ON requires UMF_BUILD_OS_MEMORY_PROVIDER=ON on Windows
```
bratpiorka, line 88: This should be fixed by making def and map files generated - we had a JIRA for that, but please insert here a TODO

src/provider/provider_os_memory_windows.c
bratpiorka: @igor please add more details about that - we could even sync. However, whatever potential issue we have here should be detected by our tests. Do you have any idea how we can improve them?

src/provider/provider_os_memory_windows.c
```diff
@@ -0,0 +1,64 @@
...
+int os_purge(void *addr, size_t length, int advice) {
+    fprintf(stderr, "os_purge() NOT IMPLEMENTED\n");
+    // assert(0);
+    return -1;
+}
+
+size_t os_get_page_size(void) { return 4096; } // TODO fix this
+
+void os_strerror(int errnum, char *buf, size_t buflen) {
+    fprintf(stderr, "os_strerror() NOT IMPLEMENTED\n");
```
bratpiorka, line 62: could we use strerror from string.h?

src/provider/provider_os_memory_windows.c
```diff
@@ -0,0 +1,64 @@
...
+
+int os_munmap(void *addr, size_t length) {
+    // If VirtualFree() succeeds, the return value is nonzero.
+    // If VirtualFree() fails, the return value is 0 (zero).
+    (void)length; // unused
+    return (VirtualFree(addr, 0, MEM_RELEASE) == 0);
+}
+
+int os_purge(void *addr, size_t length, int advice) {
+    fprintf(stderr, "os_purge() NOT IMPLEMENTED\n");
```
bratpiorka, line 54: I think this is what the MEM_RELEASE flag does

src/provider/provider_os_memory_windows.c
igchor: I'm not 100% sure (I haven't tested it) but from reading the code it seems that hwloc_set_area_membind is not supported on windows. Only hwloc_alloc_membind seems to be supported: https://github.com/open-mpi/hwloc/blob/39fae7e3151fbd677f953f6990acd9eb2a0b9bfb/hwloc/topology-windows.c#L1354

igchor: As for the tests, I think the existing OS provider tests should cover it - we just need to enable them.

.github/workflows/basic.yml
```diff
@@ -195,24 +197,30 @@ jobs:
             pool_tracking: 'ON'
             shared_library: 'ON'
             os_provider: 'ON'
-          - os: 'windows-2022'
```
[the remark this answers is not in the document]
ldorau: Because "UMF_BUILD_SHARED_LIBRARY=ON requires UMF_BUILD_OS_MEMORY_PROVIDER=ON on Windows"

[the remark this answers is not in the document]
ldorau: I will comment it out and add a comment and TODO here

src/CMakeLists.txt
ldorau: Changed: added a separate def file for that case.

.github/workflows/basic.yml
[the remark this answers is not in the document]
ldorau: Reverted: added a separate def file for that case.

src/CMakeLists.txt
ldorau: Right. Thanks!
Done.

src/provider/provider_os_memory_windows.c
ldorau: We should use `strerror_s()` - I just didn't get to this place yet ...

src/provider/provider_os_memory_windows.c
```diff
@@ -0,0 +1,64 @@
...
+    // assert(0);
+    return -1;
+}
+
+size_t os_get_page_size(void) { return 4096; } // TODO fix this
```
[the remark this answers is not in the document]
ldorau: Done.

src/provider/provider_os_memory_windows.c
ldorau: `MEM_RELEASE` does unmapping (=`munmap()`)

src/provider/provider_os_memory_windows.c
ldorau: The existing OS provider tests require libnuma, but there is no libnuma on Windows ... :-(

bratpiorka: right.. and even if we write a test for windows, we currently do not have a machine / VM with multiple NUMA nodes. Still, I think it is worth merging this PR to enable building more UMF components on Windows

ldorau:
The failure of the Windows build when `UMF_BUILD_SHARED_LIBRARY=ON` is probably caused by https://github.com/oneapi-src/unified-memory-framework/issues/274

src/provider/provider_os_memory_windows.c
igchor: Ok, in that case, can we add an issue to rewrite those tests (perhaps using hwloc)?

Also I think on Windows we will just end up here: https://github.com/oneapi-src/unified-memory-framework/blob/main/src/provider/provider_os_memory.c#L467 Can we add a issue to fix this (remove this check altogether or make it configurable) as well?

ldorau: I ignore this check and added TODO here.

.github/workflows/basic.yml
```diff
@@ -208,11 +210,28 @@ jobs:
     - name: Checkout
       uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
 
+    - name: Get CMake for lukka/run-vcpkg
+      uses: lukka/get-cmake@latest
```
lukaszstolarczuk, line 214: please use hash commits in all actions, for specific versions (pls see e.g. checkout action, above)

`@latest` is definetely a no-go 😉

vcpkg.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "main",
+  "version-string": "latest",
```
lukaszstolarczuk, line 3: `latest` is no good from the security perspective

.github/workflows/basic.yml
lukaszstolarczuk, line 214: hmm... is this really a requirement for run-vcpkg action? I can't see it listed on their GH page

.github/workflows/basic.yml
```diff
@@ -208,11 +210,28 @@ jobs:
...
+      with:
+        cmakeVersion: 3.22.6
+
+    - name: Initialize vcpkg
+      uses: lukka/run-vcpkg@v11.5
+      with:
+        vcpkgGitCommitId: 3dd44b931481d7a8e9ba412621fa810232b66289
```
lukaszstolarczuk, line 221: how did you come up with this? is it the version of vcpkg available on GHA runners? or something else?

.github/workflows/basic.yml
```diff
@@ -208,11 +210,28 @@ jobs:
...
+        vcpkgDirectory: ${{ github.workspace }}/build/vcpkg
+        vcpkgConfigurationJsonGlob: '**/vcpkg-configuration.json'
```
lukaszstolarczuk, line 223: I don't think we use this file...?

lukaszstolarczuk: actually we may want to use `vcpkgJsonGlob` instead and perhaps hide `vcpkg.json` file from root directory to `.github/...`

.github/workflows/basic.yml
```diff
@@ -208,11 +210,28 @@ jobs:
...
+
+    - name: Install dependencies
```
lukaszstolarczuk, line 225: can this be run as part of `lukka/run-vcpkg` action...? it seems so, based on their README page.

if we could use only 1 action instead of current 3 actions, it would be really nice 😉

.github/workflows/benchmarks.yml
```diff
@@ -67,3 +68,69 @@ jobs:
...
+
+    - name: Run benchmarks
+      working-directory: ${{env.BUILD_DIR}}
+      shell: pwsh # Specifies PowerShell as the shell for running the script.
+      run: |
+        # add path to umf.dll
+        $env:PATH += ";${{env.BUILD_DIR}}\bin\${{matrix.build_type}}"
+        # add path to hwloc-15.dll
+        $env:PATH += ";${{env.HWLOC_PATH}}\bin"
+        .\benchmark\${{matrix.build_type}}\ubench.exe
```
lukaszstolarczuk, line 136: hmm.. could/should these paths be set in our cmake...?

CMakeLists.txt
```diff
@@ -111,17 +111,22 @@ install(
...
     add_subdirectory(test)
 endif()
 
 if(UMF_BUILD_BENCHMARKS)
-    if(LINUX)
+    if(LINUX OR WINDOWS)
         add_subdirectory(benchmark)
     else()
-        message(FATAL_ERROR "benchmarks are supported on Linux only")
+        message(FATAL_ERROR "benchmarks are supported on Linux and Windows only")
```
lukaszstolarczuk, line 129: pls update top-level README

igchor:
👍

.github/workflows/basic.yml
ldorau: @KFilipek how did you come up with this? ;-)

.github/workflows/basic.yml
ldorau: Done, but not pushed out.

vcpkg.json
ldorau: Done, but not pushed out.

.github/workflows/basic.yml
ldorau: I can't see it listed on their GH page either. @KFilipek  is this really a requirement for run-vcpkg action?

It installs the latest CMake and Ninja. Let's try remove it.

.github/workflows/basic.yml
ldorau: Replaced with `vcpkgJsonGlob: '**/vcpkg.json'`

.github/workflows/basic.yml
ldorau: I tried and didn't succeed. Someone can play with it later.

.github/workflows/benchmarks.yml
ldorau: They already are set by CMake.

CMakeLists.txt
ldorau: Done

src/provider/provider_os_memory_windows.c
ldorau: IMHO `umfProviderPurge()` should **NOT** do exactly the same what `umfProviderFree()` does (e.g. `VirtualFree(addr, 0, MEM_RELEASE)`), otherwise `umfProviderFree()` would always fail when called after `umfProviderPurge()`

bratpiorka: But you can't call the free after calling the purge - please refer to e.g. jemalloc documentation. free() is weaker than purge(). free() means that the pointer shouldn't be used to access a memory but there are no requirements on what to do with this memory (you could either destroy == unmap it or return to the pool) while the purge() means that you free this memory AND destroys it which returns it to the OS or device. You could do this immediately (force version) or later (lazy version) as it potentially is costly. @igchor please correct me if I'm wrong

ldorau: but jemalloc calls in a sequence its hooks:
1) `arena_extent_purge_lazy()` which calls `umfMemoryProviderPurgeLazy()` then:
2) `arena_extent_purge_forced()` which calls `umfMemoryProviderPurgeForce()` and finally:
3) `arena_extent_destroy()` which calls `umfMemoryProviderFree()` and finally:

ldorau: > But you can't call the free after calling the purge - please refer to e.g. jemalloc documentation. free() is weaker than purge(). 

AFAIK it is the opposite: you can't call purge() after calling free() - free() unmaps the memory and you cannot do anything else with it later. On Linux `purge_lazy()` is implemented as `madvise(MADV_FREE)` and `purge_force()` is implemented as `madvise(MADV_DONTNEED)`, but `madvise()` can be called only on the mapped region of memory, so if you free (munmap()) it, you cannot do anything else more with it.

.github/workflows/basic.yml
ldorau: Removed

src/provider/provider_os_memory_windows.c
bratpiorka: ok I'm resolving this as we agreed that we'll fix the provider API in the separate PR

bratpiorka (approved):

.github/workflows/basic.yml
KFilipek: According to: https://github.com/lukka/run-vcpkg#use-vcpkgs-vcpkgjson-file-to-specify-the-dependencies
Other way to install:
Or it can also be used by invoking `vcpkg install` in a directory where vcpkg.json is located (e.g., input runVcpkgInstall : true).

.github/workflows/basic.yml
KFilipek: As far as I remember, it's taken from their (vcpkg) repository.

ldorau:
Added 
```
Co-developed-by: Krzysztof Filipek <krzysztof.filipek@intel.com>
```
to the first commit "_Install hwloc on Windows using Vcpkg_"

KFilipek (approved):

.github/workflows/basic.yml
KFilipek: https://github.com/microsoft/vcpkg/commit/3dd44b931481d7a8e9ba412621fa810232b66289

src/libumf_noOS.def
```diff
@@ -0,0 +1,47 @@
+;;;; Begin Copyright Notice
```
PatKamin, line 1: Perhaps libumf_no_os_provider would be a better name for this file. The current name suggests "no operating system" which is confusing. It's not clear that OS refers to the kind of a provider.

src/provider/provider_os_memory_windows.c
```diff
@@ -0,0 +1,80 @@
+/*
+ * Copyright (C) 2023 Intel Corporation
```
PatKamin, line 2: Bump the date

.github/workflows/benchmarks.yml
ldorau: @lukaszstolarczuk  It is Windows. You have to have paths to all required DLLs in your PATH environment variable.
I do not know how we could do it in another way. If you know a working solution, please let me know.

ldorau: > I can see we have something similar set for tests (`PATH_LIST`) - not sure if this is applyable here...?

Do you mean `PATH=path_list_append:` ? I am just doing here exactly what `PATH=path_list_append:` does :-)

lukaszstolarczuk: discussed offline, agreed we can keep it as is

.github/workflows/basic.yml
lukaszstolarczuk: 😄 the one under which is my issue

We shouldn't use `lukka/run-vcpkg@v11.5` like this - we have to use a hash sha corresponding to the tag, and tag put only in the comment (for us, for readability)

example of properly used action is above - the `checkout` action.
<br />

Solution for you (please apply everywhere):
```
lukka/run-vcpkg@5e0cab206a5ea620130caf672fce3e4a6b5666a1 # v11.5
```

.github/workflows/benchmarks.yml
```diff
@@ -67,3 +68,65 @@ jobs:
...
+
+  benchmarks-windows:
+    name: Windows
+    env:
+      BUILD_DIR : "${{github.workspace}}/build/"
+      HWLOC_PATH: "${{github.workspace}}/build/vcpkg/packages/hwloc_x64-windows"
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ['windows-2022']
```
PatKamin, line 80: This should be hardcoded in-place as it is used only in one place and this variable doesn't have any other values than `windows-2022`.

.github/workflows/benchmarks.yml
```diff
@@ -67,3 +68,65 @@ jobs:
...
+
+    - name: Install dependencies
+      run: vcpkg install
+      shell: pwsh # Specifies PowerShell as the shell for running the script.
+
+    - name: Configure build
+      run: >
+        cmake
+        -B ${{env.BUILD_DIR}}
+        -DCMAKE_BUILD_TYPE=${{matrix.build_type}}
```
PatKamin, line 105: This is ignored on Windows. Passing a `--config` parameter in the build command suffices.

.github/workflows/benchmarks.yml
ldorau: Done

src/provider/provider_os_memory_windows.c
ldorau: Done

src/libumf_noOS.def
ldorau: Done

.github/workflows/basic.yml
ldorau: Done

PatKamin (approved):

.github/workflows/benchmarks.yml
```diff
@@ -67,3 +68,63 @@ jobs:
...
+        compiler: [{c: cl, cxx: cl}]
+        shared_library: ['ON', 'OFF']
+    runs-on: 'windows-2022'
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+
+    - name: Initialize vcpkg
+      uses: lukka/run-vcpkg@v11.5
```
lukaszstolarczuk, line 90: not done here :(

.github/workflows/basic.yml
lukaszstolarczuk: > This issue is under:
> ```
>     - name: Get CMake for lukka/run-vcpkg
>       uses: lukka/get-cmake@latest
> ```

haha, ok, mea culpa I didn't realize it's the old action 😄 

thx!

.github/workflows/benchmarks.yml
ldorau: Done

lukaszstolarczuk (approved):

igchor (approved):

PR #272, opened 2024-02-28
Global base alloc free function now does not require size param

igchor:
Also, implement malloc_usable_size and aligned_alloc for global
base alloc.

### DIFF
--- a/src/base_alloc/base_alloc_global.c
+++ b/src/base_alloc/base_alloc_global.c
@@ -14,13 +14,16 @@
 #include "base_alloc.h"
 #include "base_alloc_global.h"
 #include "base_alloc_internal.h"
+#include "utils_common.h"
 #include "utils_concurrency.h"
 #include "utils_math.h"
 
 // global base allocator used by all providers and pools
 static UTIL_ONCE_FLAG ba_is_initialized = UTIL_ONCE_FLAG_INIT;
 
-// allocation classes need to be powers of 2
+#define ALLOC_METADATA_SIZE (sizeof(size_t))
+
+// allocation classes need to be consecutive powers of 2
 #define ALLOCATION_CLASSES                                                     \
     { 16, 32, 64, 128 }
 #define NUM_ALLOCATION_CLASSES 4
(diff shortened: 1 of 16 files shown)


ldorau (approved):

src/base_alloc/base_alloc_global.c
```diff
@@ -82,41 +83,121 @@ static int size_to_idx(size_t size) {
...
+    if (alignment <= sizeof(size_t)) {
+        user_ptr = (void *)((uintptr_t)ptr + sizeof(size_t));
+    } else {
+        user_ptr = (void *)ALIGN_UP((uintptr_t)ptr + sizeof(size_t), alignment);
+    }
+
+    size_t ptr_offset_from_original = (uintptr_t)user_ptr - (uintptr_t)ptr;
+
+    size_t *metadata_loc = (size_t *)((char *)user_ptr - sizeof(size_t));
+    *metadata_loc = size | (ptr_offset_from_original << 32);
```
bratpiorka, line 105: please assert that the bits we truncate are zeros

src/base_alloc/base_alloc_global.c
```diff
@@ -82,41 +83,121 @@ static int size_to_idx(size_t size) {
...
+    void *user_ptr;
+    if (alignment <= sizeof(size_t)) {
+        user_ptr = (void *)((uintptr_t)ptr + sizeof(size_t));
+    } else {
+        user_ptr = (void *)ALIGN_UP((uintptr_t)ptr + sizeof(size_t), alignment);
+    }
+
+    size_t ptr_offset_from_original = (uintptr_t)user_ptr - (uintptr_t)ptr;
+
+    size_t *metadata_loc = (size_t *)((char *)user_ptr - sizeof(size_t));
```
bratpiorka, line 104: you use "sizeof(size_t)" several times - could you make a define for this with a meaningful name like "METADATA_OFFSET" or similar?

src/base_alloc/base_alloc_global.c
```diff
@@ -82,41 +83,121 @@ static int size_to_idx(size_t size) {
     return index;
 }
 
-void *umf_ba_global_alloc(size_t size) {
+// stores metadata just before 'ptr' and returns beginning of usable
+// space to the user. metadata consists of 'size' that is the allocation
+// size and 'offset' that specifies how far is the returned ptr from
+// the origin ptr (used for aligned alloc)
+static void *transform_ptr(void *ptr, size_t size, size_t alignment) {
```
bratpiorka, line 90: I think the name "transform_ptr" is too generic - maybe better would be an "add_metadata"? or a pair of "pack/unpack_ptr"?

bratpiorka (changes requested):

src/base_alloc/base_alloc_global.c
igchor: This is already asserted above `assert(size < (1ULL << 32));` and `assert(alignment < (1ULL << 32));`

src/base_alloc/base_alloc_global.c
igchor: Done.

src/base_alloc/base_alloc_global.c
igchor: Done.

src/base_alloc/base_alloc_global.c
bratpiorka: I meant the check for ptr_offset_from_original.

assert(ptr_offset_from_original < (1ULL << 32));

igchor: Since alignment is checked to be < (1 << 32) the offset can't be bigger than that. But done, I've added the check anyway,

bratpiorka (approved):

Issue #217, opened 2024-02-06
Add valgrind annotations to base_alloc (and os provider?)


igchor:
So that valgrind can verify double-free, use-after-free, etc. We can use macros like VALGRIND_MAKE_MEM_UNDEFINED, VALGRIND_MAKE_MEM_DEFINED  and VALGRIND_DO_MAKE_MEM_NOACCESS.

Also, investigate whether there are some ASAN/UBSAN annotations we could use.

PR #275, opened 2024-02-29
Add memcheck and asan annotations

igchor:
Closes https://github.com/oneapi-src/unified-memory-framework/issues/217

TBB pool is not annotated because tbb keeps metadata inline and hence we cannot mark the memory as undefined/unaccessible on free.

Ref: https://valgrind.org/docs/manual/valgrind_manual.pdf

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,6 +23,7 @@ option(USE_ASAN "Enable AddressSanitizer checks" OFF)
 option(USE_UBSAN "Enable UndefinedBehaviorSanitizer checks" OFF)
 option(USE_TSAN "Enable ThreadSanitizer checks" OFF)
 option(USE_MSAN "Enable MemorySanitizer checks" OFF)
+option(USE_VALGRIND "Enable Valgrind instrumentation" OFF)
 
 # For using the options listed in the OPTIONS_REQUIRING_CXX variable
 # a C++17 compiler is required. Moreover, if these options are not set,
--- a/src/base_alloc/base_alloc.c
+++ b/src/base_alloc/base_alloc.c
@@ -11,6 +11,7 @@
 #include "base_alloc_internal.h"
 #include "utils_common.h"
 #include "utils_concurrency.h"
+#include "utils_sanitizers.h"
 
 // minimum size of a single pool of the base allocator
 #define MINIMUM_POOL_SIZE (ba_os_get_page_size())
@@ -80,7 +81,10 @@ static void ba_debug_checks(umf_ba_pool_t *pool) {
     umf_ba_chunk_t *next_chunk = pool->metadata.free_list;
     while (next_chunk) {
         n_free_chunks++;
+        utils_annotate_memory_defined(next_chunk, sizeof(umf_ba_chunk_t));
+        umf_ba_chunk_t *tmp = next_chunk;
         next_chunk = next_chunk->next;
+        utils_annotate_memory_inaccessible(tmp, sizeof(umf_ba_chunk_t));
     }
     assert(n_free_chunks == pool->metadata.n_chunks - pool->metadata.n_allocs);
 }
@@ -89,6 +93,9 @@ static void ba_debug_checks(umf_ba_pool_t *pool) {
 // ba_divide_memory_into_chunks - divide given memory into chunks of chunk_size and add them to the free_list
 static void ba_divide_memory_into_chunks(umf_ba_pool_t *pool, void *ptr,
                                          size_t size) {
+    // mark the memory temporarily accessible to perform the division
+    utils_annotate_memory_undefined(ptr, size);
+
     assert(pool->metadata.free_list == NULL);
     assert(size > pool->metadata.chunk_size);
 
@@ -112,6 +119,17 @@ static void ba_divide_memory_into_chunks(umf_ba_pool_t *pool, void *ptr,
 
     current_chunk->next = NULL;
     pool->metadata.free_list = ptr; // address of the first chunk
+
+    // mark the memory as unaccessible again
+    utils_annotate_memory_inaccessible(ptr, size);
+}
+
+static void *ba_os_alloc_annotated(size_t pool_size) {
+    void *ptr = ba_os_alloc(pool_size);
+    if (ptr) {
+        utils_annotate_memory_inaccessible(ptr, pool_size);
+    }
+    return ptr;
 }
 
 umf_ba_pool_t *umf_ba_create(size_t size) {
@@ -127,10 +145,13 @@ umf_ba_pool_t *umf_ba_create(size_t size) {
 
     pool_size = ALIGN_UP(pool_size, ba_os_get_page_size());
 
-    umf_ba_pool_t *pool = (umf_ba_pool_t *)ba_os_alloc(pool_size);
+    umf_ba_pool_t *pool = (umf_ba_pool_t *)ba_os_alloc_annotated(pool_size);
     if (!pool) {
         return NULL;
     }
+
+    // annotate metadata region as accessible
+    utils_annotate_memory_undefined(pool, offsetof(umf_ba_pool_t, data));
 
     pool->metadata.pool_size = pool_size;
     pool->metadata.chunk_size = chunk_size;
@@ -140,6 +161,8 @@ umf_ba_pool_t *umf_ba_create(size_t size) {
     pool->metadata.n_pools = 1;
     pool->metadata.n_chunks = 0;
 #endif /* NDEBUG */
+
+    utils_annotate_memory_defined(pool, offsetof(umf_ba_pool_t, data));
 
     char *data_ptr = (char *)&pool->data;
     size_t size_left = pool_size - offsetof(umf_ba_pool_t, data);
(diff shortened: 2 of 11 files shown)


CMakeLists.txt
```diff
@@ -22,6 +22,7 @@ option(USE_ASAN "Enable AddressSanitizer checks" OFF)
 option(USE_UBSAN "Enable UndefinedBehaviorSanitizer checks" OFF)
 option(USE_TSAN "Enable ThreadSanitizer checks" OFF)
 option(USE_MSAN "Enable MemorySanitizer checks" OFF)
+option(USE_VALGRIND "Enable Valgrind instrumentation" OFF)
```
lukaszstolarczuk, line 25: pls add this new option to README

igchor: Done.

src/base_alloc/base_alloc.c
```diff
@@ -112,6 +119,17 @@ static void ba_divide_memory_into_chunks(umf_ba_pool_t *pool, void *ptr,
...
     pool->metadata.free_list = ptr; // address of the first chunk
+
+    // mark the memory as unaccessible again
+    utils_annotate_memory_unaccessible(ptr, size);
+}
+
+static void *ba_os_alloc_annotated(size_t pool_size) {
+    void *ptr = ba_os_alloc(pool_size);
+    if (ptr) {
+        utils_annotate_memory_unaccessible(ptr, pool_size);
```
bratpiorka, line 130: shouldn't this be just unndefined?

src/base_alloc/base_alloc.c
```diff
@@ -127,11 +145,14 @@ umf_ba_pool_t *umf_ba_create(size_t size) {
...
     pool_size = ALIGN_UP(pool_size, ba_os_get_page_size());
 
-    umf_ba_pool_t *pool = (umf_ba_pool_t *)ba_os_alloc(pool_size);
+    umf_ba_pool_t *pool = (umf_ba_pool_t *)ba_os_alloc_annotated(pool_size);
     if (!pool) {
         return NULL;
     }
 
+    // annotate metadata region as accessible
+    utils_annotate_memory_undefined(pool, offsetof(umf_ba_pool_t, data));
```
bratpiorka, line 154: I think ba_os_alloc_annotated() should already mark this mem as undefined

src/pool/pool_disjoint.cpp
```diff
@@ -392,6 +397,7 @@ static void *memoryProviderAlloc(umf_memory_provider_handle_t hProvider,
     if (ret != UMF_RESULT_SUCCESS) {
         throw MemoryProviderError{ret};
     }
+    utils_annotate_memory_unaccessible(ptr, size);
```
bratpiorka, line 400: after alloc shouldn't it be undefined?

src/pool/pool_jemalloc.c
```diff
@@ -72,7 +74,14 @@ static void *arena_extent_alloc(extent_hooks_t *extent_hooks, void *new_addr,
         return NULL;
     }
 
+#ifndef __SANITIZE_ADDRESS__
+    // jemalloc might write to new extents in realloc, so we cannot
+    // mark them as unaccessible under asan
+    utils_annotate_memory_unaccessible(ptr, size);
```
bratpiorka, line 80: why not undefined? I think this is not consistent with the memset() called later where you mark mem as defined

src/pool/pool_jemalloc.c
```diff
@@ -72,7 +74,14 @@ static void *arena_extent_alloc(extent_hooks_t *extent_hooks, void *new_addr,
...
+#endif
+
     if (*zero) {
+        utils_annotate_memory_defined(ptr, size);
```
bratpiorka, line 84: I think we should mark mem as defined only in calloc, not in extent management

src/pool/pool_jemalloc.c
```diff
@@ -72,7 +74,14 @@ static void *arena_extent_alloc(extent_hooks_t *extent_hooks, void *new_addr,
...
         memset(ptr, 0, size); // TODO: device memory is not accessible by host
     }
```
bratpiorka, line 87: don't you miss setting mem as unaccesible after extent dealloc?

src/pool/pool_jemalloc.c
```diff
@@ -290,6 +302,7 @@ static umf_result_t je_free(void *pool, void *ptr) {
     assert(pool);
 
     if (ptr != NULL) {
+        VALGRIND_DO_MEMPOOL_FREE(pool, ptr);
```
bratpiorka, line 305: missing utils_annotate_memory_unaccessible?

src/pool/pool_jemalloc.c
```diff
@@ -314,6 +329,7 @@ static void *je_realloc(void *pool, void *ptr, size_t size) {
     if (size == 0 && ptr != NULL) {
         dallocx(ptr, MALLOCX_TCACHE_NONE);
         TLS_last_allocation_error = UMF_RESULT_SUCCESS;
+        VALGRIND_DO_MEMPOOL_FREE(pool, ptr);
```
bratpiorka, line 332: missing utils_annotate_memory_unaccessible?

src/pool/pool_jemalloc.c
```diff
@@ -329,6 +345,12 @@ static void *je_realloc(void *pool, void *ptr, size_t size) {
         return NULL;
     }
 
+    if (new_ptr != ptr) {
+        VALGRIND_DO_MEMPOOL_ALLOC(pool, new_ptr, size);
+        VALGRIND_DO_MEMPOOL_FREE(pool, ptr);
```
bratpiorka, line 350: missing utils_annotate_memory_unaccessible(ptr)?

test/supp/umf_test-jemalloc_pool.supp
```diff
@@ -0,0 +1,8 @@
+{
+   realloc suppresion
+   Memcheck:Addr8
+   fun:memmove
```
bratpiorka, line 4: could we add comments here? Are they TODO to fix or false positives?

bratpiorka (changes requested):

bratpiorka:
@igchor about TBB - why do we can't annotate mem as unaccessible on free()? Even if TBB keeps metadata inline, you pass ptr+size to the Valgrind so you annotate only the portion of memory - internal TBB accesses outside this portion shouldn't cause problems?

src/pool/pool_jemalloc.c
igchor: No, since we don't mark memory as inaccessible under address sanitizer, we don't need to do anything here for sanitizer. We only rely on valgrind here which already marks memory as inaccessible inside MEMPOOL_FREE.

src/pool/pool_jemalloc.c
igchor: No, since we don't mark memory as inaccessible under address sanitizer, we don't need to do anything here for sanitizer. We only rely on valgrind here which already marks memory as inaccessible inside MEMPOOL_FREE.

igchor:
> @igchor about TBB - why do we can't annotate mem as unaccessible on free()? Even if TBB keeps metadata inline, you pass ptr+size to the Valgrind so you annotate only the portion of memory - internal TBB accesses outside this portion shouldn't cause problems?

TBB seems to store some metadata in the allocation itself (first 8 bytes) after free (similarly as we do in base_alloc - but in base_alloc we know exactly where and how much metadata we store). Which means that if we mark the allocation as inaccessible, it will trigger an error when tbb writes it's metadata.

bratpiorka:
> TBB seems to store some metadata in the allocation itself (first 8 bytes) after free (similarly as we do in base_alloc - but in base_alloc we know exactly where and how much metadata we store). Which means that if we mark the allocation as inaccessible, it will trigger an error when tbb writes it's metadata.

so can we annotate the rest of memory, after these 8 bytes? or this doesn't make sense?

bratpiorka (approved):

bratpiorka:
@kswiecicki we need second approver - could you look at this?

igchor:
> > TBB seems to store some metadata in the allocation itself (first 8 bytes) after free (similarly as we do in base_alloc - but in base_alloc we know exactly where and how much metadata we store). Which means that if we mark the allocation as inaccessible, it will trigger an error when tbb writes it's metadata.
> 
> so can we annotate the rest of memory, after these 8 bytes? or this doesn't make sense?

We could if we're sure that tbb always uses only 8 bytes. This might be true but I don't know if we want to rely on such behavior.

src/pool/pool_jemalloc.c
```diff
@@ -305,6 +317,8 @@ static void *je_calloc(void *pool, size_t num, size_t size) {
         return NULL;
     }
 
+    utils_annotate_memory_defined(ptr, num * size);
```
kswiecicki, line 320: We mark the memory in `je_calloc` and `je_realloc` as defined since its zero-initialized or copied, so why is the memory from `je_malloc` not marked as undefined?

kswiecicki (approved):

PR #276, opened 2024-03-01
return error when hwloc_get_nbobjs_by_type fails

bratpiorka:
This PR fixes coverity issue #437904 https://scan3.scan.coverity.com/#/project-view/65229/15141?selectedIssue=437904

### DIFF
--- a/src/memspaces/memspace_host_all.c
+++ b/src/memspaces/memspace_host_all.c
@@ -40,7 +40,10 @@ static umf_result_t umfMemspaceHostAllCreate(umf_memspace_handle_t *hMemspace) {
     // object that can be present on multiple levels.
     // Source: https://www.open-mpi.org/projects/hwloc/doc/hwloc-v2.10.0-letter.pdf
     int nNodes = hwloc_get_nbobjs_by_type(topology, HWLOC_OBJ_NUMANODE);
-    assert(nNodes != -1);
+    if (nNodes < 0) {
+        umf_ret = UMF_RESULT_ERROR_UNKNOWN;
+        goto err_topology_destroy;
+    }
 
     size_t *nodeIds = umf_ba_global_alloc(nNodes * sizeof(size_t));
     if (!nodeIds) {


src/memspaces/memspace_host_all.c
```diff
@@ -40,7 +40,10 @@ static umf_result_t umfMemspaceHostAllCreate(umf_memspace_handle_t *hMemspace) {
     // object that can be present on multiple levels.
```
lukaszstolarczuk, line 40: should you also update the comment here...?

src/memspaces/memspace_host_all.c
```diff
@@ -40,7 +40,10 @@ static umf_result_t umfMemspaceHostAllCreate(umf_memspace_handle_t *hMemspace) {
...
     // Source: https://www.open-mpi.org/projects/hwloc/doc/hwloc-v2.10.0-letter.pdf
     int nNodes = hwloc_get_nbobjs_by_type(topology, HWLOC_OBJ_NUMANODE);
-    assert(nNodes != -1);
+    if (nNodes < 0) {
+        umf_ret = UMF_RESULT_ERROR_UNKNOWN;
```
lukaszstolarczuk, line 44: is there anything we can print here, perhaps...?

kswiecicki (approved):

src/memspaces/memspace_host_all.c
KFilipek: AFAIK, the logger is not ready yet.

KFilipek (approved):

src/memspaces/memspace_host_all.c
bratpiorka: please propose how it should be changed

lukaszstolarczuk: We changed from checking `!= -1` to `<0`, so I ghouth we could update the part about `-1`... but I think I got ahead of myself, this func can only return `-1` and `0`, so no need to change the comment, after all.

src/memspaces/memspace_host_all.c
bratpiorka: This should be done for each error in this file (and the whole project), in the separate PR. This PR is about fixing the Coverity issue.

lukaszstolarczuk (approved):

PR #278, opened 2024-03-01
Fix problem with improper use of negative value (nNodes)

szadam:
CID: 437905
Probably it will fix CID: 437905


kswiecicki:
There seems to already by a fix for this here: https://github.com/oneapi-src/unified-memory-framework/pull/276.

PR #280, opened 2024-03-01
refactor ba_global_alloc and fix coverity issues

bratpiorka:
This PR slightly simplifies umf_ba_global_alloc/free and fixes coverity errors #436585 and #436586

### DIFF
--- a/src/base_alloc/base_alloc_global.c
+++ b/src/base_alloc/base_alloc_global.c
@@ -69,8 +69,6 @@ static void umf_ba_create_global(void) {
 
 // returns index of the allocation class for a given size
 static int size_to_idx(size_t size) {
-    assert(size <= BASE_ALLOC.ac_sizes[NUM_ALLOCATION_CLASSES - 1]);
-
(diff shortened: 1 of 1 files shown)


src/base_alloc/base_alloc_global.c
```diff
@@ -94,24 +91,16 @@ void *umf_ba_global_alloc(size_t size) {
         return ba_os_alloc(size);
     }
 
-    int ac_index = size_to_idx(size);
-    if (!BASE_ALLOC.ac[ac_index]) {
```
igchor, line 98: I don't think you should remove this check. Creating ac can fail and there's no reason we can't serve the allocation using ba_os_alloc then. If you want to simplify the code, you can just combine this check with the one above.

igchor (changes requested):

src/base_alloc/base_alloc_global.c
bratpiorka: right, done

igchor (approved):

ldorau (approved):

PR #289, opened 2024-03-05
Add badge for the Coverity build

ldorau:
The current badge reflects the status of the Coverity report and it is green even if the Coverity build fails.
The Coverity build fails and nobody knows about it, becase there is no badge that reflects it.
So let's add a badge for the Coverity build.


kswiecicki (approved):

.github/workflows/coverity.yml
```diff
@@ -1,5 +1,5 @@
 # Coverity check
-name: Coverity
```
lukaszstolarczuk, line 2: I would rather not change the name and just use one badge (the one with report). The Coverity build badge is not that useful, I think

ldorau: Read the commit description, please:

_The current badge reflects the status of the Coverity report and it is green even if the Coverity build fails.
The Coverity build fails and nobody knows about it, becase there is no badge that reflects it._

You will not see when the Coverity build fails and the Coverity report will be out-dated.

lukaszstolarczuk: I get that, but I didn't think anyone looks at these badges, heh 😉 

Nonehtless, I wouldn't change the name of the workflow (badges names can and should be different, though, of course)

ldorau: Done

ldorau: @lukaszstolarczuk Done

KFilipek (approved):

Issue #288, opened 2024-03-05
UMF build fails by default with fatal error: level_zero/ze_api.h: No such file or directory


ldorau:
UMF build fails by default with fatal error: level_zero/ze_api.h: No such file or directory
```
/home/runner/work/unified-memory-framework/unified-memory-framework/src/provider/provider_level_zero.c:15:10: fatal error: level_zero/ze_api.h: No such file or directory
   15 | #include <level_zero/ze_api.h>
      |          ^~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```
(see https://github.com/oneapi-src/unified-memory-framework/actions/runs/8148957053/job/22272779208)
because `UMF_BUILD_LEVEL_ZERO_PROVIDER=ON` by default and there is no check for the level zero library.

## Environment Information

- UMF version (hash commit or a tag): 0d83bd097a7b84a3816f65b3ce1baf1b026007dc
- OS(es) version(s): Linux

bratpiorka:
fixed by #297

PR #290, opened 2024-03-05
Fix failing Coverity build

ldorau:
Ref: #288


kswiecicki (approved):

lukaszstolarczuk:
hmm... shouldn't rather L0 Provider be OFF by default? README says so, for sure 😉 

Also, in README we have improper name of this CMake option.

ldorau:
> hmm... shouldn't rather L0 Provider be OFF by default? README says so, for sure 😉
> 
> Also, in README we have improper name of this CMake option.

You are right.
I have submitted a new PR: https://github.com/oneapi-src/unified-memory-framework/pull/291
But this one can be merged too.

lukaszstolarczuk (approved):

PR #295, opened 2024-03-05
Fix sporadic issue in memspaceHostAllProviderTest.memoryPolicyOOM test

vinser52:
I removed the check that caused a sporadic test failure. 
I think we cannot guarantee in the test that all nodes will be touched because some other processes on the system might consume all memory from a particular NUMA node and as a result test is not getting memory allocated from that NUMA node.

### DIFF
--- a/test/memspaces/memspace_host_all.cpp
+++ b/test/memspaces/memspace_host_all.cpp
@@ -157,9 +157,14 @@ TEST_F(memspaceHostAllProviderTest, memoryPolicyOOM) {
         UT_ASSERTeq(umf_ret, UMF_RESULT_SUCCESS);
     }
 
+    // TODO: we want to enable this check only when tests are running under QEMU.
+    // Otherwise it might sporadically fail on a real system where other processes
+    // occupied all memory from a aparticular NUMA node.
+#if 0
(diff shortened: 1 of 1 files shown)


igchor:
I think we will want to actually have this check once we have a setup with qemu where we can guarantee proper amount of memory. Could we just put this check under an ifdef for now?

vinser52:
> I think we will want to actually have this check once we have a setup with qemu where we can guarantee proper amount of memory. Could we just put this check under an ifdef for now?

Sure. do we want to introduce a special macro name now? Do you have ideas?

vinser52:
@igchor I added `if def` with hardcoded `0` and `TODO` comments.

igchor:
> > I think we will want to actually have this check once we have a setup with qemu where we can guarantee proper amount of memory. Could we just put this check under an ifdef for now?
> 
> Sure. do we want to introduce a special macro name now? Do you have ideas?

Maybe something like `UMF_TEST_STRICT_NUMA_CHECK`? I think we might also use this for things like I described here: https://github.com/oneapi-src/unified-memory-framework/pull/241#issuecomment-1954662229

igchor (approved):

ldorau (approved):

PR #299, opened 2024-03-06
Fix data race condition - addlock critnib.mutex

szadam:
Fix coverity issue:
436590 [Data race condition](https://scan3.scan.coverity.com/doc/en/cov_checker_ref.html#static_checker_MISSING_LOCK)


igchor (changes requested):
There is no race here, coverity just does not understand the lock-free algorithm. See top of the file for description. We should just mark this as false-positive.

PR #300, opened 2024-03-06
Unify benchmarks' CI workflows for Linux and Windows

ldorau:
Unify benchmarks' CI workflows for Linux and Windows.
Use ctest to run benchmarks.

A benchmark passes if it prints "PASSED" in the output,
because ubench of scalable pool fails if the confidence interval
exceeds maximum permitted 2.5%.

Fixes: #301

### DIFF
--- a/.github/workflows/benchmarks.yml
+++ b/.github/workflows/benchmarks.yml
@@ -7,14 +7,20 @@ permissions:
   contents: read
 
 jobs:
-  benchmarks-ubuntu:
-    name: Ubuntu
+  benchmarks:
+    name: Benchmarks
+    env:
+      BUILD_DIR : "${{github.workspace}}/build/"
+      VCPKG_PATH: "${{github.workspace}}/build/vcpkg/packages/hwloc_x64-windows;${{github.workspace}}/build/vcpkg/packages/tbb_x64-windows"
     strategy:
       matrix:
-        os: ['ubuntu-22.04']
-        build_type: [Release]
-        compiler: [{c: gcc, cxx: g++}]
-        shared_library: ['ON', 'OFF']
+        include:
+          - os: ubuntu-latest
+            pool_jemalloc: 'ON'
+
+          - os: windows-latest
+            pool_jemalloc: 'OFF'
+
     runs-on: ${{matrix.os}}
 
     steps:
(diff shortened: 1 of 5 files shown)


benchmark/multithread.cpp
```diff
@@ -65,15 +65,15 @@ static void mt_alloc_free(poolCreateExtParams params,
     auto values = umf_bench::measure<std::chrono::milliseconds>(
         bench.n_repeats, bench.n_threads,
         [&, pool = pool.get()](auto thread_id) {
-            for (int i = 0; i < bench.n_iterations; i++) {
+            for (int i = 0; i < (int)bench.n_iterations; i++) {
```
bratpiorka, line 68: please change the "i" type instead casting n_iterations

.github/workflows/benchmarks.yml
```diff
@@ -7,125 +7,85 @@ permissions:
...
+            pool_jemalloc: 'ON'
+
+          - os: ubuntu-latest # static library on Linux
+            shared_library: 'OFF'
+            pool_scalable: 'ON'
+            pool_jemalloc: 'ON'
+
+          - os: windows-latest # shared library on Windows
+            shared_library: 'ON'
+            pool_scalable: 'OFF'
```
bratpiorka, line 30: could we wait to merge this PR until we enable a scalable pool on Windows? Then this would be simplified

.github/workflows/benchmarks.yml
```diff
@@ -7,125 +7,85 @@ permissions:
...
-        compiler: [{c: gcc, cxx: g++}]
-        shared_library: ['ON', 'OFF']
+        include:
+          - os: ubuntu-latest # shared library on Linux
+            shared_library: 'ON'
+            pool_scalable: 'ON'
+            pool_jemalloc: 'ON'
+
+          - os: ubuntu-latest # static library on Linux
+            shared_library: 'OFF'
```
bratpiorka, line 24: do we have any perf difference between shared lib ON and OFF? If not, I would test only shared lib and simplify this matrix

bratpiorka (changes requested):

benchmark/CMakeLists.txt
```diff
@@ -2,84 +2,84 @@
...
-	set(LIBS_OPTIONAL ${LIBS_OPTIONAL} disjoint_pool)
-endif()
-
-if(UMF_BUILD_LIBUMF_POOL_JEMALLOC)
-	set(LIBS_OPTIONAL ${LIBS_OPTIONAL} jemalloc_pool)
-endif()
-
-if(UMF_BUILD_LIBUMF_POOL_SCALABLE)
-	set(LIBS_OPTIONAL ${LIBS_OPTIONAL} scalable_pool)
+if (UMF_BUILD_BENCHMARKS_MT)
```
lukaszstolarczuk, line 9: Perhaps, consider moving these CMake changes into a separate commit..?

.github/workflows/benchmarks.yml
```diff
@@ -7,125 +7,85 @@ permissions:
...
 
+      - name: Initialize vcpkg
+        if: matrix.os == 'windows-latest'
+        uses: lukka/run-vcpkg@5e0cab206a5ea620130caf672fce3e4a6b5666a1 # v11.5
+        with:
+          vcpkgGitCommitId: 3dd44b931481d7a8e9ba412621fa810232b66289
+          vcpkgDirectory: ${{env.BUILD_DIR}}/vcpkg
+          vcpkgJsonGlob: '**/vcpkg.json'
+
+      - name: Install dependencies
```
lukaszstolarczuk, line 57: you can rename it to e.g. `Install vcpkg packages` (to be aligned in name with the `apt` step)

benchmark/CMakeLists.txt
```diff
@@ -2,84 +2,84 @@
...
+        WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
 
-if (UMF_BUILD_OS_MEMORY_PROVIDER)
-	target_compile_definitions(ubench PRIVATE UMF_BUILD_OS_MEMORY_PROVIDER=1)
+    # Benchmark passes if it prints "PASSED" in the output,
+    # because ubench of scalable pool fails if the confidence interval
+    # exceeds maximum permitted 2.5%.
+    set_tests_properties(${BENCH_NAME} PROPERTIES
+        LABELS "benchmark"
+        PASS_REGULAR_EXPRESSION "PASSED")
```
lukaszstolarczuk, line 40: hmm I'm a little confused. Didn't we check for lack of `exceeds maximum permitted 2.5` previously?

From what I can see we can print something like:
```
[ PASSED ] <number of passed>
[ FAILED ] <number of failed>
```
and this test will be green...?

benchmark/multithread.cpp
ldorau: Done

.github/workflows/benchmarks.yml
ldorau: As you wish (but it is not a big difference - just one option less).

.github/workflows/benchmarks.yml
ldorau: Done

.github/workflows/benchmarks.yml
ldorau: Done

benchmark/CMakeLists.txt
ldorau: Done

benchmark/CMakeLists.txt
ldorau: Yes, of course: https://github.com/oneapi-src/unified-memory-framework/blob/main/.github/workflows/benchmarks.yml#L54-L67

You can just check it in the diff of this patch (Files changed) ;-)

ldorau:
@bratpiorka Done
@lukaszstolarczuk Done

benchmark/CMakeLists.txt
lukaszstolarczuk: 😄 thanks, but of course I implied (with the second part of my comment), that the current solution will always pass and it is not what we wanted...? Or maybe I'm reading the code wrong...?

ldorau: From: https://cmake.org/cmake/help/latest/prop_test/PASS_REGULAR_EXPRESSION.html

"The test output (stdout or stderr) must match this regular expression for the test to pass. **The process exit code is ignored.** Tests that exceed the timeout specified by [TIMEOUT](https://cmake.org/cmake/help/latest/prop_test/TIMEOUT.html#prop_test:TIMEOUT) still fail regardless of PASS_REGULAR_EXPRESSION. **System-level test failures including segmentation faults, signal abort, or heap errors may fail the test even if PASS_REGULAR_EXPRESSION is matched.**"

I think it is OK for us, since it is not a test, but a benchmark.

ldorau: @lukaszstolarczuk I decided to use `PASS_REGULAR_EXPRESSION` in order:
1) to remove this complicated bash script in Linux and
2) to unify usage of `ubench ` in both Linux and Windows.

ldorau:
@bratpiorka please re-review

bratpiorka (approved):

lukaszstolarczuk (approved):

PR #303, opened 2024-03-06
Fix data race condition: Add lock pool in umf_ba_linear_destroy

szadam:
Fix coverity issue:
436584 [Data race condition](https://scan3.scan.coverity.com/doc/en/cov_checker_ref.html#static_checker_MISSING_LOCK)


ldorau (changes requested):
@szadam As @igchor wrote above - a destructor must NOT take the lock it is destroying. This fix is wrong. Please change it or close this PR.

Issue #161, opened 2024-01-18
The umf_test-jemalloc_pool test segfaults with debug version of jemalloc


igchor:
**The current state:**

The `umf_test-jemalloc_pool` test segfaults with jemalloc v5.3.0 compiled with `--enable-debug`.

```
# ./test/umf_test-jemalloc_pool
base_alloc: allocation size (336) larger than the biggest allocation class. Falling back to OS memory allocation.
Running main() from /root/unified-memory-framework/build/_deps/googletest-src/googletest/src/gtest_main.cc
[==========] Running 16 tests from 2 test suites.
[----------] Global test environment set-up.
[----------] 1 test from test
[ RUN      ] test.metadataNotAllocatedUsingProvider
base_alloc: allocation size (152) larger than the biggest allocation class. Falling back to OS memory allocation.
Segmentation fault (core dumped)
```

-----------------------------------

**The original issue:**

When running with the latest version of jemalloc, build in debug mode we got the following output:

```
<jemalloc>: Should own 0 locks of rank >= 14: extent_grow(17)
Aborted (core dumped)
```

The problem seems to be caused by recursive call to jemalloc (we are calling malloc which is being intercepted withing extent hooks).

This can be solved by using a custom allocator for every allocation made inside extent hooks (this include critnib and tracking provider allocations).

bratpiorka:
JIRA UMFW-237

ldorau:
@igchor I cannot reproduce it now. Could you verify if it still exists and close this issue?

igchor:
It still fails for me for jemalloc, version 5.3.0_163 compiled with --enable-debug. However, I don't get that output anymore, there's just segfault now:

```
# ./test/umf_test-jemalloc_pool
base_alloc: allocation size (336) larger than the biggest allocation class. Falling back to OS memory allocation.
Running main() from /root/unified-memory-framework/build/_deps/googletest-src/googletest/src/gtest_main.cc
[==========] Running 16 tests from 2 test suites.
[----------] Global test environment set-up.
[----------] 1 test from test
[ RUN      ] test.metadataNotAllocatedUsingProvider
base_alloc: allocation size (152) larger than the biggest allocation class. Falling back to OS memory allocation.
Segmentation fault (core dumped)
```

ldorau:
@igchor Thanks!

ldorau:
**Solution:**

[...]

PR #311, opened 2024-03-08
Do not link directly with libc

ldorau:
Do not link directly with libc.
It was used to force using malloc from glibc.
We do not need it any more, because we use our own base allocator now.


bratpiorka (approved):

ldorau:
Fixes: https://github.com/oneapi-src/unified-memory-framework/issues/161

PR #313, opened 2024-03-08
Adjust memory size that should be left...

kswiecicki:
unallocated when running HOST ALL memspace test that checks the NUMA node saturation.

### Checklist

- [ ] All tests pass locally


vinser52 (approved):
General comment: I think, such test still can sporadically fail on a real system. Only the environment controlled by us can guarantee that this test will pass.

kswiecicki:
> General comment: I think, such test still can sporadically fail on a real system. Only the environment controlled by us can guarantee that this test will pass.

I agree, but for some reason this test was failing only on the QEMU runner.

bratpiorka (approved):

PR #320, opened 2024-03-11
Enable the proxy library on windows

ldorau:


src/proxy_lib/proxy_lib.c
```diff
@@ -17,6 +17,21 @@
...
+// Disable warning about inconsistent dll linkage:
+// warning C4273: 'malloc': inconsistent dll linkage
+// ucrt\corecrt_malloc.h(101): note: see previous definition of 'malloc'
+#pragma warning(disable : 4273)
+#pragma warning(push, 1)
+#endif
+
+#ifdef _WIN32
+// fix the error: 'malloc' redeclared without 'dllimport' attribute: 'dllexport' attribute added
+#define WIN32_dllexport __declspec(dllexport)
```
igchor, line 30: Can't we just use the .map file like we do for other functions?

ldorau: We already use the `.def` file and this error occurs without `__declspec(dllexport)` anyway.

ldorau:
Tests added

CMakeLists.txt
```diff
@@ -117,6 +117,10 @@ install(
 if(WINDOWS)
     # set PATH to DLLs on Windows
     set(DLL_PATH_LIST "PATH=path_list_append:../bin/$<CONFIG>")
+    # add path to the proxy lib DLL
+    set(DLL_PATH_LIST "${DLL_PATH_LIST};PATH=path_list_append:../src/proxy_lib")
```
kswiecicki, line 121: nit: Could we use `${CMAKE_CURRENT_SOURCE_DIR}` or other predefined CMake `DIR` variables to refer to those directories?

src/CMakeLists.txt
```diff
@@ -181,20 +181,21 @@ install(TARGETS umf
 
 add_subdirectory(pool)
 
-# TODO: enable proxy_lib on Windows
-if(LINUX)
-    if(UMF_PROXY_LIB_BASED_ON_POOL STREQUAL SCALABLE)
-        set(PROXY_LIB_USES_SCALABLE_POOL ON)
-        set(PROXY_LIBS umf scalable_pool)
+if(LINUX OR WINDOWS)
```
kswiecicki, line 184: Now that proxy lib is available on linux and windows is this `if` necessary? Also, should we create a cmake option for building proxy lib?

src/proxy_lib/proxy_lib.c
```diff
@@ -260,7 +296,19 @@ void *aligned_alloc(size_t alignment, size_t size) {
...
+#ifdef _WIN32
+size_t _msize(void *ptr) {
+#else
 size_t malloc_usable_size(void *ptr) {
+#endif
+
+#ifndef NDEBUG
+    // debug check to verify we are running the proxy library
+    if (ptr == (void *)0x01) {
+        return 0xDEADBEAF;
```
kswiecicki, line 308: Totally nit: `0xDEADBEEF`.

src/CMakeLists.txt
ldorau: Yes, we have MACOSX too ... :-(
The option is not necessary as for now.

src/proxy_lib/proxy_lib.c
ldorau: Done.

CMakeLists.txt
ldorau: How do you want to do that? Did you try it? Does it work for you?

kswiecicki: I didn't test that, it's just that we usually refer to CMake vars when setting appropriate paths.

test/CMakeLists.txt
```diff
@@ -177,3 +184,25 @@ add_umf_test(NAME base_alloc_linear
...
+add_umf_test(NAME pool_os_alloc
+            SRCS pools/pool_os_alloc.cpp malloc_compliance_tests.cpp)
+
+# tests for the proxy library
+if((LINUX OR WINDOWS) AND UMF_BUILD_OS_MEMORY_PROVIDER AND UMF_BUILD_SHARED_LIBRARY)
+    if(WINDOWS AND NOT (CMAKE_BUILD_TYPE STREQUAL Release))
+        message(STATUS "Skipping the tests of the proxy library, because they are supported only in the Release build on Windows")
+    elseif((UMF_PROXY_LIB_BASED_ON_POOL STREQUAL SCALABLE) AND (NOT UMF_BUILD_LIBUMF_POOL_SCALABLE))
+        message(STATUS "Skipping the tests of the proxy library, because UMF_PROXY_LIB_BASED_ON_POOL==SCALABLE but UMF_BUILD_LIBUMF_POOL_SCALABLE is OFF")
+    elseif((UMF_PROXY_LIB_BASED_ON_POOL STREQUAL JEMALLOC) AND (NOT UMF_BUILD_LIBUMF_POOL_JEMALLOC))
```
igchor, line 197: Wouldn't it be better to just set some var UMF_SCALABLE_POOL_ENABLED in the main CMake after validating all the options and just have a single if() here?

test/pools/pool_os_alloc.cpp
```diff
@@ -0,0 +1,11 @@
+// Copyright (C) 2024 Intel Corporation
+// Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "pool.hpp"
+#include "poolFixtures.hpp"
+
+INSTANTIATE_TEST_SUITE_P(PoolOsAlloc, umfPoolTest,
```
igchor, line 8: What do you want to test here? We already have a test that uses MALLOC_POOL_OPS (memoryPoolAPI.cpp:143).

KFilipek (approved):

test/pools/pool_os_alloc.cpp
ldorau: Right. Replaced with `memoryPoolAPI.cpp`. 
`pool_os_alloc.cpp` removed.
Done.

test/CMakeLists.txt
ldorau: Right. Done.

test/CMakeLists.txt
```diff
@@ -177,3 +184,25 @@ add_umf_test(NAME base_alloc_linear
...
             SRCS ${BA_SOURCES_FOR_TEST} pools/pool_base_alloc.cpp malloc_compliance_tests.cpp
             LIBS umf_utils)
+
+add_umf_test(NAME pool_os_alloc
+            SRCS pools/pool_os_alloc.cpp malloc_compliance_tests.cpp)
+
+# tests for the proxy library
+if((LINUX OR WINDOWS) AND UMF_BUILD_OS_MEMORY_PROVIDER AND UMF_BUILD_SHARED_LIBRARY)
+    if(WINDOWS AND NOT (CMAKE_BUILD_TYPE STREQUAL Release))
+        message(STATUS "Skipping the tests of the proxy library, because they are supported only in the Release build on Windows")
```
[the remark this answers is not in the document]
ldorau: Because it does not work on Windows in case of the Debug build - there is the `_free_dbg()` function called instead of `free()` that cannot be found for unknown reason (debugged with @bratpiorka)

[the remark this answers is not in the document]
igchor: Ok, I see, but that's a problem for the entire proxy pool, not only in tests right? Should have this check in the main CMake? Also, we should probably add some todo and an issue to track this.

test/test_proxy_lib.cpp
```diff
@@ -0,0 +1,31 @@
...
+#include "base.hpp"
+#include "test_helpers.h"
+
+using umf_test::test;
+
+TEST_F(test, proxyLibBasic) {
+
+    ::free(::malloc(64));
+
+#ifndef NDEBUG
```
igchor, line 19: So, right now this is never run since proxy lib only works for Release? Perhaps use some other define for now, until we fix it for Debug?

test/CMakeLists.txt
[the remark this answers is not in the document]
ldorau: Right. Done: https://github.com/oneapi-src/unified-memory-framework/issues/350

test/test_proxy_lib.cpp
ldorau: I think we can enable this check in the Release build too.

CMakeLists.txt
lukaszstolarczuk: I believe the `../` part is disturbing a little - we don't know exactly how this behaves if cmake command would be run from some different place... but we can leave that as is, for now, I guess

In general we should probably use `${PROJECT_BINARY_DIR}`, but it would require checking the structure of dirs in Windows builds

lukaszstolarczuk (commented):
LGTM 👍

CMakeLists.txt
ldorau: @kswiecicki  @lukaszstolarczuk  Done

lukaszstolarczuk: Thx!

ldorau:
Rebased

bratpiorka (approved):

PR #322, opened 2024-03-11
Split memory provider ops to mandatory and optional fields

vinser52:
### Description

As we agreed on our sync, the memory provider ops structure is split into mandatory and optional fields. Mandatory fields should be always initialized with a pointer to the appropriate function. Optional fields can be NULL. If an optional field is NULL the UMF initializes it with a default implementation that returns `UMF_RESULT_ERROR_NOT_SUPPORTED` error code.

### Checklist

- [x] All tests pass locally

### DIFF
--- a/include/umf/memory_provider_ops.h
+++ b/include/umf/memory_provider_ops.h
@@ -17,6 +17,66 @@ extern "C" {
 #endif
 
 ///
+/// @brief This structure comprises optional function pointers used
+/// by corresponding umfMemoryProvider* calls. A memory provider implementation
+/// can keep them NULL.
+///
+typedef struct umf_memory_provider_ext_ops_t {
+    ///
+    /// @brief Discard physical pages within the virtual memory mapping associated at the given addr
+    ///        and \p size. This call is asynchronous and may delay purging the pages indefinitely.
+    /// @param provider pointer to the memory provider
+    /// @param ptr beginning of the virtual memory range
+    /// @param size size of the virtual memory range
+    /// @return UMF_RESULT_SUCCESS on success or appropriate error code on failure.
+    ///         UMF_RESULT_ERROR_INVALID_ALIGNMENT if ptr or size is not page-aligned.
+    ///         UMF_RESULT_ERROR_NOT_SUPPORTED if operation is not supported by this provider.
+    ///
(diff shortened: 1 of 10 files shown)


include/umf/memory_provider_ops.h
```diff
@@ -16,6 +16,64 @@
 extern "C" {
 #endif
 
+///
+/// @brief This structure comprises optional function pointers used
+/// by corresponding umfMemoryProvider* calls. A memory provider implementation
+/// can keep them NULL.
+///
+typedef struct umf_memory_provider_ext_ops_t {
```
igchor, line 24: I'm wondering if it wouldn't be better to put allocation_merge and allocation_split in a separate struct and enforce user to set both of them or neither of them. Allowing setting only one of them would be error-prone.

vinser52: Perhaps we can think of a smarter validation check (e.g. check that both are set or both are NULL), but I am not sure that moving each pair of corresponding functions to a separate struct is a good idea. We might come up with too many optional structs.

igchor: Ok, I don't think having more structs is necessarily a problem, but it might be tricky to name them so I guess we can leave it as is.

src/memory_provider.c
```diff
@@ -24,11 +25,70 @@ typedef struct umf_memory_provider_t {
...
+}
+
+void assignOpsExtDefaults(umf_memory_provider_ops_t *ops) {
+    if (!ops->ext.purge_lazy) {
+        ops->ext.purge_lazy = umfDefaultPurgeLazy;
+    }
+    if (!ops->ext.purge_force) {
+        ops->ext.purge_force = umfDefaultPurgeForce;
+    }
+    if (!ops->ext.allocation_split) {
```
igchor, line 70: So, can we check it here that either both `allocation_split` and `allocation_merge` is set or neither is? Also, we could mention that in the documentation.

vinser52: Done. I extended the validation function to check that both are set or both are NULL. Also, I extended corresponding comments in the `umf_memory_provider_ext_ops_t` structure.

bratpiorka (approved):

igchor (approved):

PR #325, opened 2024-03-12
Move assert-related stuff to utils_assert.h

ldorau:
These changes are needed by https://github.com/oneapi-src/unified-memory-framework/pull/320

### DIFF
--- a/src/utils/utils_assert.h
+++ b/src/utils/utils_assert.h
@@ -1,0 +1,67 @@
+/*
+ *
+ * Copyright (C) 2023-2024 Intel Corporation
+ *
+ * Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ *
+ */
+
+#ifndef UMF_ASSERT_H
+#define UMF_ASSERT_H 1
+
+#include <stdint.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#define NOFUNCTION                                                             \
+    do {                                                                       \
+    } while (0)
+#define VALGRIND_ANNOTATE_NEW_MEMORY(p, s) NOFUNCTION
+#define VALGRIND_HG_DRD_DISABLE_CHECKING(p, s) NOFUNCTION
+
+#ifdef NDEBUG
+#define ASSERT(x) NOFUNCTION
+#define ASSERTne(x, y) ASSERT(x != y)
+#else
+#define ASSERT(x)                                                              \
+    do {                                                                       \
+        if (!(x)) {                                                            \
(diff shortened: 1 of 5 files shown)


src/utils/utils_assert.h
```diff
@@ -0,0 +1,67 @@
...
+                    X, __LINE__);                                              \
+            abort();                                                           \
+        }                                                                      \
+    } while (0)
+#endif
+
+#define UMF_CHECK(condition, errorStatus)                                      \
+    do {                                                                       \
+        if (!(condition)) {                                                    \
+            fprintf(stderr, "UMF check failed: " #condition " in %s\n",        \
```
vinser52, line 57: The question not related to this PR.
Are we going to replace `fprintf` with logger API once we have it? Printing to `stderr` is a temporary solution, right?

src/utils/utils_common.h
```diff
@@ -116,8 +73,8 @@ static inline void align_ptr_size(void **ptr, size_t *size, size_t alignment) {
         s -= alignment - rest;
     }
 
-    ASSERT((p & (alignment - 1)) == 0);
-    ASSERT((s & (alignment - 1)) == 0);
+    assert((p & (alignment - 1)) == 0);
```
vinser52, line 76: Why did you replace UMF `ASSERT` with standard `assert`?

ldorau: So that utils_common.h is independent from utils_assert.h

src/utils/utils_assert.h
ldorau: I think so

src/utils/utils_common.h
vinser52: I understood that, but why do you want `utils_common.h` to be independent from `utils_assert.h`?

ldorau:
These changes are needed by https://github.com/oneapi-src/unified-memory-framework/pull/320

src/utils/utils_common.h
ldorau: It is needed by https://github.com/oneapi-src/unified-memory-framework/pull/320

vinser52: Sorry, but I still do not understand what is the exact issue...

vinser52: > @vinser52 we don't want to include fprintf into proxy lib because it wouldn't build. But I don't like this solution either - we shouldn't use standard assert

Now it is clear. Thank you for your explanations.  It should be fixed when we introduce a logger, right?

bratpiorka: right - with logger we wouldn't need to include stdlib which cause problems on windows

vinser52: So, maybe we can go with current PR, but create TODO comment and issue to track it. So that we can fix it when logger is merged.

vinser52 (approved):

ldorau:
Not needed any more

Issue #230, opened 2024-02-12
Expose memspace and policy API for managing host memory


igchor:
A memspace is an ordered collection of one or more memory targets. It can represent a certain class of memory sources like all NUMA nodes in the system or contain memory targets with different properties (e.g., latency).

A memory target represents a single memory source (e.g., NUMA node) and can have certain attributes (e.g., capacity).

Memspaces to expose (should be exposed as functions or a single function accepting an enum to avoid problems with exporting variables on windows):
- [ ] UMF_MEMSPACE_HOST
- [ ] UMF_MEMSPACE_HBW
- [ ] UMF_MEMSPACE_LOWEST_LATENCY
- [ ] UMF_MEMSPACE_HIGHEST_CAPACITY
- [ ] UMF_MEMSPACE_HIGHEST_BANDWITH

Memspace attributes to expose:
- [ ] Capacity
- [ ] Latency (as distance to local node / other nodes)
- [ ] Bandwidth
- [ ] Numa node id

Memspace management API:
- [ ] Filtering based on one (or more) of the attributes
- [ ] Sorting based on one (or more) of the attributes
- [ ] Iteration over memory targets within a memspace

All memspaces should be a part of libumf (not separate libraries).

Memspace, along with a policy, can be used to create pools and providers (umfPoolCreateFromMemspace and umfMemoryProviderCreateFromMemspace). Implementation of `umfPoolCreateFromMemspace` should select the best-suited pool based on memspace and policy type and availability of memory pools.

Policies to expose:
- [ ] UMF_MEMSPACE_POLICY_STRICT
- [ ] UMF_MEMSPACE_POLICY_PREFERRED
- [ ] UMF_MEMSPACE_POLICY_BLOCKED
- [ ] UMF_MEMSPACE_POLICY_INTERLEAVED

Optional:
- [ ] expose system-level memory metrics for memspaces (.e.g total available/free/allocated memory)

bratpiorka:
@igchor @vinser52 How we should name the default memspace (one that contains all available NUMA nodes)? "host" or "default"? Should we add an additional, dram-only memspace "REGULAR" (named as in the Memkind - "Regular means general purpose memory from the NUMA nodes containing CPUs")? What about HUGETLB and KMEM_DAX?

igchor:
[...]

PR #331, opened 2024-03-13
High capacity memspace

igchor:
This PR adds new 'high-capacity' memspace

Ref: https://github.com/oneapi-src/unified-memory-framework/issues/230

### Checklist

- [x] All tests pass locally

- [x] Extended the README/documentation
- [ ] Add qemu tests (possibly in the next PR)

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,9 +192,10 @@ TODO: Add general information about memspaces.
 Memspace backed by all available NUMA nodes discovered on the platform. Can be retrieved
 using umfMemspaceHostAllGet.
 
-##### Requirements
+#### Highest capacity memspace
 
-This feature relies on the OS memory provider.
+Memspace backed by all available NUMA nodes discovered on the platform sorted by capacity.
+Can be retrieved using umfMemspaceHighestCapacityGet.
 
 ## Contributions
 
--- a/src/topology.c
+++ b/src/topology.c
@@ -1,0 +1,45 @@
+/*
+ *
+ * Copyright (C) 2024 Intel Corporation
+ *
+ * Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ *
+ */
+
+#include <hwloc.h>
+
+#include "base_alloc_global.h"
+#include "utils_concurrency.h"
+
+static hwloc_topology_t topology = NULL;
+static UTIL_ONCE_FLAG topology_initialized = UTIL_ONCE_FLAG_INIT;
+
+void umfDestroyTopology(void) {
+    if (topology) {
+        hwloc_topology_destroy(topology);
+    }
+}
+
+static void umfCreateTopology(void) {
+    if (hwloc_topology_init(&topology)) {
+        fprintf(stderr, "Error: Failed to initialize topology\n");
+        topology = NULL;
+        return;
+    }
+
+    if (hwloc_topology_load(topology)) {
+        fprintf(stderr, "Error: Failed to initialize topology\n");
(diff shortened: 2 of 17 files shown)


test/memspaces/memspace_highest_capacity.cpp
```diff
@@ -0,0 +1,66 @@
...
+    for (auto nodeId : nodeIds) {
+        if (numa_node_size64(nodeId, nullptr) > maxCapacity) {
+            maxCapacityNode = nodeId;
+            maxCapacity = numa_node_size64(nodeId, nullptr);
+        }
+    }
+
+    // Confirm that the HighestCapacity memspace indeed has highest capacity
+    void *ptr;
+    auto ret = umfMemoryProviderAlloc(hProvider, alloc_size, 0, &ptr);
```
DamianDuy, line 53: Do you free this memory?

kswiecicki:
Wouldn't it be better to have a `large_cap` memspace that has memory targets with capacity above certain threshold, and let the user filter this memspace for the memspace with highest capacity/bandwidth/latency if they want?

kswiecicki:
> Wouldn't it be better to have a `large/high_cap` memspace that has memory targets with capacity above certain threshold, and let the user filter this memspace to get the highest capacity/bandwidth/latency if they want?

Nevermind, I just noticed that it's not a memspace with a single memory target, but a memspace with memory targets sorted by capacity property, which may be the better approach.

igchor:
> > Wouldn't it be better to have a `large/high_cap` memspace that has memory targets with capacity above certain threshold, and let the user filter this memspace to get the highest capacity/bandwidth/latency if they want?
> 
> Nevermind, I just noticed that it's not a memspace with a single memory target, but a memspace with memory targets sorted by capacity property, which may be the better approach.

Yeah, the idea is that this memspace contains all the nodes and the policy will dictate how they will be used (e.g. we only try to allocate from the first one or fallback to the 2nd, 3rd, etc. if there's not enough memory).

igchor:
The windows build fails because I haven't added checks to disable it on Windows/MacOS. I can add those but I would prefer to merge: https://github.com/oneapi-src/unified-memory-framework/pull/348 instead and then the checks won't be needed anymore.

README.md
```diff
@@ -194,6 +194,11 @@ TODO: Add general information about memspaces.
 Memspace backed by all available NUMA nodes discovered on the platform. Can be retrieved
 using umfMemspaceHostAllGet.
 
+#### Highest capacity memspace
```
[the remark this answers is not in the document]
igchor: I can do that, but perhaps in a separate PR to avoid conflicts (or once #344 is merged).

README.md
```diff
@@ -194,6 +194,11 @@ TODO: Add general information about memspaces.
...
+
+Memspace backed by all available NUMA nodes discovered on the platform sorted by capacity.
+Can be retrieved using umfMemspaceHighestCapacityGet.
+
 ##### Requirements
```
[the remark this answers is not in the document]
igchor: Right, done.

src/topology.c
```diff
@@ -0,0 +1,45 @@
+/*
+ *
+ * Copyright (C) 2023 Intel Corporation
```
bratpiorka, line 3: please set correct/bump dates

bratpiorka:
overall looks very good
other reviewers? @lukaszstolarczuk @kswiecicki ?

kswiecicki (approved):

src/topology.c
igchor: Right, done.

lukaszstolarczuk (commented):
👍LGMT, but I'm still counting on the docs 😉

PR #333, opened 2024-03-13
Add missing container job param

KFilipek:
### Checklist

- [ ] All tests pass locally


bratpiorka:
@KFilipek it this works, perhaps we should add "image:" param to all workflows to avoid similar problems in the future

KFilipek:
> @KFilipek if it this works, perhaps we should add "image:" param to all workflows to avoid similar problems in the future

Extending it to the basic workload can be problematic due to NUMA things. Some tests cannot allocate even 64B. I made changes, and we'll see if privileged will solve the problem.

.github/workflows/basic.yml
```diff
@@ -80,80 +80,84 @@ jobs:
...
             build_type: Release
             compiler: {c: gcc, cxx: g++}
             shared_library: 'OFF'
             os_provider: 'ON'
             pool_tracking: 'ON'
             level_zero_provider: 'OFF'
-    runs-on: ${{matrix.os}}
+    runs-on: ubuntu-${{matrix.os_version}}
+    container:
+      image: ubuntu:${{matrix.os_version}}
```
lukaszstolarczuk, line 145: we should use pinned version (due to security score), so we have to use hash version instead of a "nice" number

and at best we should use the full link to the image (e.g. `registry.hub.docker.com/library/ubuntu`)

lukaszstolarczuk (commented):
soo, should we use it eveywhere or just in these 2 workflows?

KFilipek:
Solved by PR from @PatKamin

Issue #338, opened 2024-03-13
Two disjointPool tests fail under memcheck and helgrind


igchor:
**The current status:** (see below the original issue)

The following two disjointPool tests fail under memcheck and helgrind:
1) disjointPoolTests/umfPoolTest.pow2AlignedAlloc/0
2) disjointPoolTests/umfPoolTest.multiThreadedpow2AlignedAlloc/0

They are temporarily disabled by the 2648539890172e26648d573d615acc7d800741f6 commit.

-------------------------------------------------

**The original issue:**
```
./umf_test-disjointPool - FAILED!
ERROR SUMMARY: 1076 errors from 106 contexts (suppressed: 0 from 0)

/home/runner/work/unified-memory-framework/unified-memory-framework/test/test_valgrind.sh: line 74:  4[27](https://github.com/igchor/unified-memory-framework/actions/runs/8269325299/job/22624345633#step:6:28)1 Aborted                 (core dumped) HWLOC_CPUID_PATH=./cpuid valgrind $OPTION $OPT_SUP --gen-suppressions=all $test > $LOG 2>&1
./umf_test-memspace_host_all (valgrind failed) - FAILED!
ERROR SUMMARY: 7 errors from 7 contexts (suppressed: 1 from 1)

./umf_test-provider_os_memory_config (valgrind failed) - FAILED!
ERROR SUMMARY: 3 errors from 3 contexts (suppressed: 3 from 3)
ERROR SUMMARY: 3 errors from 3 contexts (suppressed: 2 from 2)
ERROR SUMMARY: 3 errors from 3 contexts (suppressed: 1 from 1)
```

The disjointPool test failure seems to be related to newly added valgrind/asan annotations. The remaining tests seem to be incompatible with valgrind (mbind does not work under valgrind in provider_os_memory_config and memspace_host_all tests OOM which does not work with valgrind well as well)  - we should consider disabling them under memcheck.

ldorau:
The current status:

The following two disjointPool tests fail under memcheck and helgrind:
1) disjointPoolTests/umfPoolTest.pow2AlignedAlloc/0
2) disjointPoolTests/umfPoolTest.multiThreadedpow2AlignedAlloc/0

They are temporarily disabled by the 2648539890172e26648d573d615acc7d800741f6 commit.
Fix those failures and revert this commit.

ldorau:
The current repro:
[...]

PR #343, opened 2024-03-14
Skip tests incompatible with valgrind

ldorau:
Ref: #338

### DIFF
--- a/test/test_valgrind.sh
+++ b/test/test_valgrind.sh
@@ -79,7 +79,18 @@ for test in $(ls -1 ./umf_test-*); do
 	SUP="${WORKSPACE}/test/supp/${test}.supp"
 	OPT_SUP=""
 	[ -f ${SUP} ] && OPT_SUP="--suppressions=${SUP}"
-	HWLOC_CPUID_PATH=./cpuid valgrind $OPTION $OPT_SUP --gen-suppressions=all $test >$LOG 2>&1 || echo -n "(valgrind failed) "
+
+	# skip tests incompatible with valgrind
+	case $test in
+	./umf_test-memspace_host_all)
+		FILTER='--gtest_filter="-*allocsSpreadAcrossAllNumaNodes"'
+		;;
+	./umf_test-provider_os_memory_config)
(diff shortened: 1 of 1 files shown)


ldorau:
The NIghtly build: https://github.com/ldorau/unified-memory-framework/actions/runs/8279332960/job/22653507638

bratpiorka:
will this be a temporary solution? - I agree that we have to enable nightly but we should also investigate this more. Please add a TODO in the code :)

ldorau:
> will this be a temporary solution? - I agree that we have to enable nightly but we should also investigate this more. Please add a TODO in the code :)

No, it is not a temporary solution, see https://github.com/oneapi-src/unified-memory-framework/issues/338

test/test_valgrind.sh
```diff
@@ -79,7 +79,18 @@ for test in $(ls -1 ./umf_test-*); do
 	SUP="${WORKSPACE}/test/supp/${test}.supp"
 	OPT_SUP=""
 	[ -f ${SUP} ] && OPT_SUP="--suppressions=${SUP}"
-	HWLOC_CPUID_PATH=./cpuid valgrind $OPTION $OPT_SUP --gen-suppressions=all $test >$LOG 2>&1 || echo -n "(valgrind failed) "
+
+	# skip tests incompatible with valgrind
+	case $test in
+	./umf_test-memspace_host_all)
+		FILTER='--gtest_filter="-*allocsSpreadAcrossAllNumaNodes"'
```
lukaszstolarczuk, line 86: one thing to consider: do we have some functionality in these tests that still should be run under Valgrind? If so and we could split the test - we should split them.

ldorau: Yes, maybe ... It is an option to be considered.

ldorau: ... but not in this PR for sure :-D

bratpiorka: @ldorau please add an issue about that so we don't forget to do it at our free time

ldorau: Done: https://github.com/oneapi-src/unified-memory-framework/issues/347

PR #348, opened 2024-03-14
Enable OS memory provider building for all systems and remove the option to disable it

igchor:
### Description

The main reason we had a switch for OS memory provider building was to avoid having a dependency on hwloc for certain cases. Since we want to make hwloc a hard dependency there is no reason to keep this option anymore.

### Checklist

- [x] All tests pass locally


bratpiorka:
@igor - what about projects that are interested in the CPU ONLY? like oneCCL? We had to leave this option

README.md
```diff
@@ -89,7 +90,6 @@ List of options provided by CMake:
 | Name | Description | Values | Default |
 | - | - | - | - |
 | UMF_BUILD_SHARED_LIBRARY | Build UMF as shared library | ON/OFF | OFF |
-| UMF_BUILD_OS_MEMORY_PROVIDER | Build OS memory provider | ON/OFF | ON |
```
bratpiorka, line 92: this will not work with oneCCL or other GPU-only scenarios

bratpiorka (changes requested):

vinser52:
> @igor - what about projects that are interested in the CPU ONLY? like oneCCL? We had to leave this option

I do not understand this comment. Lets remember that oneAPI components will use UMF binaries that is part of oneAPI package. And for oneAPI UMF will be built with OS provider ON, L0 provider ON, etc. oneCCL is not going to built UMF from source.

bratpiorka:
> I do not understand this comment. Lets remember that oneAPI components will use UMF binaries that is part of oneAPI package. And for oneAPI UMF will be built with OS provider ON, L0 provider ON, etc. oneCCL is not going to built UMF from source.

At the moment, in my branch, the POC actually builds it from sources downloaded by FetchContent.... yes, I can change that

bratpiorka (approved):

vinser52:
> > I do not understand this comment. Lets remember that oneAPI components will use UMF binaries that is part of oneAPI package. And for oneAPI UMF will be built with OS provider ON, L0 provider ON, etc. oneCCL is not going to built UMF from source.
> 
> At the moment, in my branch, the POC actually builds it from sources downloaded by FetchContent.... yes, I can change that

Just to be clear, my comment was about oneAPI packaging. The idea is that UMF is a singleton shared library in oneAPI used by other component for memory management and interop.
Fetch content does not work because each UMF‘s client will have its own copy and interop capabilities (like IPC handles) would not work.

Regarding build option to ON/OFF OS memory provider I have no strong opinion. Some time ago we agreed that for L0 provider we need such option because it requires „non-trivial“ L0 dependency at a build time and by default we OFF it, while for oneAPI distribution we will ON it at our build/packaging system.
For OS provider I do not see the issue always have it ON. So I agree that this option could be removed. Another argument for that: it is required for mespace support.

kswiecicki (approved):

ldorau (approved):

PR #358, opened 2024-03-18
Enable building on MacOSX

ldorau:


CMakeLists.txt
```diff
@@ -193,42 +193,40 @@ if(UMF_BUILD_LIBUMF_POOL_JEMALLOC AND (LINUX OR WINDOWS))
     )
 endif()
 
-# set UMF_PROXY_LIB_ENABLED
-if(LINUX OR WINDOWS)
-    # TODO: enable the proxy library in the Debug build on Windows
-    if(WINDOWS AND NOT (CMAKE_BUILD_TYPE STREQUAL "Release"))
+# set UMF_PROXY_LIB_ENABLED TODO: enable the proxy library in the Debug build on
```
lukaszstolarczuk, line 196: please add an empty line (like `#  `) between the actual comment and the TODO for the TODO to be in a separate line

src/CMakeLists.txt
```diff
@@ -127,7 +127,7 @@ else()
 endif()
 
 # Currently, Level Zero provider is only built for Linux and Windows
```
lukaszstolarczuk, line 129: pls remove this comment

igchor (approved):

src/CMakeLists.txt
ldorau: Done

CMakeLists.txt
ldorau: Done

CMakeLists.txt
```diff
@@ -194,41 +194,39 @@ if(UMF_BUILD_LIBUMF_POOL_JEMALLOC AND (LINUX OR WINDOWS))
 endif()
 
 # set UMF_PROXY_LIB_ENABLED
-if(LINUX OR WINDOWS)
+if(WINDOWS AND NOT (CMAKE_BUILD_TYPE STREQUAL "Release"))
```
PatKamin, line 197: `CMAKE_BUILD_TYPE` is ignored on Windows and not needed to set. The type of build is determined only during the build phase on Windows, with the `--config` parameter. Try using `AND NOT $<CONFIG:Release>` instead.

lukaszstolarczuk: just a remark from me: this seems like a very good catch... but actually it worked somehow... interestingly, maybe CMake sets this variable internally for Windows...?

lukaszstolarczuk (approved):

CMakeLists.txt
ldorau: @PatKamin We set and use the `CMAKE_BUILD_TYPE` variable internally and make decisions depending on that in many places. so it does work correctly on Windows too. The only difference in case of Windows is that we have to add the `--config` option to the build command.

All Windows CI builds pass as you can see, so it works.

ldorau: @PatKamin Your proposal does not work:
https://github.com/ldorau/unified-memory-framework/actions/runs/8341093885/job/22826590124

PatKamin: > @PatKamin We set and use the `CMAKE_BUILD_TYPE` variable internally and make decisions depending on that in many places. so it does work correctly on Windows too. The only difference in case of Windows is that we have to add the `--config` option to the build command.
> 
> All Windows CI builds pass as you can see, so it works.

That's because in CI this variable is explicitly set during the configuration. What I mean is, this variable is not required, so it probably won't work when someone doesn't set this config variable explicitly on Windows locally (which is perfectly fine as only the build time param is required on Windows to build a repo with a given build type).

I think that this config time variable should be removed in CI for Windows builds and nothing should be based on this linux-specific config variable as it's not being set by default on Windows. And we shoudn't require Windows users to set this explicitly as all they need to do for a given type of build should be passing a type in the actual build command, not configuration one. That's a thing to do in some other PR, though.

PatKamin (approved):

CMakeLists.txt
ldorau: @PatKamin OK, so please resolve this issue if you approve this PR

PR #360, opened 2024-03-18
fix coverity issue in logger tests.

lplewa:
helper_log_init sets global "const char *" variable. if we exit scope when this global variable is set to ptr from std::string::c_str method, this ptr is invalid (std::string object is freed), and coverity do not like this.

### DIFF
--- a/test/utils/utils_log.cpp
+++ b/test/utils/utils_log.cpp
@@ -78,11 +78,13 @@ void helper_checkConfig(util_log_config_t *expected, util_log_config_t *is) {
 TEST_F(test, parseEnv_errors) {
     expected_message = "";
     loggerConfig = {0, 0, LOG_ERROR, LOG_ERROR, NULL};
+
     expect_fput_count = 0;
     expected_stream = stderr;
     util_log_config_t b = loggerConfig;
     helper_log_init(NULL);
     helper_checkConfig(&b, &loggerConfig);
+
     expect_fput_count = 1;
     helper_log_init("");
     helper_checkConfig(&b, &loggerConfig);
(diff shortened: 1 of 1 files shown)


KFilipek (approved):
I hope this will help.

lukaszstolarczuk (approved):

igchor:
Wouldn't it be easier to just make `env_variable`  a `std::string`? (and make `helper_log_init` accept a std::string as well). You won't need to worry about any leaks in that case. Also, you wouldn't need this explicit cast from `const char*` to `char*`.

If you do this, you would still be able to call `helper_log_init("something")` and this would just create std::string implicitly.

lplewa:
> Wouldn't it be easier to just make `env_variable` a `std::string`? (and make `helper_log_init` accept a std::string as well). You won't need to worry about any leaks in that case. Also, you wouldn't need this explicit cast from `const char*` to `char*`.
> 
> If you do this, you would still be able to call `helper_log_init("something")` and this would just create std::string implicitly.

Technically you can - but you have to add extra "hack" to handle test case if env variable is NULL (getenv() return NULL), so i think this solution is better.

igchor:
> > Wouldn't it be easier to just make `env_variable` a `std::string`? (and make `helper_log_init` accept a std::string as well). You won't need to worry about any leaks in that case. Also, you wouldn't need this explicit cast from `const char*` to `char*`.
> > If you do this, you would still be able to call `helper_log_init("something")` and this would just create std::string implicitly.
> 
> Technically you can - but you have to add extra "hack" to handle test case if env variable is NULL (getenv() return NULL), so i think this solution is better.

Ok, this is only a test anyway so we can leave it as is.

bratpiorka (approved):

PR #369, opened 2024-03-20
Always install examples

lukaszstolarczuk:
### Description
Examples are installed in system as a documentation, hence only the source files are installed, not binaries... so we can install examples all the time, not under an `if`.

A minor cleanup in the installation area of CMake is done as well (in a sep. commit).

### Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -379,6 +379,9 @@ endif()
 install(FILES ${CMAKE_SOURCE_DIR}/LICENSE.TXT
         DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/doc/${PROJECT_NAME}/")
 
+install(FILES examples/basic/basic.c
+        DESTINATION "${CMAKE_INSTALL_DOCDIR}/examples")
+
 # Add the list of installed targets to the install. This includes the namespace
 # which all installed targets will be prefixed with, e.g. for the headers target
 # users will depend on ${PROJECT_NAME}::headers.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -145,12 +145,6 @@ target_include_directories(
     umf_headers INTERFACE $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/include>
                           $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
 
-# Add the include directory and the headers target to the install.
-install(DIRECTORY "${PROJECT_SOURCE_DIR}/include/"
-        DESTINATION ${CMAKE_INSTALL_INCLUDEDIR})
-
-install(TARGETS umf_headers EXPORT ${PROJECT_NAME}-targets)
-
 if(WINDOWS)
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -372,10 +366,24 @@ if(UMF_FORMAT_CODE_STYLE)
...
+# Configure make install/uninstall and packages
+# --------------------------------------------------------------------------- #
 install(FILES ${CMAKE_SOURCE_DIR}/LICENSE.TXT
         DESTINATION "${CMAKE_INSTALL_DATAROOTDIR}/doc/${PROJECT_NAME}/")
 
+install(
+    DIRECTORY examples/
+    DESTINATION ${CMAKE_INSTALL_DOCDIR}/examples
+    FILES_MATCHING
+    PATTERN "*.c")
```
lukaszstolarczuk, line 379: one issue to solve: should we install only `c` files or should we also include directories, for each example.

On `main` we install `basic/basic.c`
On my PR we install `basic.c`

Which is expected? At the moment, this is the reason to fail the CI...

PatKamin: The current layout with subdirs will be most useful if there will be  examples with multiple files. However, if there is slight chance that we'll create examples with more than one file, then putting all source files into one directory would be a better solution.
I think we could put all examples in one directory now and eventually change this in the future, if needed.

bratpiorka: @lukaszstolarczuk I agree with @PatKamin

test/test_installation.py
```diff
@@ -114,16 +111,15 @@ def _create_match_list(self) -> List[str]:
...
-            examples.insert(0, "share/doc/unified-memory-framework/examples")
-            share.extend(examples)
+        examples_dir = Path(self.workspace_dir, "examples")
+        examples_dirs = [dir for dir in examples_dir.iterdir() if dir.is_dir()]
+        examples = [
+            f"share/doc/unified-memory-framework/examples/{file_path.name}"
+            for example_dir in examples_dirs
+            for file_path in example_dir.iterdir()
+        ]
+        examples.insert(0, "share/doc/unified-memory-framework/examples")
```
bratpiorka, line 121: could we just use share/doc/umf/examples path? this would be consistent with /include/umf/..

bratpiorka: same with lib/cmake/umf - I know this is not strictly related to this PR but it would be good to cleanup this. Please let me know if you could update this also or we'll do this in separate PR

PatKamin: Created #370 for this

bratpiorka: ok @PatKamin thanks!

CMakeLists.txt
lukaszstolarczuk: I couldn't quickly make it work with the `PATTERN` approach. I think we would have to use `glob` to iterate over all examples' dirs or something.

That's why, since we only have one example now, I made it simple like it was before - just copy a file to the proper place.

Please let me know, what you think.

PatKamin (approved):

bratpiorka (approved):

PR #374, opened 2024-03-21
add utils_fprintf and use it in print_numa_nodes()

lplewa:
The print_numa_nodes function is expected to print something to STDOUT (or STDERR in case of error), though its output should not be controlled by the logger. This patch adds a malloc-less fprintf function and converts print_numa_nodes() to use it instead of the system one.

### Checklist

- [x] All tests pass locally

### DIFF
--- a/src/provider/provider_os_memory.c
+++ b/src/provider/provider_os_memory.c
@@ -16,6 +16,7 @@
 
 #include "base_alloc_global.h"
 #include "provider_os_memory_internal.h"
+#include "utils_log.h"
 #include <umf.h>
 #include <umf/memory_provider_ops.h>
 #include <umf/providers/provider_os_memory.h>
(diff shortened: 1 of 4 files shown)


igchor:
@lplewa why "its output should not be controlled by the logger"? Do we really on this information being printed to stderr anywhere? I think we could safely use logger here.

bratpiorka:
I'm not sure why we can't use a logger. I saw that it is used only for printing debug info about OS Memory Provider allocations - I don't think that these logs (traces) should be handled differently than other log messages.

src/provider/provider_os_memory.c
```diff
@@ -318,24 +319,26 @@ static void print_numa_nodes(os_memory_provider_t *os_provider, void *addr,
 
     hwloc_bitmap_t nodeset = hwloc_bitmap_alloc();
     if (!nodeset) {
-        fprintf(stderr,
-                "cannot print assigned NUMA node due to allocation failure\n");
+        util_fprintf(
```
bratpiorka, line 322: why we don't use LOG_ERR ?

bratpiorka (changes requested):

PR #378, opened 2024-03-22
add Coarse Memory Provider

bratpiorka:
### Description

This PR adds the Coarse Memory Provider that could be used as a cache for upstream providers (like Level Zero Memory Provider) or handle user-provided buffers e.g. allocated using mmap.

### Checklist

- [x] All tests pass locally

### DIFF
--- a/src/provider/provider_coarse.c
+++ b/src/provider/provider_coarse.c
@@ -1,0 +1,1405 @@
+/*
+    Copyright (c) 2023 Intel Corporation
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+        http://www.apache.org/licenses/LICENSE-2.0
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+
+#include <assert.h>
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include <umf/providers/provider_coarse.h>
+
+#include "base_alloc_global.h"
+#include "memory_provider_internal.h"
+#include "ravl.h"
+#include "utils_concurrency.h"
+
+#ifndef BYTE
+#define BYTE unsigned char
+#endif
+
+static umf_result_t coarse_memory_provider_alloc(void *provider, size_t size,
+                                                 size_t alignment,
+                                                 void **resultPtr);
+
+static umf_result_t coarse_memory_provider_free(void *provider, void *ptr,
+                                                size_t bytes);
+
+static umf_result_t
+coarse_memory_provider_get_stats(void *provider,
+                                 coarse_memory_provider_stats_t *stats);
+
+static void ravl_cb_count(void *data, void *arg);
+static void ravl_cb_count_free(void *data, void *arg);
+
+typedef struct block_t {
+    size_t size;
+    BYTE *data;
+
+    // Origin is the element of the provider's upstream_alloc that contains the
+    // beginning of data in current block. Note that data address could be
+    // higher than the origin - this means, that the origin allocation
+    // covers current block only partially.
+    // If the size of the block is greater than the size of the allocation,
+    // it means that there are multiple allocations.
+    // Note that provider's upstream_alloc doesn't use "origin" and "used" fields.
+    struct block_t *origin;
+    bool used;
+
+    struct block_t *next;
+    struct block_t *prev;
+
+    // Node in the list of free blocks of the same size pointing to this block.
+    // The list is located in the (coarse_provider->free_blocks) RAVL tree.
+    struct ravl_free_blocks_elem_t *free_list_ptr;
+} block_t;
+
+// A general node in a RAVL tree.
+// 1) coarse_provider->all_blocks RAVL tree (tree of all blocks - sorted by an address of data):
+//    key   - pointer (block_t->data) to the beginning of the block data
+//    value - pointer (block_t) to the block of the allocation
+// 2) coarse_provider->free_blocks RAVL tree (tree of free blocks - sorted by a size of data):
+//    key   - size of the allocation (block_t->size)
+//    value - pointer (ravl_free_blocks_head_t) to the head of the list of free blocks of the same size
+typedef struct ravl_data_t {
+    uintptr_t key;
+    void *value;
+} ravl_data_t;
+
+// The head of the list of free blocks of the same size,
+// so there is a separate mutex for each size.
+typedef struct ravl_free_blocks_head_t {
+    struct ravl_free_blocks_elem_t *head;
+    struct os_mutex_t mutex;
+} ravl_free_blocks_head_t;
+
+// The node of the list of free blocks of the same size
+typedef struct ravl_free_blocks_elem_t {
+    struct block_t *block;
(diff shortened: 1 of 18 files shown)


src/provider/provider_coarse.c
```diff
@@ -0,0 +1,1405 @@
...
+    assert(coarse_provider->used_size == 0);
+
+    ravl_delete(coarse_provider->upstream_alloc);
+    ravl_delete(coarse_provider->all_blocks);
+    ravl_delete(coarse_provider->free_blocks);
+
+    umf_ba_global_free(coarse_provider);
+}
+
+static umf_result_t coarse_memory_provider_alloc(void *provider, size_t size,
```
igchor, line 956: this function is too long, please split it into multiple function

src/provider/provider_coarse.c
```diff
@@ -0,0 +1,1405 @@
...
+    ravl_data_t *rdata = ravl_data(rnode);
+    assert(rdata);
+    block = rdata->value;
+    assert(block);
+    // make sure it is really the head
+    assert(block->prev == NULL);
+    return block;
+}
+
+static bool debug_check(coarse_memory_provider_t *provider) {
```
igchor, line 562: please split this into multiple checks/functions

test/disjointCoarseMallocPool.cpp
```diff
@@ -0,0 +1,610 @@
...
+
+    // test 1
+
+    size_t s1 = 74659 * KB;
+    size_t s2 = 8206 * KB;
+
+    size_t max_alloc_size = 0;
+
+    // s1
+    for (int j = 0; j < 2; j++) {
```
igchor, line 367: magic values: `2`, `6`, please use some consts

test/disjointCoarseMallocPool.cpp
```diff
@@ -0,0 +1,610 @@
...
+
+                ASSERT_EQ(((unsigned char *)ptr)[0], alloc_check_val);
+                ASSERT_EQ(((unsigned char *)ptr)[size - 1], alloc_check_val);
+                ((unsigned char *)ptr)[0] = 0;
+                ((unsigned char *)ptr)[size - 1] = 0;
+
+                umf_result_t ret = umfPoolFree(pool, ptr);
+                ASSERT_EQ(ret, UMF_RESULT_SUCCESS);
+
+                allocs.erase((*it));
```
igchor, line 593: `erase(it)` would probably be faster

test/disjointCoarseMallocPool.cpp
```diff
@@ -0,0 +1,610 @@
...
+
+    return UMF_RESULT_SUCCESS;
+}
+
+static const char *mallocName(void *provider) {
+    (void)provider;
+    return "malloc";
+}
+
+struct umf_memory_provider_ops_t UMF_MALLOC_MEMORY_PROVIDER_OPS = {
```
igchor, line 116: You don't have to implement this provider - we already have one: https://github.com/oneapi-src/unified-memory-framework/blob/main/test/common/provider.hpp#L113

If you need to customize the `provider_malloc`  you can also just create a struct that inherits from `provider_base_t ` and overrides the functions that you need.

test/disjointCoarseMallocPool.cpp
```diff
@@ -0,0 +1,610 @@
...
+    // each alloc would be done few times
+    std::vector<size_t> counts = {1, 3, 4, 8, 9, 11};
+    std::uniform_int_distribution<int> counts_dist(0, (int)(counts.size() - 1));
+
+    // action to take will be random
+    // alloc = <0, .5), free = <.5, 1)
+    std::uniform_real_distribution<float> actions_dist(0, 1);
+
+    std::set<alloc_ptr_size> allocs;
+    for (size_t i = 0; i < 100; i++) {
```
igchor, line 540: This whole loop could be simplified (and made more readable) if you'd implement the logic for choosing size/action, etc. in a separate function/class:

```
allocation_state state; // holds std::set<alloc_ptr_size>
for (size_t i = 0; i < 100; i++) {
   auto action = generator.get_action();
   if (action.is_alloc()) {
       state.do_alloc(action.count, ...);
   } else {
       state.do_free(action.acount, ...);
   }
}
```

igchor: Also, this way you could reuse this for other tests (even multithreaded perhaps)

test/disjointCoarseMallocPool.cpp
```diff
@@ -0,0 +1,610 @@
...
+    const size_t init_buffer_size = 200 * MB;
+
+    // Preallocate some memory
+    void *buf = malloc(init_buffer_size);
+    memset(buf, 0, init_buffer_size);
+    ASSERT_NE(buf, nullptr);
+
+    const unsigned char alloc_check_val = 11;
+    coarse_memory_provider_params_t coarse_memory_provider_params = {
+        NULL,  // upstream_memory_provider
```
igchor, line 494: Can't you just set each field separately as you do with disjoint_pool_params below instead of using comments?

test/disjointCoarseMallocPool.cpp
```diff
@@ -0,0 +1,610 @@
...
+};
+
+TEST_F(test, disjointCoarseMMapPool_random) {
+
+    const size_t KB = 1024;
+    const size_t MB = 1024 * KB;
+    const size_t init_buffer_size = 200 * MB;
+
+    // Preallocate some memory
+    void *buf = malloc(init_buffer_size);
```
igchor, line 488: Please avoid raw allocations in C++ and use unique_ptr or a container (e.g. vector or string). If this test is run under valgrind and some assert triggers or an exception is thrown buf will not be freed and valgrind will report a memory leak (which makes it harder to debug).

igchor: Ideally we would use unique_ptr for pools and providers as well but that would require more refactoring I guess.

igchor (commented):
I haven't looked at the provider implementation yet but regarding tests, I think it would be good to:
- add some multithreaded tests
- add asan/memcheck annotations to coarse provider ( and enable them at least when we're testing coarse-provider directly)

lplewa:
This PR is huuuuugeeeee. also it seems there is a lot small random fixes which do not belong to this change. 

To make reviewer life easier please:
- Split small fixes to separate commits and PRs. (things like typos, small refactoring etc)
- Try to split this PR to smaller one (i.e Ravl can be added in separate PR with a few tests to prof it's working)

Also having more smaller self-contained commits, helps in case of debuging/bisecting,

test/disjointCoarseMallocPool.cpp
ldorau: Fixed in https://github.com/oneapi-src/unified-memory-framework/pull/715

test/disjointCoarseMallocPool.cpp
ldorau: Fixed in https://github.com/oneapi-src/unified-memory-framework/pull/715

test/disjointCoarseMallocPool.cpp
ldorau: Fixed in https://github.com/oneapi-src/unified-memory-framework/pull/715

test/disjointCoarseMallocPool.cpp
ldorau: Fixed in https://github.com/oneapi-src/unified-memory-framework/pull/715

src/provider/provider_coarse.c
ldorau: Fixed in https://github.com/oneapi-src/unified-memory-framework/pull/715

src/provider/provider_coarse.c
ldorau: Fixed in https://github.com/oneapi-src/unified-memory-framework/pull/715

Issue #362, opened 2024-03-19
CMAKE_BUILD_TYPE is used for some CMake checks on Windows


PatKamin:
## Environment Information

- UMF version (hash commit or a tag): [cd152ea](https://github.com/oneapi-src/unified-memory-framework/commit/cd152ea7b336118eef5b765e687cc469d8f04fe4)                         
- OS(es) version(s): Windows                                         
- kernel version(s):  n/a                                        
- compiler, libraries, and other related tools version(s):  MSVC compilers  

## Please provide a reproduction of the bug:
Run a CMake configuration with the proxy lib enabled and build UMF:
```
cmake -B build -D UMF_BUILD_LIBUMF_POOL_SCALABLE=ON
```
The `CMAKE_BUILD_TYPE` variable is set internally to `Release`:
```
-- No build type selected (CMAKE_BUILD_TYPE), defaulting to Release
-- The CXX compiler identification is MSVC 19.34.31935.0
```
Now all internal CMake checks doesn't work well when we decide to build in a Debug mode:
```
cmake --build build --config Debug
```

## How often bug is revealed:

always

## Actual behavior:
Ie. the proxy library is built whether it shouldn't for Debug mode on Windows.

## Expected behavior:
All intetrnal CMake decisions should be based on a config generators which work well both on Linux and Windows:
Replace ie.:
```
if (CMAKE_BUILD_TYPE STREQUAL "Release")
```
with:
```
if ($<CONFIG:Release>)
```

PR #386, opened 2024-03-25
Disable CMake build type checks on Windows

PatKamin:
### Description
Do not rely on the `CMAKE_BUILD_TYPE` configuration variable in Windows builds. This variable is ignored by `MSVC`, the build type is determined during the build time based on the `--config` parameter.

There seems to be no way for disabling the proxy library build based on the used config, so this change provides just notification messages on all Windows builds and builds proxy library always.

Fixes: #362 

### Checklist

- [x] All tests pass locally

### DIFF
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -226,7 +226,6 @@ jobs:
         -B ${{env.BUILD_DIR}}
         ${{matrix.toolset}}
         -DCMAKE_PREFIX_PATH="${{env.VCPKG_PATH}}"
-        -DCMAKE_BUILD_TYPE=${{matrix.build_type}}
         -DCMAKE_C_COMPILER=${{matrix.compiler.c}}
         -DCMAKE_CXX_COMPILER=${{matrix.compiler.cxx}}
         -DUMF_BUILD_SHARED_LIBRARY=${{matrix.shared_library}}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,31 +41,51 @@ set(UMF_PROXY_LIB_BASED_ON_POOL
 set_property(CACHE UMF_PROXY_LIB_BASED_ON_POOL
              PROPERTY STRINGS ${KNOWN_PROXY_LIB_POOLS})
 
-set(KNOWN_BUILD_TYPES Release Debug RelWithDebInfo MinSizeRel)
-string(REPLACE ";" " " KNOWN_BUILD_TYPES_STR "${KNOWN_BUILD_TYPES}")
-
-if(NOT CMAKE_BUILD_TYPE)
-    message(
-        STATUS
-            "No build type selected (CMAKE_BUILD_TYPE), defaulting to Release")
-    set(CMAKE_BUILD_TYPE "Release")
+if(${CMAKE_SYSTEM_NAME} MATCHES "Linux")
+    set(LINUX TRUE)
+    set(OS_NAME "linux")
+elseif(${CMAKE_SYSTEM_NAME} MATCHES "Windows")
+    set(WINDOWS TRUE)
+    set(OS_NAME "windows")
+elseif(${CMAKE_SYSTEM_NAME} MATCHES "Darwin")
+    set(MACOSX TRUE)
+    set(OS_NAME "macosx")
 else()
-    message(STATUS "CMAKE_BUILD_TYPE: ${CMAKE_BUILD_TYPE}")
-    if(NOT CMAKE_BUILD_TYPE IN_LIST KNOWN_BUILD_TYPES)
-        message(
-            WARNING
-                "Unusual build type was set (${CMAKE_BUILD_TYPE}), please make sure it is a correct one. "
-                "The following ones are supported by default: ${KNOWN_BUILD_TYPES_STR}."
-        )
-    endif()
+    message(FATAL_ERROR "Unknown OS type")
 endif()
 
-set(CMAKE_BUILD_TYPE
-    "${CMAKE_BUILD_TYPE}"
-    CACHE STRING
-          "Choose the type of build, options are: ${KNOWN_BUILD_TYPES_STR} ..."
-          FORCE)
-set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS ${KNOWN_BUILD_TYPES})
+# This build type check is not possible on Windows when CMAKE_BUILD_TYPE is not
+# set, because in this case the build type is determined after a CMake
+# configuration is done (at the build time)
+if(NOT WINDOWS)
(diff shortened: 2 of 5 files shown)


CMakeLists.txt
```diff
@@ -190,13 +197,16 @@ if(UMF_BUILD_LIBUMF_POOL_JEMALLOC)
 endif()
 
 # set UMF_PROXY_LIB_ENABLED
-if(WINDOWS AND NOT (CMAKE_BUILD_TYPE STREQUAL "Release"))
-    # TODO: enable the proxy library in the Debug build on Windows
```
lukaszstolarczuk, line 194: isn't this TODO still valid...?

CMakeLists.txt
```diff
@@ -190,13 +197,16 @@ if(UMF_BUILD_LIBUMF_POOL_JEMALLOC)
...
+    # the CMake configuration step. Therefore, this message is printed in all
+    # MSVC builds.
     message(
         STATUS
-            "Disabling the proxy library, because it is supported only in the Release build on Windows (CMAKE_BUILD_TYPE = ${CMAKE_BUILD_TYPE})"
+            "The proxy library will be built, however it is supported only in the Release build on Windows"
     )
-elseif(UMF_PROXY_LIB_BASED_ON_POOL STREQUAL SCALABLE)
+endif()
+if(UMF_PROXY_LIB_BASED_ON_POOL STREQUAL SCALABLE)
```
lukaszstolarczuk, line 209: shouldn't this be still `elseif`?

.github/workflows/basic.yml
```diff
@@ -226,7 +226,6 @@ jobs:
         -B ${{env.BUILD_DIR}}
         ${{matrix.toolset}}
         -DCMAKE_PREFIX_PATH="${{env.VCPKG_PATH}}"
-        -DCMAKE_BUILD_TYPE=${{matrix.build_type}}
```
lukaszstolarczuk, line 229: I guess there might be more of such entries (e.g. in benchmarks)

CMakeLists.txt
```diff
@@ -41,31 +41,51 @@ set(UMF_PROXY_LIB_BASED_ON_POOL
...
 else()
-    message(STATUS "CMAKE_BUILD_TYPE: ${CMAKE_BUILD_TYPE}")
-    if(NOT CMAKE_BUILD_TYPE IN_LIST KNOWN_BUILD_TYPES)
+    message(FATAL_ERROR "Unknown OS type")
+endif()
+
+# This build type check is not possible on Windows when CMAKE_BUILD_TYPE is not
+# set, because in this case the build type is determined after a CMake
+# configuration is done (at the build time)
+if(NOT WINDOWS)
```
lukaszstolarczuk, line 60: Perhaps for Windows we could write something - maybe simplified version of the comment above...?

lukaszstolarczuk (commented):
hmm.. TBH I expected some changes with the syntax you described in the issue (e.g. `if ($<CONFIG:Release>)`) - I guess they weren't needed after all...?

CMakeLists.txt
```diff
@@ -41,31 +41,51 @@ set(UMF_PROXY_LIB_BASED_ON_POOL
...
+    set(OS_NAME "macosx")
 else()
-    message(STATUS "CMAKE_BUILD_TYPE: ${CMAKE_BUILD_TYPE}")
-    if(NOT CMAKE_BUILD_TYPE IN_LIST KNOWN_BUILD_TYPES)
+    message(FATAL_ERROR "Unknown OS type")
+endif()
+
+# This build type check is not possible on Windows when CMAKE_BUILD_TYPE is not
+# set, because in this case the build type is determined after a CMake
+# configuration is done (at the build time)
```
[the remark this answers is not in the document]
PatKamin: It's automatically handled in MSVC:
```
error MSB8013: This project doesn't contain the Configuration and Platform combination of WRONG_BUILD_TYPE|x64
```

.github/workflows/basic.yml
PatKamin: I've verified that it's the only place where this variable is used for Windows tests in the CI.

CMakeLists.txt
PatKamin: See my comment above.

CMakeLists.txt
PatKamin: You're right, reverted this change.

CMakeLists.txt
PatKamin: No, because on Windows we build the proxy lib, too. Now on all configurations.

benchmark/CMakeLists.txt
```diff
@@ -2,7 +2,10 @@
 # Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
-if(CMAKE_BUILD_TYPE STREQUAL "Debug")
+# In MSVC builds, there is no way to determine the actual build type during the
+# CMake configuration step. Therefore, this message is printed in all MSVC
+# builds.
+if(WINDOWS)
     message(WARNING "The benchmarks SHOULD NOT be run in the Debug build type!")
```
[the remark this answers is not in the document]
PatKamin: Changed to a STATUS type of a message. Also restored the previous condition as this message is relevant also on Linux/OSX

CMakeLists.txt
lukaszstolarczuk: ahh, right

CMakeLists.txt
lukaszstolarczuk: I meant, we could print some message to user :)

.github/workflows/basic.yml
lukaszstolarczuk: I think, it is still in benchmarks:

https://github.com/oneapi-src/unified-memory-framework/actions/runs/8421036255/job/23057332971#step:6:54

PatKamin: Oh, thanks!

CMakeLists.txt
PatKamin: With MSVC we can't print other message than always print a list of available build types. I'm not sure if such an information should be added, though.

PatKamin:
> hmm.. TBH I expected some changes with the syntax you described in the issue (e.g. `if ($<CONFIG:Release>)`) - I guess they weren't needed after all...?

I've tried quite a few approaches of using such generators ie. to disable the proxy library creation based on config. However, generator expressions can be used only in some cases, in some of CMake functions (which I didn't find helpful in this case). Originally I thought there are no such limitations.

However, we still cannot rely on CMAKE_BUILD_TYPE on Windows in the CI and this has to be changed somehow.

lukaszstolarczuk (approved):

bratpiorka (approved):

Issue #387, opened 2024-03-25
Build all libs as shared object when libumf is built that way


bratpiorka:
Build all libraries as shared objects when libumf is built this way.
It makes no sense to dynamically link only part of the libraries.

igchor:
I'm not so sure if this is the best approach. For example, I think we prefer to keep using disjoint_pool as a static library in UR - at least for now.

Eventually, I think we should make disjoint_pool a part of libumf itself (it would require rewriting it in C), tbbmalloc could be a part of libumf as well. I'm still not sure what we should do with jemalloc though since there are known problems with dlopen.

bratpiorka:
But why do we want to link disjoint pool as static and umf as shared object?

bratpiorka:
hm, but from a user's perspective (say - me), when I decide to use dynamic libraries, the reason is e.g. to use the latest libraries installed on the system. So if I use the latest libumf, I am also interested in using the latest libdisjoint_pool.

@vinser52 @pbalcer what do you think?

ldorau:
@bratpiorka Is it really a bug?

ldorau:
> hm, but from a user's perspective (say - me), when I decide to use dynamic libraries, the reason is e.g. to use the latest libraries installed on the system. So if I use the latest libumf, I am also interested in using the latest libdisjoint_pool.
> 

@vinser52 @pbalcer what do you think?

bratpiorka:
> @bratpiorka Is it really a bug?

this is inconsistency

ldorau:
> > @bratpiorka Is it really a bug?
> 
> this is inconsistency

Yes, I agree.
Frankly I do not understand it either.

igchor:
> What about scalable pool implementation? As I remember in the past we agreed that we do not need to introduce separate library for a scalable pool. libumf at runtime will try to dlopen tbbmalloc library and if it is found than scalable pool is available.

Yes, and I belive there's nothing stopping us from doing it right now. I wanted to create a PR with this change but I didn't have time yet.

PR #388, opened 2024-03-25
Build pool libs as shared if umf is shared also

bratpiorka:
### Description

Build memory pool libs (disjoint, scalable, etc.) as shared if the main libumf is also built as shared lib.

Ref: #387

### Checklist

- [ ] All tests pass locally

### DIFF
--- a/benchmark/CMakeLists.txt
+++ b/benchmark/CMakeLists.txt
@@ -102,13 +102,13 @@ endif()
 add_umf_benchmark(
     NAME ubench
     SRCS ubench.c
-    LIBS ${LIBS_OPTIONAL}
+    LIBS ${LIBS_OPTIONAL} umf_utils
     LIBDIRS ${LIB_DIRS})
 
 if(UMF_BUILD_BENCHMARKS_MT)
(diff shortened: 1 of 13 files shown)


vinser52:
> Build memory pool libs (disjoint, scalable, etc.) as shared if the main libumf is also built as shared lib.

What is the reason for that?

bratpiorka:
> > Build memory pool libs (disjoint, scalable, etc.) as shared if the main libumf is also built as shared lib.
> 
> What is the reason for that?

I think we should either provide all libs as shared objects or none. Please think about what we will install in oneAPI toolkit. The discussion is here: https://github.com/oneapi-src/unified-memory-framework/issues/387

vinser52:
> I think we should either provide all libs as shared objects or none. Please think about what we will install in oneAPI toolkit. The discussion is here: https://github.com/oneapi-src/unified-memory-framework/issues/387

Ok, I have replied to the discussion thread #387.

bratpiorka:
closing this as we agreed in #387. that we don't want this change

PR #389, opened 2024-03-25
Add bandit scanner workflow

lukaszstolarczuk:
### Description
Run [Bandit tool](https://github.com/PyCQA/bandit) as part of the CI.

Enabling this workflow require adding a few nosec "skips", as I believe they don't apply for us - in test and dev. scripts.

### Checklist
- [x] All tests pass locally

### DIFF
--- a/scripts/generate_docs.py
+++ b/scripts/generate_docs.py
@@ -1,5 +1,5 @@
 """
- Copyright (C) 2023 Intel Corporation
+ Copyright (C) 2023-2024 Intel Corporation
  
  Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
  SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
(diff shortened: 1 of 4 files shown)


scripts/generate_docs.py
```diff
@@ -1,13 +1,13 @@
...
+ Copyright (C) 2023-2024 Intel Corporation
  
  Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
  SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 """
 
 from pathlib import Path
 from shutil import rmtree
-import subprocess
+import subprocess   # nosec B404
```
PatKamin, line 10: This one and `B603` could be placed in some bandit config file, if possible. It seems like they will always be needed when using `subprocess.run()`.

PatKamin (approved):

scripts/generate_docs.py
lukaszstolarczuk: I guess I didn't want to introduce "global" skips. In-place skips are only placed in dev scripts/tests, after looking into the issues.

PR #390, opened 2024-03-25
Fix highest capacity memspace with OS provider integration.

igchor:
The integration was not correct - umfMemoryProviderCreateFromMemspace was passing nodemask with all bits set to OS provider (with MODE_BIND) which resulted in memory being allocated from the arbitrary node. This wasn't caught by the test due to wrong way of checking allocation location.

Fix this, by changing the implementation of numa_memory_target to only pass a single node to OS provider for now. This should be changed once we have implemented memspace policies and/or extend OS provider to possibly support setting priority for nodes from which we want to allocate.

### Checklist

- [x] All tests pass locally

### DIFF
--- a/test/common/numa_helpers.h
+++ b/test/common/numa_helpers.h
@@ -1,0 +1,32 @@
+// Copyright (C) 2024 Intel Corporation
+// Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#ifndef UMF_NUMA_HELPERS_H
+#define UMF_NUMA_HELPERS_H 1
+
+#include <numa.h>
+#include <numaif.h>
+#include <stdint.h>
+#include <stdio.h>
+
+#include "test_helpers.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// returns the node where page starting at 'ptr' resides
+int getNumaNodeByPtr(void *ptr) {
+    int nodeId;
+    int retm =
+        get_mempolicy(&nodeId, nullptr, 0, ptr, MPOL_F_ADDR | MPOL_F_NODE);
+    UT_ASSERTeq(retm, 0);
+    return nodeId;
+}
+
+#ifdef __cplusplus
+}
+#endif
+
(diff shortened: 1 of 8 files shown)


bratpiorka (approved):

test/memspaces/memspace_highest_capacity.cpp
```diff
@@ -40,28 +41,27 @@ TEST_F(memspaceHighestCapacityProviderTest, highestCapacityVerify) {
...
     long long maxCapacity = 0;
-    int maxCapacityNode = -1;
+    std::vector<int> maxCapacityNodes{};
     for (auto nodeId : nodeIds) {
         if (numa_node_size64(nodeId, nullptr) > maxCapacity) {
-            maxCapacityNode = nodeId;
             maxCapacity = numa_node_size64(nodeId, nullptr);
         }
+
+        if (numa_node_size64(nodeId, nullptr) == maxCapacity) {
```
lukaszstolarczuk, line 50: shouldn't this `if` be in a sep. `for`...?

test/common/numa_helpers.h
```diff
@@ -0,0 +1,33 @@
+// Copyright (C) 2023-2024 Intel Corporation
+// Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// This file contains helpers for tests for UMF pool API
```
lukaszstolarczuk, line 4: I guess this could be updated (you can add an empty line between the license and the description)

test/memspaces/memspace_highest_capacity.cpp
igchor: Yes, of course, done.

lukaszstolarczuk (commented):
👍 lgtm

test/common/numa_helpers.h
```diff
@@ -0,0 +1,32 @@
+// Copyright (C) 2023-2024 Intel Corporation
```
kswiecicki, line 1: Only `2024`.

kswiecicki (approved):

test/common/numa_helpers.h
igchor: Done.

Issue #399, opened 2024-04-02
The proxy library tests are built in the Debug build on Windows in case of the shared version of the UMF library.


ldorau:
The `umf-proxy_lib_basic` and `umf-proxy_lib_memoryPool` tests are built in the Debug build on Windows in case of the shared version of the UMF library.

The proxy library is not supported in the Debug build on Windows, so the `umf-proxy_lib_basic` and `umf-proxy_lib_memoryPool` tests  should not be built. They are built now and they fail or hang, see: https://github.com/oneapi-src/unified-memory-framework/actions/runs/8520522407/job/23337072750?pr=398

Ref: https://github.com/oneapi-src/unified-memory-framework/pull/398

## Environment Information

- UMF version (hash commit or a tag): 22a30de6fa8442f91b4b30132335306b52632adf
- OS(es) version(s): Windows, Debug build only
- `UMF_BUILD_SHARED_LIBRARY` == `ON`

PR #401, opened 2024-04-02
Disable proxy lib tests on Windows Debug

PatKamin:
### Description
Fixes: #399

This PR disables the two tests which use proxy lib when ctest is run for a Debug config on Windows:
```
ctest -C Debug
```

### Checklist

- [x] All tests pass locally

### DIFF
--- a/.cmake-format
+++ b/.cmake-format
@@ -26,7 +26,8 @@ with section("parse"):
       'kwargs': {
         'NAME': '*', 
         'SRCS': '*', 
-        'LIBS': '*'}},
+        'LIBS': '*',
+        'CFGS': '*'}},
     'add_umf_library': {
       "pargs": 0,
       "flags": [],
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -30,7 +30,7 @@ function(add_umf_test)
     # * SRCS - source files
     # * LIBS - libraries to be linked with
     set(oneValueArgs NAME)
-    set(multiValueArgs SRCS LIBS)
+    set(multiValueArgs SRCS LIBS CFGS)
     cmake_parse_arguments(
         ARG
         ""
@@ -73,6 +73,7 @@ function(add_umf_test)
     add_test(
         NAME ${TEST_NAME}
         COMMAND ${TEST_TARGET_NAME}
+        CONFIGURATIONS ${ARG_CFGS}
         WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
 
     set_tests_properties(${TEST_NAME} PROPERTIES LABELS "umf")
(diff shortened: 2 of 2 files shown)


ldorau (changes requested):
Now you have to add the `-C` option to all `ctest` calls, even on Linux! Because this
```sh
$ ctest
```
will NOT run the proxy library tests, only that:
```sh
$ ctest -C Release
```
will do.

ldorau (changes requested):
What if a user wants to use a non-standard (self made up) build type?

ldorau:
> > What if a user wants to use a non-standard (self made up) build type?
> 
> I think we shouldn't bother with this right now. I hope we'll fix the problems with debug proxy lib in the future and then we could remove this code.

OK

bratpiorka:
> Now you have to add the `-C` option to all `ctest` calls, even on Linux! Because this
> 
> ```shell
> $ ctest
> ```
> 
> will NOT run the proxy library tests, only that:
> 
> ```shell
> $ ctest -C Release
> ```
> 
> will do.

but I see the CI pass?

PatKamin:
> > Now you have to add the `-C` option to all `ctest` calls, even on Linux! Because this
> > ```shell
> > $ ctest
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > will NOT run the proxy library tests, only that:
> > ```shell
> > $ ctest -C Release
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > will do.
> 
> but I see the CI pass?

Yes, in CI we explicitly use the `-C` parameter.

PatKamin:
> Now you have to add the `-C` option to all `ctest` calls, even on Linux! Because this
> 
> ```shell
> $ ctest
> ```
> 
> will NOT run the proxy library tests, only that:
> 
> ```shell
> $ ctest -C Release
> ```
> 
> will do.

Now the behavior is the same regardless of the platform, one have to add the `-C` option.

ldorau (changes requested):
The `-C` option have to be added to all ctest calls **in CI**, also on Linux!

ldorau:
The proxy library tests are not run on Linux now!:
After adding:
https://github.com/ldorau/unified-memory-framework/commit/684d635ac81da1705d51ad28a7e72527c41e5012
all tests passed:
https://github.com/ldorau/unified-memory-framework/actions/runs/8534974992

ldorau:
> > > Now you have to add the `-C` option to all `ctest` calls, even on Linux! Because this
> > > ```shell
> > > $ ctest
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > will NOT run the proxy library tests, only that:
> > > ```shell
> > > $ ctest -C Release
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > will do.
> > 
> > 
> > but I see the CI pass?
> 
> Yes, in CI we explicitly use the `-C` parameter.

But not in Linux jobs! - it has to be added there.

PatKamin:
> > > > Now you have to add the `-C` option to all `ctest` calls, even on Linux! Because this
> > > > ```shell
> > > > $ ctest
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > will NOT run the proxy library tests, only that:
> > > > ```shell
> > > > $ ctest -C Release
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > will do.
> > > 
> > > 
> > > but I see the CI pass?
> > 
> > 
> > Yes, in CI we explicitly use the `-C` parameter.
> 
> But not in Linux jobs! - it has to be added there.

Thanks, I hadn't noticed that. I've removed the list of configs from the linux builds so there should be no impact of this change on linux tests now.

.cmake-format
```diff
@@ -26,7 +26,8 @@ with section("parse"):
       'kwargs': {
         'NAME': '*', 
         'SRCS': '*', 
-        'LIBS': '*'}},
+        'LIBS': '*',
+        'CFGS': '*',}},
```
ldorau, line 30: Redundant `,`. It should be `'CFGS': '*'}},`

ldorau (changes requested):

.cmake-format
PatKamin: Removed

ldorau (approved):

lukaszstolarczuk (approved):

PR #409, opened 2024-04-04
Add black formatter for Python

lukaszstolarczuk:
### Description
- Add `black` formatter support (via CMake custom commands)
- New CI check enabled
- Apply formatting on python source files
- Add a minor fixes in clang/cmake formatting commands

### Checklist
- [x] Extended the README/documentation

### DIFF
--- a/.github/workflows/pr_push.yml
+++ b/.github/workflows/pr_push.yml
@@ -106,10 +106,10 @@ jobs:
         -DUMF_BUILD_LIBUMF_POOL_SCALABLE=OFF
         -DUMF_BUILD_LIBUMF_POOL_JEMALLOC=OFF
 
-    - name: Check clang-format
+    - name: Check C/C++ formatting
       run: cmake --build build --target clang-format-check
 
-    - name: Check cmake-format
+    - name: Check CMake formatting
       run: cmake --build build --target cmake-format-check
 
     - name: Check Python formatting
--- a/.github/workflows/pr_push.yml
+++ b/.github/workflows/pr_push.yml
@@ -94,7 +94,7 @@ jobs:
     - name: Install apt packages
       run: |
         sudo apt-get update
-        sudo apt-get install -y cmake clang-format-15 cmake-format libhwloc-dev
+        sudo apt-get install -y black cmake clang-format-15 cmake-format libhwloc-dev
 
     - name: Configure CMake
       run: >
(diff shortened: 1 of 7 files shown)


.github/workflows/pr_push.yml
```diff
@@ -94,7 +94,7 @@ jobs:
     - name: Install apt packages
       run: |
         sudo apt-get update
-        sudo apt-get install -y cmake clang-format-15 cmake-format libhwloc-dev
+        sudo apt-get install -y black cmake clang-format-15 cmake-format libhwloc-dev
```
PatKamin, line 97: Can you use a pip package instead? I'm not sure if find_package() will work for pip modules, but this way it will be more convenient to use `black` on Windows.

lukaszstolarczuk: hmm I'm not sure what do you mean... In this job, as you can see, we use `apt` for all tools. This is because this job is only run on Linux.

Anyway, black is added as a python dep. in our `requirements.txt` file, so it should work well on all OSes. I've verified it on my machine with pip installation.

Please let me know if you propose to change something

bratpiorka (approved):

.github/workflows/pr_push.yml
ldorau: @PatKamin This job (`Coding style`) is run only on Linux Ubuntu

ldorau (approved):

CMakeLists.txt
```diff
@@ -365,24 +372,52 @@ if(UMF_FORMAT_CODE_STYLE)
...
-    if(CLANG_FORMAT AND CMAKE_FORMAT)
+    if(BLACK)
+        # black should maintain backward compatibility, we don't have to require
+        # a specific version
+        get_program_version_major_minor(${BLACK} BLACK_VERSION)
+        message(STATUS "Found black: ${BLACK} (version: ${BLACK_VERSION})")
+
+        message(
+            STATUS
+                "Adding 'black-format-check' and 'black-format-apply' make targets"
```
PatKamin, line 383: `make targets` - does it mean that MSVC targets won't be created?

.github/workflows/pr_push.yml
```diff
@@ -112,6 +112,9 @@ jobs:
     - name: Check cmake-format
       run: cmake --build build --target cmake-format-check
 
+    - name: Check formatting with black
```
PatKamin, line 115: nit: I think these steps' names could refer to the language of files formatted, like: C/C++ formatting check etc.

lukaszstolarczuk: voila

CMakeLists.txt
lukaszstolarczuk: Removed all `make` refs in these messages

PatKamin (approved):

PR #410, opened 2024-04-04
Remove assert from umf_ba_destroy

igchor:
This assert would be triggered every time a memory pool or memory provider is leaked. Assert should not be used to verify the correctness of a user code but only to make sure the state of UMF is correct.  Also, in certain cases, leaking a pool/provider might be done on purpose.

### Checklist

- [ ] All tests pass locally

### DIFF
--- a/src/base_alloc/base_alloc.c
+++ b/src/base_alloc/base_alloc.c
@@ -291,7 +291,6 @@ void umf_ba_destroy(umf_ba_pool_t *pool) {
     if (pool->metadata.n_allocs) {
         LOG_ERR("umf_ba_destroy(): pool->metadata.n_allocs = %zu",
                 pool->metadata.n_allocs);
-        assert(pool->metadata.n_allocs == 0);
     }
 #endif /* NDEBUG */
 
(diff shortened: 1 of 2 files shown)


lukaszstolarczuk (commented):
👍

ldorau (approved):
Could you remove also the same assert from the linear allocator?
https://github.com/oneapi-src/unified-memory-framework/blob/025c0a8161c54d77108eb7020f68d2eb0890a92e/src/base_alloc/base_alloc_linear.c#L255

lplewa (approved):

bratpiorka:
Agree with this change but we should check for errors in logs in our CI - I will add a task for this

igchor:
> Agree with this change but we should check for errors in logs in our CI - I will add a task for this

+1 We could also add some option (env var?) to fail on each LOG_ERROR. We could set that option for tests only.

bratpiorka:
@igchor ?

> [unified-memory-framework/src/base_alloc/base_alloc_linear.c](https://github.com/oneapi-src/unified-memory-framework/blob/025c0a8161c54d77108eb7020f68d2eb0890a92e/src/base_alloc/base_alloc_linear.c#L255)
> 
> Line 255

igchor:
> @igchor ?
> 
> > [unified-memory-framework/src/base_alloc/base_alloc_linear.c](https://github.com/oneapi-src/unified-memory-framework/blob/025c0a8161c54d77108eb7020f68d2eb0890a92e/src/base_alloc/base_alloc_linear.c#L255)
> > Line 255

Good catch, done.

ldorau (approved):

PR #419, opened 2024-04-08
check for null dereference in trackingOpenIpcHandle

bratpiorka:
### Description

Check for null dereference in trackingOpenIpcHandle - this is a fix for coverity issue https://scan3.scan.coverity.com/#/project-view/65229/15141?selectedIssue=453142

### Checklist
- [x] All tests pass locally
- [x] Logger (with debug/info/... messages) is used

### DIFF
--- a/src/provider/provider_tracking.c
+++ b/src/provider/provider_tracking.c
@@ -586,17 +586,23 @@ static umf_result_t trackingOpenIpcHandle(void *provider, void *providerIpcData,
         (umf_tracking_memory_provider_t *)provider;
     umf_result_t ret = UMF_RESULT_SUCCESS;
 
+    if (p->hUpstream == NULL) {
+        LOG_ERR("tracking open ipc handle failed: no upstream provider");
+        return UMF_RESULT_ERROR_INVALID_ARGUMENT;
+    }
+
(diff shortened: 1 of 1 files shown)


src/provider/provider_tracking.c
```diff
@@ -586,17 +586,23 @@ static umf_result_t trackingOpenIpcHandle(void *provider, void *providerIpcData,
         (umf_tracking_memory_provider_t *)provider;
     umf_result_t ret = UMF_RESULT_SUCCESS;
 
+    if (p->hUpstream == NULL) {
```
vinser52, line 589: How it is possible? If it happens it indicates error in UMF code, right? If so, I think it should be just assert.

I agree that `&& p->hUpstream` checks below should be just removed.

bratpiorka: This is just a fix for coverity. Agree that this should never happen but I think it is better to add this code that add a false-positive

ldorau: IMHO if it is really a false-positive it is much better to ignore it in Coverity than add a redundant code.

ldorau: @bratpiorka I propose sth like that https://github.com/oneapi-src/unified-memory-framework/pull/420 instead and ignore this issue in Coverity.

bratpiorka:
this would be fixed by #420

PR #420, opened 2024-04-09
Verify trackers's parameters in trackingInitialize()

ldorau:
### Description

Verify trackers's parameters in trackingInitialize().

### Checklist
- [x] All tests pass locally

### DIFF
--- a/src/provider/provider_tracking.c
+++ b/src/provider/provider_tracking.c
@@ -147,9 +147,7 @@ static umf_result_t trackingAlloc(void *hProvider, size_t size,
         (umf_tracking_memory_provider_t *)hProvider;
     umf_result_t ret = UMF_RESULT_SUCCESS;
 
-    if (!p->hUpstream) {
-        return UMF_RESULT_ERROR_INVALID_ARGUMENT;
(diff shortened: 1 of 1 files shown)


src/provider/provider_tracking.c
```diff
@@ -147,10 +147,6 @@ static umf_result_t trackingAlloc(void *hProvider, size_t size,
         (umf_tracking_memory_provider_t *)hProvider;
     umf_result_t ret = UMF_RESULT_SUCCESS;
 
-    if (!p->hUpstream) {
```
bratpiorka, line 150: why do you remove that?

ldorau: Because this is checked once at the beginning in `trackingInitialize()`. We can have an assert here instead.

ldorau: assert added instead

bratpiorka: ok looks good for me

bratpiorka (approved):

lukaszstolarczuk (approved):

Issue #421, opened 2024-04-09
Unspecified hwloc required version in the README.md


PatKamin:
## Environment Information

- UMF version (hash commit or a tag):  bd2b34c                      
- OS(es) version(s):    All                                      
- kernel version(s):    All                                      
- compiler, libraries, and other related tools version(s):   All 

## How often bug is revealed:

always

## Actual behavior:
In the main README file there is no requirement for the exact or minimum version of the hwloc library. However, there is a hardcoded file name off this library suggesting version 1.5 required on Windows:
![image](https://github.com/oneapi-src/unified-memory-framework/assets/35933418/468177c0-406b-48ad-b930-cce6ee012fd5)

## Expected behavior:
Update the Readme file with the exact or minimum required version of hwloc. Find out which version should it be, [1.5 is very old](https://www.open-mpi.org/software/hwloc/ancient/) so requiring exactly this version is confusing.

Add a CMake check for the version required both on Linux and Windows.

## Additional information about Priority and Help Requested:

Are you willing to submit a pull request with a proposed change? No  

Requested priority: Medium

lukaszstolarczuk:
I think it will be updated in: #408

PatKamin:
> I think it will be updated in: #408

That will fix the issue, certainly!

PR #408, opened 2024-04-04
Add memspace "highest bandwidth"

kswiecicki:
### Description
This memspace contains memory targets ordered by the highest bandwidth value (targets higher in the ordering are prioritised).
There is also an alternative way to assign nodes to this memspace by setting `UMF_MEMSPACE_HIGHEST_BANDWIDTH` environment variable.

### TODO
- [x] Add tests for "highest bandwidth" memspace from `UMF_MEMSPACE_HIGHEST_BANDWIDTH`
- [x] Verify functionality on QEMU runner

### Checklist

- [x] Code compiles without errors locally
- [x] All tests pass locally
- [x] CI workflows execute properly

- [x] CI workflows, not executed per PR (e.g. Nightly), execute properly 
- [x] New tests added, especially if they will fail without my changes
- [x] Extended the README/documentation
- [x] All newly added source files have a license
- [x] All newly added source files are referenced in CMake files
- [x] Logger (with debug/info/... messages) is used

PR #426, opened 2024-04-10
Fix looking for hwloc.dll

ldorau:
### Description

Fix looking for hwloc.dll - do not assume a specific version of HWLOC.

Ref: #421

### Checklist
- [x] All tests pass locally


ldorau:
It will be fixed by #408

PR #428, opened 2024-04-10
Share multithread utils between tests and benchmarks

vinser52:
### Description
Refactor tests and benchmark code to share multithread utils between benchmarks and tests.

### Checklist

- [x] All tests pass locally

### DIFF
--- a/benchmark/CMakeLists.txt
+++ b/benchmark/CMakeLists.txt
@@ -42,8 +42,10 @@ function(add_umf_benchmark)
         LIBS ${BENCH_LIBS})
 
     target_include_directories(
-        ${BENCH_NAME} PRIVATE ${UMF_CMAKE_SOURCE_DIR}/include
-                              ${UMF_CMAKE_SOURCE_DIR}/src/utils)
+        ${BENCH_NAME}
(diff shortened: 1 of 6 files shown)


vinser52:
@igchor Could you please explain why in this code block (https://github.com/oneapi-src/unified-memory-framework/blob/5f4652f9a2f8c13732839e051f94f549a8c685a3/benchmark/multithread.hpp#L114-L121) the barrier is called twice (before and after the benchmark body)?

igchor (approved):

igchor:
> @igchor Could you please explain why in this code block
> 
> https://github.com/oneapi-src/unified-memory-framework/blob/5f4652f9a2f8c13732839e051f94f549a8c685a3/benchmark/multithread.hpp#L114-L121
> 
> the `syncthreads` barrier is called twice (before and after the benchmark body)?

I think we can remove this. This file was copied from libpmemobj-cpp and as far as I remember, we tested some datastructures there that behaved differently depending on how many threads were active - so we wanted to make sure that all threads were alive until the end of the benchmarks.

bratpiorka (approved):

vinser52:
> I think we can remove this.

Done

PR #429, opened 2024-04-10
Add support for MAP_SHARED for Linux only

ldorau:
### Description

1) Add support for `MAP_SHARED` to OS memory provider
2) Add support for `MAP_SHARED` to the proxy library
3) Implement the IPC hooks in OS memory provider
4) Add a basic shared memory (SHM) IPC test (`ipc_os_prov`) using only the OS memory provider API (not using the API from `ipc.h`).

### Checklist
- [x] All tests pass locally

### DIFF
--- a/README.md
+++ b/README.md
@@ -233,6 +233,10 @@ In case of Linux it can be done without any code changes using the `LD_PRELOAD` 
 $ LD_PRELOAD=/usr/lib/libumf_proxy.so myprogram
 ```
 
+The memory used by the proxy memory allocator is mmap'ed:
+1) with the `MAP_PRIVATE` flag by default or
+2) with the `MAP_SHARED` flag only if the `UMF_PROXY` environment variable contains the `page.disposition=shared` string.
+
 #### Windows
 
 In case of Windows it requires:
--- a/README.md
+++ b/README.md
@@ -138,6 +138,13 @@ More detailed documentation is available here: https://oneapi-src.github.io/unif
 #### OS memory provider
 
 A memory provider that provides memory from an operating system.
+It supports two types of memory mappings
+1) private memory mapping (`UMF_MEM_MAP_PRIVATE`)
+2) shared memory mapping (`UMF_MEM_MAP_SHARED` - supported on Linux only yet)
+
+If the shared memory mapping is used then an anonymous file descriptor for memory mapping is created using:
+1) `memfd_secret()` syscall - (if it is implemented and) if the `UMF_MEM_FD_FUNC` environment variable does not contain the "memfd_create" string or
+2) `memfd_create()` syscall - otherwise (and if it is implemented).
 
 ##### Requirements
 
--- a/include/umf/providers/provider_os_memory.h
+++ b/include/umf/providers/provider_os_memory.h
@@ -29,6 +29,12 @@ typedef enum umf_mem_protection_flags_t {
     /// @endcond
 } umf_mem_protection_flags_t;
 
+/// @brief Memory visibility mode
+typedef enum umf_memory_visibility_t {
+    UMF_MEM_MAP_PRIVATE = 1, ///< private memory mapping
+    UMF_MEM_MAP_SHARED, ///< shared memory mapping (supported on Linux only)
+} umf_memory_visibility_t;
+
 /// @brief Memory binding mode
 /// Specifies how memory is bound to NUMA nodes on systems that support NUMA.
 /// Not every mode is supported on every system.
@@ -61,6 +67,8 @@ typedef enum umf_numa_mode_t {
 typedef struct umf_os_memory_provider_params_t {
     /// Combination of 'umf_mem_protection_flags_t' flags
     unsigned protection;
+    /// memory visibility mode
+    umf_memory_visibility_t visibility;
 
     // NUMA config
     /// ordered list of numa nodes
@@ -91,6 +99,7 @@ static inline umf_os_memory_provider_params_t
 umfOsMemoryProviderParamsDefault(void) {
     umf_os_memory_provider_params_t params = {
         UMF_PROTECTION_READ | UMF_PROTECTION_WRITE, /* protection */
+        UMF_MEM_MAP_PRIVATE,                        /* visibility mode */
         NULL,                                       /* numa_list */
         0,                                          /* numa_list_len */
         UMF_NUMA_MODE_DEFAULT,                      /* numa_mode */
--- a/src/provider/provider_os_memory.c
+++ b/src/provider/provider_os_memory.c
@@ -15,6 +15,7 @@
 #include <string.h>
 
 #include "base_alloc_global.h"
+#include "critnib.h"
 #include "provider_os_memory_internal.h"
 #include "utils_log.h"
 
@@ -26,6 +27,16 @@
 
 typedef struct os_memory_provider_t {
     unsigned protection; // combination of OS-specific protection flags
+    unsigned visibility; // memory visibility mode
+    int fd;              // file descriptor for memory mapping
+    size_t size_fd;      // size of file used for memory mapping
+    size_t max_size_fd;  // maximum size of file used for memory mapping
+    // A critnib map storing (ptr, fd_offset + 1) pairs. We add 1 to fd_offset
+    // in order to be able to store fd_offset equal 0, because
+    // critnib_get() returns value or NULL, so a value cannot equal 0.
+    // It is needed mainly in the get_ipc_handle and open_ipc_handle hooks
+    // to mmap a specific part of a file.
+    critnib *fd_offset_map;
 
     // NUMA config
     hwloc_bitmap_t nodeset;
@@ -191,6 +202,34 @@ static umf_result_t translate_params(umf_os_memory_provider_params_t *in_params,
         return result;
     }
 
+    result = os_translate_mem_visibility_flag(in_params->visibility,
+                                              &provider->visibility);
+    if (result != UMF_RESULT_SUCCESS) {
+        LOG_ERR("incorrect memory visibility flag: %u", in_params->visibility);
+        return result;
+    }
+
+    provider->fd = os_create_anonymous_fd(provider->visibility);
+    if (provider->fd == -1) {
+        LOG_ERR(
+            "creating an anonymous file descriptor for memory mapping failed");
+        return UMF_RESULT_ERROR_UNKNOWN;
+    }
+
+    provider->size_fd = 0; // will be increased during each allocation
+    provider->max_size_fd = get_max_file_size();
+
+    if (provider->fd > 0) {
+        int ret = os_set_file_size(provider->fd, provider->max_size_fd);
+        if (ret) {
+            LOG_ERR("setting file size %zu failed", provider->max_size_fd);
+            return UMF_RESULT_ERROR_INVALID_ARGUMENT;
+        }
+    }
+
+    LOG_DEBUG("size of the memory mapped file set to %zu",
+              provider->max_size_fd);
+
     // NUMA config
     int emptyNodeset = in_params->numa_list_len == 0;
     result = translate_numa_mode(in_params->numa_mode, emptyNodeset,
@@ -217,6 +256,14 @@ static umf_result_t os_initialize(void *params, void **provider) {
 
     umf_os_memory_provider_params_t *in_params =
         (umf_os_memory_provider_params_t *)params;
+
+    if (in_params->visibility == UMF_MEM_MAP_SHARED &&
+        in_params->numa_mode != UMF_NUMA_MODE_DEFAULT) {
+        LOG_ERR("Unsupported NUMA mode for the UMF_MEM_MAP_SHARED memory "
+                "visibility mode (only the UMF_NUMA_MODE_DEFAULT is supported "
+                "for now)");
+        return UMF_RESULT_ERROR_NOT_SUPPORTED;
+    }
 
     os_memory_provider_t *os_provider =
         umf_ba_global_alloc(sizeof(os_memory_provider_t));
@@ -261,10 +308,19 @@ static umf_result_t os_initialize(void *params, void **provider) {
         }
     }
 
+    os_provider->fd_offset_map = critnib_new();
+    if (!os_provider->fd_offset_map) {
+        LOG_ERR("creating file descriptor offset map failed");
+        ret = UMF_RESULT_ERROR_OUT_OF_HOST_MEMORY;
+        goto err_free_nodeset_str_buf;
+    }
+
     *provider = os_provider;
 
     return UMF_RESULT_SUCCESS;
 
+err_free_nodeset_str_buf:
+    umf_ba_global_free(os_provider->nodeset_str_buf);
 err_destroy_hwloc_topology:
     hwloc_topology_destroy(os_provider->topo);
 err_free_os_provider:
@@ -279,6 +335,9 @@ static void os_finalize(void *provider) {
     }
 
     os_memory_provider_t *os_provider = provider;
+
+    critnib_delete(os_provider->fd_offset_map);
+
     if (os_provider->nodeset_str_buf) {
         umf_ba_global_free(os_provider->nodeset_str_buf);
     }
@@ -334,7 +393,9 @@ static inline void assert_is_page_aligned(uintptr_t ptr, size_t page_size) {
 }
 
 static int os_mmap_aligned(void *hint_addr, size_t length, size_t alignment,
-                           size_t page_size, int prot, void **out_addr) {
+                           size_t page_size, int prot, int flag, int fd,
+                           size_t max_fd_size, void **out_addr,
+                           size_t *fd_size) {
     assert(out_addr);
 
     size_t extended_length = length;
@@ -347,8 +408,21 @@ static int os_mmap_aligned(void *hint_addr, size_t length, size_t alignment,
         extended_length += alignment;
     }
 
-    void *ptr = os_mmap(hint_addr, extended_length, prot);
+    size_t fd_offset = 0;
+
+    if (fd > 0) {
+        if (*fd_size + extended_length > max_fd_size) {
+            LOG_ERR("cannot grow a file size beyond %zu", max_fd_size);
+            return -1;
+        }
+
+        fd_offset = *fd_size;
+        *fd_size += extended_length;
+    }
+
+    void *ptr = os_mmap(hint_addr, extended_length, prot, flag, fd, fd_offset);
     if (ptr == NULL) {
+        LOG_PDEBUG("memory mapping failed");
         return -1;
     }
 
@@ -414,15 +488,17 @@ static umf_result_t os_alloc(void *provider, size_t size, size_t alignment,
         return UMF_RESULT_ERROR_INVALID_ARGUMENT;
     }
 
-    int protection = os_provider->protection;
+    size_t fd_offset = os_provider->size_fd; // needed for critnib_insert()
 
     void *addr = NULL;
     errno = 0;
-    ret = os_mmap_aligned(NULL, size, alignment, page_size, protection, &addr);
+    ret = os_mmap_aligned(NULL, size, alignment, page_size,
+                          os_provider->protection, os_provider->visibility,
+                          os_provider->fd, os_provider->max_size_fd, &addr,
+                          &os_provider->size_fd);
     if (ret) {
         os_store_last_native_error(UMF_OS_RESULT_ERROR_ALLOC_FAILED, errno);
         LOG_PERR("memory allocation failed");
-
         return UMF_RESULT_ERROR_MEMORY_PROVIDER_SPECIFIC;
     }
 
@@ -432,7 +508,6 @@ static umf_result_t os_alloc(void *provider, size_t size, size_t alignment,
         LOG_ERR("allocated address 0x%llx is not aligned to %zu (0x%zx) "
                 "bytes",
                 (unsigned long long)addr, alignment, alignment);
-
         goto err_unmap;
     }
 
@@ -463,6 +538,18 @@ static umf_result_t os_alloc(void *provider, size_t size, size_t alignment,
         }
     }
 
+    if (os_provider->fd > 0) {
+        // store (fd_offset + 1) to be able to store fd_offset == 0
+        ret =
+            critnib_insert(os_provider->fd_offset_map, (uintptr_t)addr,
+                           (void *)(uintptr_t)(fd_offset + 1), 0 /* update */);
+        if (ret) {
+            LOG_ERR("os_alloc(): inserting a value to the file descriptor "
+                    "offset map failed (addr=%p, offset=%zu)",
+                    addr, fd_offset);
+        }
+    }
+
     *resultPtr = addr;
 
     return UMF_RESULT_SUCCESS;
@@ -475,6 +562,12 @@ err_unmap:
 static umf_result_t os_free(void *provider, void *ptr, size_t size) {
     if (provider == NULL || ptr == NULL) {
         return UMF_RESULT_ERROR_INVALID_ARGUMENT;
+    }
+
+    os_memory_provider_t *os_provider = (os_memory_provider_t *)provider;
+
+    if (os_provider->fd > 0) {
+        critnib_remove(os_provider->fd_offset_map, (uintptr_t)ptr);
     }
 
     errno = 0;
@@ -581,23 +674,59 @@ static const char *os_get_name(void *provider) {
     return "OS";
 }
 
+// This function is supposed to be thread-safe, so it should NOT be called concurrently
+// with os_allocation_merge() with the same pointer.
 static umf_result_t os_allocation_split(void *provider, void *ptr,
                                         size_t totalSize, size_t firstSize) {
-    (void)provider;
-    (void)ptr;
     (void)totalSize;
-    (void)firstSize;
-    // nop
+
+    os_memory_provider_t *os_provider = (os_memory_provider_t *)provider;
+    if (os_provider->fd <= 0) {
+        return UMF_RESULT_SUCCESS;
+    }
+
+    void *value = critnib_get(os_provider->fd_offset_map, (uintptr_t)ptr);
+    if (value == NULL) {
+        LOG_ERR("os_allocation_split(): getting a value from the file "
+                "descriptor offset map failed (addr=%p)",
+                ptr);
+        return UMF_RESULT_ERROR_UNKNOWN;
+    } else {
+        uintptr_t new_key = (uintptr_t)ptr + firstSize;
+        void *new_value = (void *)((uintptr_t)value + firstSize);
+        int ret = critnib_insert(os_provider->fd_offset_map, new_key, new_value,
+                                 0 /* update */);
+        if (ret) {
+            LOG_ERR("os_allocation_split(): inserting a value to the file "
+                    "descriptor offset map failed (addr=%p, offset=%zu)",
+                    (void *)new_key, (size_t)new_value - 1);
+            return UMF_RESULT_ERROR_UNKNOWN;
+        }
+    }
+
     return UMF_RESULT_SUCCESS;
 }
 
+// It should NOT be called concurrently with os_allocation_split() with the same pointer.
 static umf_result_t os_allocation_merge(void *provider, void *lowPtr,
                                         void *highPtr, size_t totalSize) {
-    (void)provider;
     (void)lowPtr;
-    (void)highPtr;
     (void)totalSize;
-    // nop
+
+    os_memory_provider_t *os_provider = (os_memory_provider_t *)provider;
+    if (os_provider->fd <= 0) {
+        return UMF_RESULT_SUCCESS;
+    }
+
+    void *value =
+        critnib_remove(os_provider->fd_offset_map, (uintptr_t)highPtr);
+    if (value == NULL) {
+        LOG_ERR("os_allocation_merge(): removing a value from the file "
+                "descriptor offset map failed (addr=%p)",
+                highPtr);
+        return UMF_RESULT_ERROR_UNKNOWN;
+    }
+
     return UMF_RESULT_SUCCESS;
 }
 
--- a/src/provider/provider_os_memory.c
+++ b/src/provider/provider_os_memory.c
@@ -730,6 +730,117 @@ static umf_result_t os_allocation_merge(void *provider, void *lowPtr,
     return UMF_RESULT_SUCCESS;
 }
 
+typedef struct os_ipc_data_t {
+    int pid;
+    int fd;
+    size_t fd_offset;
+    size_t size;
+} os_ipc_data_t;
+
+static umf_result_t os_get_ipc_handle_size(void *provider, size_t *size) {
+    (void)provider; // unused
+
+    if (size == NULL) {
+        return UMF_RESULT_ERROR_INVALID_ARGUMENT;
+    }
+
+    *size = sizeof(os_ipc_data_t);
+    return UMF_RESULT_SUCCESS;
+}
+
+static umf_result_t os_get_ipc_handle(void *provider, const void *ptr,
+                                      size_t size, void *providerIpcData) {
+    if (provider == NULL || ptr == NULL || providerIpcData == NULL) {
+        return UMF_RESULT_ERROR_INVALID_ARGUMENT;
+    }
+
+    os_memory_provider_t *os_provider = (os_memory_provider_t *)provider;
+    if (os_provider->fd <= 0) {
+        return UMF_RESULT_ERROR_INVALID_ARGUMENT;
+    }
+
+    void *value = critnib_get(os_provider->fd_offset_map, (uintptr_t)ptr);
+    if (value == NULL) {
+        LOG_ERR("os_get_ipc_handle(): getting a value from the IPC cache "
+                "failed (addr=%p)",
+                ptr);
+        return UMF_RESULT_ERROR_INVALID_ARGUMENT;
+    }
+
+    os_ipc_data_t *os_ipc_data = (os_ipc_data_t *)providerIpcData;
+    os_ipc_data->pid = os_getpid();
+    os_ipc_data->fd = os_provider->fd;
+    os_ipc_data->fd_offset = (size_t)value - 1;
+    os_ipc_data->size = size;
+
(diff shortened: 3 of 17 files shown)


include/umf/providers/provider_os_memory.h
```diff
@@ -61,6 +71,8 @@ typedef enum umf_numa_mode_t {
 typedef struct umf_os_memory_provider_params_t {
     /// Combination of 'umf_mem_protection_flags_t' flags
     unsigned protection;
+    /// mmap flag (Linux only, ignored on Windows)
```
igchor, line 74: Why is this ignored on windows? Don't we want to have the same interface on linux and windows for shared memory eventually?

igchor (changes requested):
Is the main purpose of this PR to enable shared memory for forked processes?

I removed the SHARED flag a while back because it was not clear how to allow other processes to use this memory (https://github.com/oneapi-src/unified-memory-framework/pull/236) other than by using fork. I think this requires more research and we should probably consider using IPC (@vinser52 did we have any discussion about this on any of the issues?)

Also, @ldorau please extend existing `bind` tests to test MAP_SHARED as well. I'm not sure whether hwloc_set_area_membind will work with MAP_SHARED. According to the documentation of mbind: `The specified policy will be ignored for any MAP_SHARED mappings
       in the specified memory range.  Rather the pages will be
       allocated according to the memory policy of the thread that
       caused the page to be allocated.`

bratpiorka:
@igchor this PR is a part of the requirements for CAL library

vinser52:
How is this shared region supposed to be used?

As Igor said, one of the use cases for shared memory on a host is IPC.
Before introducing support of shared memory I would suggest extending IPC tests: we need to create IPC tests that can be run with memory providers that support IPC (today only L0 supports it). having such test allows us to validate the changes introduced in this PR.

Regarding CAL, do we already have exact requirements?

bratpiorka:
> How is this shared region supposed to be used?
> Regarding CAL, do we already have exact requirements?

I asked CAL team about that. I thought this should be as simple as adding MAP_SHARED flag but now I think it should be changed to a draft version until we get the requirements.

include/umf/providers/provider_os_memory.h
bratpiorka: @ldorau we had to mention this in our documentation

bratpiorka: @ldorau @igchor Also I think we could return NOT_SUPPORTED for windows

README.md
```diff
@@ -212,6 +212,10 @@ In case of Linux it can be done without any code changes using the `LD_PRELOAD`
 $ LD_PRELOAD=/usr/lib/libumf_proxy.so myprogram
 ```
 
+The memory used by the proxy memory allocator is mmap'ed:
+1) with the `MAP_PRIVATE` flag by default or
+2) with the `MAP_SHARED` flag only if the `UMF_PROXY` environment variable contains the `page.disposition=shared` string.
```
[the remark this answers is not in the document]
ldorau: Add in the documentation (`README.md`) of "OS memory provider".

include/umf/providers/provider_os_memory.h
ldorau: Done. It will return `UMF_RESULT_ERROR_INVALID_ARGUMENT` for Windows

igchor: Why not NOT_SUPPORTED as @bratpiorka suggested? I think it fits better for cases where something is only supported on one platform.

ldorau: OK, I will do that. It requires the following changes - please review: https://github.com/oneapi-src/unified-memory-framework/pull/438 and https://github.com/oneapi-src/unified-memory-framework/pull/439

README.md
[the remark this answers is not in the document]
ldorau: @bratpiorka Done

include/umf/providers/provider_os_memory.h
ldorau: Done

ldorau: @igchor Done

examples/basic/ipc_shared_memory_server.c
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    fprintf(stderr, "[server] Sent response to the client\n");
+
+    ret = 0; // SUCCESS
+
+err_CloseIPCHandle:
+    // we do NOT know the exact size of the remote shared memory! - TODO: the API should be changed
+    umf_result = umfMemoryProviderCloseIPCHandle(OS_memory_provider, SHM_ptr,
+                                                 4096 /* TODO: change API */);
```
ldorau, line 169: @vinser52 @bratpiorka I think the API should be changed here in this way:
https://github.com/ldorau/unified-memory-framework/commit/de9a1ab4e24ef958643f798ed65703869517c2e6

examples/basic/ipc_shared_memory.sh
```diff
@@ -0,0 +1,36 @@
...
+# Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+
+#!/bin/bash
+
+# port should be a number from the range <1024, 65535>
+PORT=$(( 1024 + ( $$ % ( 65535 - 1024 ))))
+
+PTRACE_SCOPE_FILE="/proc/sys/kernel/yama/ptrace_scope"
```
bratpiorka, line 13: please add a comment why we need this

examples/basic/ipc_shared_memory_client.c
```diff
@@ -0,0 +1,220 @@
+/*
```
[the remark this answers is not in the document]
ldorau: Done

examples/basic/ipc_shared_memory.sh
ldorau: Done

ldorau:
> Also, @ldorau please extend existing `bind` tests to test MAP_SHARED as well. I'm not sure whether hwloc_set_area_membind will work with MAP_SHARED. According to the documentation of mbind: `The specified policy will be ignored for any MAP_SHARED mappings in the specified memory range. Rather the pages will be allocated according to the memory policy of the thread that caused the page to be allocated.`

@igchor Only the `UMF_NUMA_MODE_DEFAULT` NUMA mode is supported for now in case of the `UMF_MEM_MAP_SHARED` memory flag. The support for other NUMA modes will be added later.

Added the code to `os_initialize()`:
```c
    if (in_params->flag == UMF_MEM_MAP_SHARED && in_params->numa_mode != UMF_NUMA_MODE_DEFAULT) {
        LOG_ERR("Unsupported NUMA mode for the UMF_MEM_MAP_SHARED memory flag "
                "(only the UMF_NUMA_MODE_DEFAULT is supported for now)");
    }

```

examples/CMakeLists.txt
```diff
@@ -108,3 +108,36 @@ else()
             "UMF_BUILD_LEVEL_ZERO_PROVIDER, UMF_BUILD_LIBUMF_POOL_DISJOINT and "
             "UMF_ENABLE_POOL_TRACKING to be turned ON - skipping")
 endif()
+
+if(LINUX)
+    set(BASE_NAME ipc_shm_osmemprov)
```
bratpiorka, line 113: could we name it ipc_os_prov?

ldorau: Maybe `ipc_shm_os_prov` ?

bratpiorka: IPC always uses shared memory, so I think it is redundant

src/provider/provider_os_memory.c
```diff
@@ -24,8 +25,15 @@
 
 #define NODESET_STR_BUF_LEN 1024
 
+// maximum size of file used for memory mapping
+#define MAX_SIZE_FD (1ULL << 44)
```
bratpiorka, line 29: please explain why 44

src/provider/provider_os_memory.c
```diff
@@ -24,8 +25,15 @@
...
+
 typedef struct os_memory_provider_t {
     unsigned protection; // combination of OS-specific protection flags
+    unsigned flag;       // memory mapping flag
```
bratpiorka, line 33: flag -> visibility?

src/provider/provider_os_memory_linux.c
```diff
@@ -32,6 +36,74 @@ umf_result_t os_translate_mem_protection_one_flag(unsigned in_protection,
...
+        return 0;
+    }
+
+    int fd = -1;
+
+#ifdef __NR_memfd_secret
+    // SYS_memfd_secret is supported since Linux 5.14
+    fd = syscall(SYS_memfd_secret, 0);
+    if (fd == -1) {
+        LOG_PDEBUG("memfd_secret()");
```
bratpiorka, line 71: this is an error? please change the log message

src/provider/provider_os_memory_linux.c
```diff
@@ -32,6 +36,74 @@ umf_result_t os_translate_mem_protection_one_flag(unsigned in_protection,
...
+        LOG_PDEBUG("memfd_create()");
+    }
+    if (fd > 0) {
+        LOG_DEBUG("file descriptor created using memfd_create()");
+    }
+#endif /* __NR_memfd_create */
+
+    return fd;
+
+#endif /* __APPLE__ */
```
bratpiorka, line 94: this is "else" so /* !__APPLE__ */

src/provider/provider_os_memory_linux.c
```diff
@@ -32,6 +36,74 @@ umf_result_t os_translate_mem_protection_one_flag(unsigned in_protection,
...
+    // SYS_memfd_create is supported since Linux 3.17, glibc 2.27
+    fd = syscall(SYS_memfd_create, "fd_name", 0);
+    if (fd == -1) {
+        LOG_PDEBUG("memfd_create()");
+    }
+    if (fd > 0) {
+        LOG_DEBUG("file descriptor created using memfd_create()");
+    }
+#endif /* __NR_memfd_create */
+
```
bratpiorka, line 91: could you add an fallback to error if neither SYS_memfd_create nor SYS_memfd_secret is supported?

examples/basic/ipc_shm_osmemprov_consumer.c
```diff
@@ -0,0 +1,206 @@
...
+    if (umf_result != UMF_RESULT_SUCCESS) {
+        fprintf(stderr,
+                "[consumer] ERROR: creating OS memory provider failed\n");
+        return -1;
+    }
+
+    // get the size of the IPC handle
+    size_t IPC_handle_size;
+    umf_result =
+        umfMemoryProviderGetIPCHandleSize(OS_memory_provider, &IPC_handle_size);
```
bratpiorka, line 62: I think it would be better to do a test based on these files rather than an example. It makes no sense to call Memory Provider IPC API from the user code

bratpiorka (changes requested):

examples/basic/ipc_shm_osmemprov_consumer.c
ldorau: Done

src/provider/provider_os_memory.c
```diff
@@ -598,6 +676,110 @@ static umf_result_t os_allocation_merge(void *provider, void *lowPtr,
...
+                                      size_t size, void *providerIpcData) {
+    if (provider == NULL || providerIpcData == NULL) {
+        return UMF_RESULT_ERROR_INVALID_ARGUMENT;
+    }
+
+    os_memory_provider_t *os_provider = (os_memory_provider_t *)provider;
+
+    uintptr_t rkey;
+    void *rvalue;
+    int ret = critnib_find(os_provider->ptr_off, (uintptr_t)ptr, FIND_LE, &rkey,
```
vinser52, line 705: Following our offline discussion, we need to investigate why we can get a `ptr` that does not exactly match a key from the `ptr_off` map. Keys stored in the `ptr_off` map are the addresses pointed to at the beginning of the coarse grain allocations (returned by `os_alloc`).
My guess is that caused by split/merge functionality. If so, it is incorrect fix, and we should maintain the state fo this `ptr_off` map in split/merge functions.

include/umf/providers/provider_os_memory.h
```diff
@@ -29,6 +29,14 @@ typedef enum umf_mem_protection_flags_t {
     /// @endcond
 } umf_mem_protection_flags_t;
 
+/// @brief Memory flags
+typedef enum umf_memory_flags_t {
```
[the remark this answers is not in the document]
ldorau: Done

include/umf/providers/provider_os_memory.h
```diff
@@ -61,6 +69,8 @@ typedef enum umf_numa_mode_t {
 typedef struct umf_os_memory_provider_params_t {
     /// Combination of 'umf_mem_protection_flags_t' flags
     unsigned protection;
+    /// 'umf_memory_flags_t' flag
+    unsigned flag;
```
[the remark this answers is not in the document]
ldorau: Done

src/provider/provider_os_memory.c
ldorau: Removed. Code changed.

src/provider/provider_os_memory.c
ldorau: Done

src/provider/provider_os_memory_linux.c
ldorau: Done, but we do not use this convention almost anywhere.

src/provider/provider_os_memory.c
```diff
@@ -191,6 +198,33 @@ static umf_result_t translate_params(umf_os_memory_provider_params_t *in_params,
...
+    result =
+        os_translate_memory_flag(in_params->visibility, &provider->visibility);
+    if (result != UMF_RESULT_SUCCESS) {
+        LOG_ERR("incorrect memory flag: %u", in_params->visibility);
+        return result;
+    }
+
+    provider->fd = os_create_anonymous_fd(provider->visibility);
+    if (provider->fd == -1) {
+        LOG_PERR(
```
bratpiorka, line 210: why PERR?

src/provider/provider_os_memory.c
```diff
@@ -191,6 +198,33 @@ static umf_result_t translate_params(umf_os_memory_provider_params_t *in_params,
         return result;
     }
 
+    result =
+        os_translate_memory_flag(in_params->visibility, &provider->visibility);
```
bratpiorka, line 202: please change name to "os_translate_mem_visibility_flag" as this is more consistent with other functions

src/provider/provider_os_memory.c
```diff
@@ -191,6 +198,33 @@ static umf_result_t translate_params(umf_os_memory_provider_params_t *in_params,
...
+    if (result != UMF_RESULT_SUCCESS) {
+        LOG_ERR("incorrect memory flag: %u", in_params->visibility);
```
bratpiorka, line 204: "... memory visibility flag ..."

src/provider/provider_os_memory.c
```diff
@@ -191,6 +198,33 @@ static umf_result_t translate_params(umf_os_memory_provider_params_t *in_params,
...
+        LOG_PERR(
+            "creating an anonymous file descriptor for memory mapping failed");
+        return UMF_RESULT_ERROR_UNKNOWN;
+    }
+
+    provider->size_fd = 0; // will be increased during each allocation
+    provider->max_size_fd = get_max_file_size();
+
+    errno = 0;
+    if (provider->fd > 0 &&
```
bratpiorka, line 219: what if provider->fd == 0? you check only -1.
Also, could you move calling os_set_file_size outside if condition and use additional var to hold the ret value? This way it would be easier to debug and analyse the code if needed.

bratpiorka (changes requested):

src/provider/provider_os_memory.c
ldorau: Because `os_create_anonymous_fd` sets `errno`

ldorau: Replaced with `LOG_ERR()`

src/provider/provider_os_memory.c
ldorau: Done

src/provider/provider_os_memory.c
ldorau: Done

src/provider/provider_os_memory.c
ldorau: 1) It is OK. The following code should be executed only if `provider->fd > 0`
2) Done

src/provider/provider_os_memory.c
```diff
@@ -26,6 +27,12 @@
 
 typedef struct os_memory_provider_t {
     unsigned protection; // combination of OS-specific protection flags
+    unsigned visibility; // memory visibility mode
+    int fd;              // file descriptor for memory mapping
+    size_t size_fd;      // size of file used for memory mapping
+    size_t max_size_fd;  // maximum size of file used for memory mapping
+    // a critnib map storing (ptr, fd_offset + 1) pairs (+ 1 to be able to store fd_offset == 0)
```
lukaszstolarczuk, line 34: you could write why do we need to keep these pairs

src/provider/provider_os_memory_linux.c
```diff
@@ -6,13 +6,26 @@
...
+#include <sys/syscall.h>
 #include <unistd.h>
 
-#include "provider_os_memory_internal.h"
 #include <umf/providers/provider_os_memory.h>
 
+#include "provider_os_memory_internal.h"
+#include "utils_log.h"
+
+// maximum value of the off_t type
```
lukaszstolarczuk, line 21: you could write a comment why this arbitrary `-999999`

src/provider/provider_os_memory.c
```diff
@@ -459,6 +529,18 @@ static umf_result_t os_alloc(void *provider, size_t size, size_t alignment,
         }
     }
 
+    if (os_provider->fd > 0) {
+        // store (fd_offset + 1) to be able to store fd_offset == 0
+        ret =
+            critnib_insert(os_provider->fd_offset_map, (uintptr_t)addr,
+                           (void *)(uintptr_t)(fd_offset + 1), 0 /* update */);
+        if (ret) {
+            LOG_ERR("os_alloc(): inserting a value to the IPC cache failed "
```
lukaszstolarczuk, line 538: soo.. this critnib is an IPC cache? perhaps the name of the variable is wrong...? or you could comment it somehow

src/provider/provider_os_memory.c
```diff
@@ -579,21 +667,157 @@ static const char *os_get_name(void *provider) {
...
+
+typedef struct os_ipc_data_t {
+    int pid;
+    int fd;
+    size_t fd_offset;
+    size_t size;
+} os_ipc_data_t;
+
+static umf_result_t os_get_ipc_handle_size(void *provider, size_t *size) {
+    if (provider == NULL || size == NULL) {
```
lukaszstolarczuk, line 726: do we care if `provider == NULL`?

src/provider/provider_os_memory.c
```diff
@@ -579,21 +667,157 @@ static const char *os_get_name(void *provider) {
...
+        return UMF_RESULT_ERROR_INVALID_ARGUMENT;
+    }
+
+    *size = sizeof(os_ipc_data_t);
+    return UMF_RESULT_SUCCESS;
+}
+
+static umf_result_t os_get_ipc_handle(void *provider, const void *ptr,
+                                      size_t size, void *providerIpcData) {
+    if (provider == NULL || providerIpcData == NULL) {
```
lukaszstolarczuk, line 736: check `ptr` as well?

src/provider/provider_os_memory_linux.c
```diff
@@ -72,3 +173,39 @@ int os_purge(void *addr, size_t length, int advice) {
...
+
+    *fd_out = fd_dup;
+
+    return UMF_RESULT_SUCCESS;
+#else
+    (void)pid;                             // unused
+    (void)fd_in;                           // unused
+    (void)fd_out;                          // unused
+    return UMF_RESULT_ERROR_NOT_SUPPORTED; // unsupported
+                                           // TODO: find another way
```
lukaszstolarczuk, line 209: this TODO is a remark for what exactly? 😉 perhaps move this before `if defined...` ?

test/CMakeLists.txt
```diff
@@ -240,3 +240,36 @@ endif()
...
+
+    file(COPY ${CMAKE_CURRENT_SOURCE_DIR}/${BASE_NAME}.sh
+         DESTINATION ${CMAKE_CURRENT_BINARY_DIR})
+
+    add_test(
+        NAME ${TEST_NAME}
+        COMMAND ${BASE_NAME}.sh
+        WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
+
+    set_tests_properties(${TEST_NAME} PROPERTIES LABELS "example")
```
lukaszstolarczuk, line 270: not sure if this should have `example` label - this may be misleading somehow - it's still a test, not an example file

src/provider/provider_os_memory_linux.c
ldorau: Done

src/provider/provider_os_memory.c
ldorau: Done

src/provider/provider_os_memory.c
ldorau: No, fixed.

src/provider/provider_os_memory.c
ldorau: Replaced with "file descriptor offset map"

src/provider/provider_os_memory_linux.c
ldorau: Fixed

test/CMakeLists.txt
lukaszstolarczuk: not done?

src/provider/provider_os_memory.c
lukaszstolarczuk: thx!

test/CMakeLists.txt
```diff
@@ -240,3 +240,36 @@ endif()
...
+else()
+    message(
+        STATUS "IPC shared memory example is supported on Linux only - skipping"
```
lukaszstolarczuk, line 273: so this msg should also be updated, I believe

ldorau: Done

test/CMakeLists.txt
ldorau: Done now :-)

src/provider/provider_os_memory.c
ldorau: Done

src/provider/provider_os_memory.c
ldorau: @bratpiorka Done

src/provider/provider_os_memory.c
ldorau: @bratpiorka Done

bratpiorka (approved):

lukaszstolarczuk (commented):
:+1:

src/provider/provider_os_memory.c
```diff
@@ -583,21 +672,159 @@ static const char *os_get_name(void *provider) {
...
+    if (provider == NULL || providerIpcData == NULL || ptr == NULL) {
+        return UMF_RESULT_ERROR_INVALID_ARGUMENT;
+    }
+
+    os_memory_provider_t *os_provider = (os_memory_provider_t *)provider;
+    os_ipc_data_t *os_ipc_data = (os_ipc_data_t *)providerIpcData;
+
+    int fd;
+    umf_result_t umf_result =
+        os_duplicate_fd(os_ipc_data->pid, os_ipc_data->fd, &fd);
```
vinser52, line 795: Where `fd` is closed?

ldorau: Right :-) Fixed.

ldorau: Done

ldorau:
@igchor please re-review

ldorau:
@vinser52 please re-review

test/ipc_os_prov_consumer.c
```diff
@@ -0,0 +1,250 @@
...
+- Consumer closes the IPC handle received from the producer
+- Producer receives the response from the consumer: "This is the consumer. I just wrote a new number directly into your shared memory!"
+- Producer verifies the consumer wrote the correct value (the old one / 2) to the producer's shared memory: 70361291106696
+- Producer puts the IPC handle
+- Consumer shuts down
+- Producer shuts down
+*/
+
+int main(int argc, char *argv[]) {
+    struct sockaddr_in consumer_addr;
```
igchor, line 52: nit: could you put the code responsible for sending/receving the data over the sockets into separate functions? it's a bit hard to review it as is.

include/umf/providers/provider_os_memory.h
```diff
@@ -61,6 +67,8 @@ typedef enum umf_numa_mode_t {
 typedef struct umf_os_memory_provider_params_t {
     /// Combination of 'umf_mem_protection_flags_t' flags
     unsigned protection;
+    /// 'umf_memory_visibility_t' visibility mode
+    unsigned visibility;
```
igchor, line 71: why unsigned and not umf_memory_visibility_t?

src/provider/provider_os_memory.c
```diff
@@ -583,21 +672,162 @@ static const char *os_get_name(void *provider) {
 
 static umf_result_t os_allocation_split(void *provider, void *ptr,
```
igchor, line 673: I think this function is supposed to be thread-safe. Currently, we take a lock in the tracking provider and this function would be called under that lock so everything is fine. We might want to add some kind of assert here to make sure that tracking is enabled though.

vinser52: > I think this function is supposed to be thread-safe. Currently, we take a lock in the tracking provider and this function would be called under that lock so everything is fine.

If this function is supposed to be thread-safe it does not matter if it is called under the lock. Or I did not get your point. 

> We might want to add some kind of assert here to make sure that tracking is enabled though.

I think split/merge functionality is independent of memory tracking capabilities, right?

Anyway, I think in general @igchor raised a good point, we should specify the requirements for memory provider interface implementations. 
In this particular case, do we assume that split/merge can be called concurrently for the same memory region? I think no but maybe I miss something. If split/merge is called concurrently for different memory regions I think it is fine.

I would say the following:

- The implementation of these functions should be thread-safe.
- It should not be called concurrently with the same pointer.

igchor: I think we do allow calling split/merge on the same memory region concurrently right now - that's why we have a mutex in memory tracker.

src/provider/provider_os_memory.c
```diff
@@ -26,6 +27,13 @@
 
 typedef struct os_memory_provider_t {
     unsigned protection; // combination of OS-specific protection flags
+    unsigned visibility; // memory visibility mode
+    int fd;              // file descriptor for memory mapping
+    size_t size_fd;      // size of file used for memory mapping
+    size_t max_size_fd;  // maximum size of file used for memory mapping
+    // A critnib map storing (ptr, fd_offset + 1) pairs (+ 1 to be able to store fd_offset == 0).
```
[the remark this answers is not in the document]
ldorau: `critnib_get()` returns value or NULL, so a value cannot equal 0

src/provider/provider_os_memory.c
```diff
@@ -218,6 +254,14 @@ static umf_result_t os_initialize(void *params, void **provider) {
     umf_os_memory_provider_params_t *in_params =
         (umf_os_memory_provider_params_t *)params;
 
+    if (in_params->visibility == UMF_MEM_MAP_SHARED &&
+        in_params->numa_mode != UMF_NUMA_MODE_DEFAULT) {
+        LOG_ERR("Unsupported NUMA mode for the UMF_MEM_MAP_SHARED memory "
```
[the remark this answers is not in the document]
ldorau: Yes, later in a separate PR. This PR adds support only for `UMF_NUMA_MODE_DEFAULT`.

src/provider/provider_os_memory.c
```diff
@@ -583,21 +672,162 @@ static const char *os_get_name(void *provider) {
...
     (void)totalSize;
-    (void)firstSize;
-    // nop
+
+    os_memory_provider_t *os_provider = (os_memory_provider_t *)provider;
+    if (os_provider->fd <= 0) {
+        return UMF_RESULT_SUCCESS;
+    }
+
+    void *value = critnib_get(os_provider->fd_offset_map, (uintptr_t)ptr);
```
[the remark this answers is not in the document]
ldorau: Done: https://github.com/oneapi-src/unified-memory-framework/issues/463

src/provider/provider_os_memory.c
```diff
@@ -583,21 +672,162 @@ static const char *os_get_name(void *provider) {
...
+    if (value == NULL) {
+        LOG_ERR("os_allocation_split(): getting a value from the file "
```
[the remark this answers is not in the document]
ldorau: Done.

.github/workflows/pr_push.yml
```diff
@@ -84,7 +85,7 @@ jobs:
 
     - name: Run examples
       working-directory: ${{github.workspace}}/build
-      run: ctest --output-on-failure --test-dir examples -C Release
+      run: ctest -V --test-dir examples -C Release
```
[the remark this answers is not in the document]
ldorau: `-V` gives us ability to verify logs if the two processes communicate correctly

include/umf/providers/provider_os_memory.h
ldorau: Done

src/provider/provider_os_memory.c
ldorau: An assert added.
Done.

src/provider/provider_os_memory_linux.c
```diff
@@ -32,6 +43,88 @@ umf_result_t os_translate_mem_protection_one_flag(unsigned in_protection,
...
+    return 0;                     // ignored on MacOSX
+#else                             /* !__APPLE__ */
+    // fd is created only for MAP_SHARED
+    if (translated_memory_flag != MAP_SHARED) {
+        return 0;
+    }
+
+    int fd = -1;
+
+#ifdef __NR_memfd_secret
```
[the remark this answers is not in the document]
ldorau: Added the `UMF_MEM_FD_FUNC` variable.
Done.

src/provider/provider_os_memory.c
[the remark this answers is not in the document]
vinser52: Ah, got it.

src/provider/provider_os_memory.c
vinser52: > An assert added.

But what if the tracker is disabled? Do we want to support split/merge in that case?

> I think we do allow calling split/merge on the same memory region concurrently right now

Do we have such scenarios? For me, it is hard to imagine. Maybe it is better to specify that "It should **not** be called concurrently with the same pointer."

vinser52 (changes requested):

test/ipc_os_prov_consumer.c
ldorau: Done

src/provider/provider_os_memory.c
ldorau: Yeah, we all have to agree upon one version first. I removed this assert and left a comment only.

.github/workflows/pr_push.yml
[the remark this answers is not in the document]
igchor: Shouldn't we rather have some log matching to verify that the test is working correctly? It would be better to just see pass/fail in the output.

src/provider/provider_os_memory.c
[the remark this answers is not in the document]
ldorau: OK. Done

.github/workflows/pr_push.yml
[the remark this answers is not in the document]
ldorau: The return code is enough. Removed this change.
Done.

src/provider/provider_os_memory.c
ldorau: Done

src/provider/provider_os_memory.c
```diff
@@ -347,8 +408,20 @@ static int os_mmap_aligned(void *hint_addr, size_t length, size_t alignment,
         extended_length += alignment;
     }
 
-    void *ptr = os_mmap(hint_addr, extended_length, prot);
+    size_t fd_offset = 0;
+
+    if (fd > 0) {
+        fd_offset = *fd_size;
+        *fd_size += extended_length;
```
vinser52, line 415: We are updating the output parameter `fd_size`, but `if` block on the next line might return an error. Should we check the error condition and update the output arguments only if the function successes?

Another question, how practical is to reach the `max_fd_size` limit? I understand that it is `SIZE_MAX` now, but what if workflow causes continuos malloc/free from memory provider?

igchor (approved):

vinser52 (approved):

src/provider/provider_os_memory.c
ldorau: `max_fd_size = max(off_t) = 9223372036854775807 = 2^63-1 = 8388608 TB - 1 = 8192 PB - 1` is not to be reached practically, I think ...

Updating `fd_size` fixed.
Done.

ldorau: @vinser52 Done

bratpiorka: We were discussing this with @pbalcer and it's exactly what was done in PMDK and nobody ever complained about it. I know it's an anecdotal argument :), but in reality such a scenario where SIZE_MAX is not enough is *very* rare.

pbalcer: Like ldorau says, it's not possible for this variable to overflow in practice.

vinser52: yeah, I have the same feeling, just asked to confirm. The scenario I had in mind was a sequence of malloc/free when total memory consumption might not increase, but on each malloc, OS memory provider does not reuse the memory region that was freed before. The `fd_size` always monotonically increases.

PR #432, opened 2024-04-11
Add requirement for HWLOC version 

kswiecicki:
### Checklist

- [ ] All tests pass locally


cmake/FindLIBHWLOC.cmake
```diff
@@ -14,8 +14,25 @@ find_file(LIBHWLOC_HEADER NAMES hwloc.h)
...
+        message(
+            WARNING
+                "LIBHWLOC_INCLUDE_DIR found, but header with version info is missing"
+        )
+    endif()
+endif()
+
 if(WINDOWS)
-    find_file(LIBHWLOC_DLL NAMES "bin/hwloc-15.dll" "bin/libhwloc-15.dll")
+    find_file(LIBHWLOC_DLL NAMES "bin/hwloc-*.dll" "bin/libhwloc-*.dll")
```
lukaszstolarczuk, line 35: it looks like the dll file was not found...? perhaps the `*` is not working here as we expected it...?

cmake/FindLIBHWLOC.cmake
```diff
@@ -25,9 +42,21 @@ if(LIBHWLOC_LIBRARY)
...
     message(STATUS "    LIBHWLOC_INCLUDE_DIRS = ${LIBHWLOC_INCLUDE_DIRS}")
     message(STATUS "    LIBHWLOC_LIBRARY_DIRS = ${LIBHWLOC_LIBRARY_DIRS}")
+    message(STATUS "    LIBHWLOC_VERSION = ${LIBHWLOC_VERSION}")
     if(WINDOWS)
         message(STATUS "    LIBHWLOC_DLL_DIRS = ${LIBHWLOC_DLL_DIRS}")
     endif()
+
+    if(LIBHWLOC_FIND_VERSION)
+        if(NOT LIBHWLOC_VERSION)
+            message(WARNING "Failed to retrieve libhwloc version")
```
lukaszstolarczuk, line 52: it looks like we got a lot of warnings in Windows jobs

kswiecicki: Should be resolved now.

cmake/FindLIBHWLOC.cmake
kswiecicki: Yeah, it seems that `find_file` doesn't accept regex nor wildcards.

lukaszstolarczuk (commented):
something's wrong with QEMU now - perhaps the version number is handled differently in hwloc-2.9.0 (which is installed in there)...?

On the other hand, if the issue is not easily solvable, I guess you can use the new install_hwloc.sh and it should work.

kswiecicki:
> something's wrong with QEMU now - perhaps the version number is handled differently in hwloc-2.9.0 (which is installed in there)...?
> 
> On the other hand, if the issue is not easily solvable, I guess you can use the new install_hwloc.sh and it should work.

Done. I've set the hwloc on QEMU to be installed with install_hwloc.sh.

bratpiorka (approved):

PR #442, opened 2024-04-17
Add size param to the close_ipc_handle API

vinser52:
### Description
Add size param to the close_ipc_handle API

It is needed by https://github.com/oneapi-src/unified-memory-framework/pull/429

Ref: #429 

### Checklist

- [x] All tests pass locally

### DIFF
--- a/include/umf/memory_provider.h
+++ b/include/umf/memory_provider.h
@@ -188,12 +188,13 @@ umfMemoryProviderOpenIPCHandle(umf_memory_provider_handle_t hProvider,
 /// @brief Close an IPC memory handle.
 /// @param hProvider [in] handle to the memory provider.
 /// @param ptr [in] pointer returned by umfMemoryProviderOpenIPCHandle function.
+/// @param size [in] size of the memory address range.
 /// @return UMF_RESULT_SUCCESS on success or appropriate error code on failure.
 ///         UMF_RESULT_ERROR_INVALID_ARGUMENT if invalid \p hProvider or \p ptr are passed.
 ///         UMF_RESULT_ERROR_NOT_SUPPORTED if IPC functionality is not supported by this provider.
 umf_result_t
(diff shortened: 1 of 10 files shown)


include/umf/memory_provider.h
```diff
@@ -188,12 +188,13 @@ umfMemoryProviderOpenIPCHandle(umf_memory_provider_handle_t hProvider,
 /// @brief Close an IPC memory handle.
 /// @param hProvider [in] handle to the memory provider.
 /// @param ptr [in] pointer returned by umfMemoryProviderOpenIPCHandle function.
+/// @param size [in] size of the memory.
```
ldorau, line 191: Maybe "`size of the memory address range`" like above in `umfMemoryProviderGetIPCHandle`?

include/umf/memory_provider_ops.h
```diff
@@ -125,10 +125,11 @@ typedef struct umf_memory_provider_ipc_ops_t {
     /// @brief Closes an IPC memory handle.
     /// @param provider pointer to the memory provider.
     /// @param ptr pointer to the memory retrieved with open_ipc_handle function.
+    /// @param size size of the memory address range.
```
ldorau, line 128: "`size size`" -> `size`

ldorau (changes requested):
only 2 minor issues

include/umf/memory_provider.h
vinser52: Done

include/umf/memory_provider_ops.h
vinser52: but the first `size` is an argument name and the second on is part of the comment. 

We have the same comments for other methods with `size` parameter.

ldorau: ah :-) OK, I see now

ldorau (approved):
LGTM

lukaszstolarczuk (commented):
👍

Issue #440, opened 2024-04-15
Add -Wsign-compare and -Wconversion flags 


lplewa:
On windows some of those problems are detected - what causes fully working on Linux PR to fail on Windows, what is annoying. 
Unfortunately, those Flags are more restrictive then on windows (by adding those flags we have to fix ~40 warnings in the code)

Question: Do we want to include those compiler flags?

bratpiorka:
I am for adding as many checks as possible :) at least if this costs us little effort. Fixing 40 easy warnings should be straightforward.

PR #445, opened 2024-04-17
add -Wsign-compare and -Wconversion

lplewa:
This is mostly needed to detect similar issues that windows compiler detect by default.

fixes: oneapi-src/unified-memory-framework#440

### Checklist

- [ ] All tests pass locally


src/provider/provider_os_memory.c
```diff
@@ -412,7 +413,7 @@ static umf_result_t os_alloc(void *provider, size_t size, size_t alignment,
         return UMF_RESULT_ERROR_INVALID_ARGUMENT;
     }
 
-    int protection = os_provider->protection;
+    int protection = (int)os_provider->protection;
```
PatKamin, line 416: Perhaps the `os_provider->protection` should be declared an int? If that's what mmap expects, this would be reasonable.

PR #451, opened 2024-04-23
Add IPC shared memory example using the ipc.h API

ldorau:
### Description

Add IPC shared memory example using the ipc.h API.

### Checklist
- [x] All tests pass locally

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -38,3 +38,36 @@ and build this example Level Zero development package should be installed.
 * Level Zero headers and libraries
 * compatible GPU with installed driver
 * set UMF_BUILD_GPU_EXAMPLES, UMF_BUILD_LIBUMF_POOL_DISJOINT, UMF_BUILD_LEVEL_ZERO_PROVIDER and UMF_ENABLE_POOL_TRACKING CMake configuration flags to ON
+
+## IPC example with shared memory
+This example also demonstrates how to use UMF IPC API. The example creates two
+processes: a producer and a consumer that communicate in the following way
+(the initial value N in the shared memory is quasi-random):
+- Consumer starts
+- Consumer creates a socket
+- Consumer listens for incoming connections
+- Producer starts
+- Producer's shared memory contains a number: N
+- Producer gets the IPC handle
+- Producer creates a socket
+- Producer connects to the consumer
+- Consumer connects at IP 127.0.0.1 and a port to the producer
+- Producer sends the size of the IPC handle to the consumer
+- Consumer receives the size of the IPC handle
+- Consumer sends the received size of the IPC handle as a confirmation back to the producer
+- Producer receives the confirmation from the consumer and verifies if it is correct
+- Producer sends the IPC handle to the consumer
+- Consumer receives the IPC handle from the producer
+- Consumer opens the IPC handle received from the producer
+- Consumer reads the number from the producer's shared memory: N
+- Consumer writes a new number directly to the producer's shared memory: N/2
+- Consumer sends a response message to the producer
+- Producer receives the response message from the consumer: "This is the consumer. I just wrote a new number directly into your shared memory!"
+- Producer verifies if the consumer wrote the correct value (the old one / 2) to the producer's shared memory: N/2
+- Consumer closes the IPC handle received from the producer
+- Producer puts the IPC handle
+- Consumer shuts down
+- Producer shuts down
+
+### Requirements
+* set UMF_BUILD_LIBUMF_POOL_SCALABLE CMake configuration flag to ON
--- a/examples/basic/ipc_shm_ipcapi.sh
+++ b/examples/basic/ipc_shm_ipcapi.sh
@@ -1,0 +1,48 @@
+#
+# Copyright (C) 2024 Intel Corporation
+#
+# Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+#
+
+#!/bin/bash
+
+# port should be a number from the range <1024, 65535>
+PORT=$(( 1024 + ( $$ % ( 65535 - 1024 ))))
+
+# The ipc_shm_ipcapi example requires using pidfd_getfd(2)
+# to obtain a duplicate of another process's file descriptor.
+# Permission to duplicate another process's file descriptor
+# is governed by a ptrace access mode PTRACE_MODE_ATTACH_REALCREDS check (see ptrace(2))
+# that can be changed using the /proc/sys/kernel/yama/ptrace_scope interface.
+PTRACE_SCOPE_FILE="/proc/sys/kernel/yama/ptrace_scope"
+VAL=0
+if [ -f $PTRACE_SCOPE_FILE ]; then
+	PTRACE_SCOPE_VAL=$(cat $PTRACE_SCOPE_FILE)
+	if [ $PTRACE_SCOPE_VAL -ne $VAL ]; then
+		# check if sudo requires password
+		if ! timeout --kill-after=5s 3s sudo date; then
+			echo "ERROR: sudo requires password - cannot set ptrace_scope to 0"
+			exit 1
+		fi
+		echo "Setting ptrace_scope to 0 (classic ptrace permissions) ..."
+		echo "$ sudo bash -c \"echo $VAL > $PTRACE_SCOPE_FILE\""
+		sudo bash -c "echo $VAL > $PTRACE_SCOPE_FILE"
+	fi
+	PTRACE_SCOPE_VAL=$(cat $PTRACE_SCOPE_FILE)
+	if [ $PTRACE_SCOPE_VAL -ne $VAL ]; then
+		echo "SKIP: setting ptrace_scope to 0 (classic ptrace permissions) FAILED - skipping the test"
+		exit 0
+	fi
+fi
+
+UMF_LOG_VAL="level:debug;flush:debug;output:stderr;pid:yes"
+
+echo "Starting ipc_shm_ipcapi CONSUMER on port $PORT ..."
+UMF_LOG=$UMF_LOG_VAL ./umf_example_ipc_shm_ipcapi_consumer $PORT &
+
+echo "Waiting 1 sec ..."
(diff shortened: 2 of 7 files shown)


examples/basic/ipc_shm_ipcapi_producer.c
```diff
@@ -0,0 +1,214 @@
...
+#include <umf/ipc.h>
+#include <umf/memory_pool.h>
+#include <umf/pools/pool_jemalloc.h>
+#include <umf/providers/provider_os_memory.h>
+
+#define INET_ADDR "127.0.0.1"
+#define MSG_SIZE 1024
+
+int main(int argc, char *argv[]) {
+    struct sockaddr_in consumer_addr;
```
vinser52, line 24: Sockets is a platform-specific way perhaps for simplification purposes we can use just files to communicate between processes. The producer creates an IPC handle and writes it to the file and the consumer process reads it from the file. Such example will be cross-platform.

examples/basic/ipc_shm_ipcapi.sh
```diff
@@ -0,0 +1,41 @@
...
+	fi
+	PTRACE_SCOPE_VAL=$(cat $PTRACE_SCOPE_FILE)
+	if [ $PTRACE_SCOPE_VAL -ne $VAL ]; then
+		echo "SKIP: setting ptrace_scope to 0 (classic ptrace permissions) FAILED - skipping the test"
+		exit 0
+	fi
+fi
+
+echo "Starting ipc_shm_ipcapi CONSUMER on port $PORT ..."
+./umf_example_ipc_shm_ipcapi_consumer $PORT &
```
vinser52, line 35: Instead of using a script to run the example can one process (e.g. consumer) just start the producer process?

ldorau: Such processes are not independent - the child process inherits some stuff from the parent process (AFAIR: environment, open file descriptors, ...). More over we have to set `ptrace_scope` to 0 - it is also done in this script.

examples/basic/ipc_shm_ipcapi_producer.c
ldorau: I am afraid it would not be simpler. `recv()` waits for a message. We would have to implement a mechanism of waiting for a complete message being written into the file (checking if the message was written to the file and if it is complete).

bratpiorka: I think that since we already have this example working we could merge this version. As for now, we don't have plans to use IPC on Windows. OS provider IPC is also implemented on Linux only, on Windows it would return the NOT_SUPPORTED error for now.

examples/basic/ipc_shm_ipcapi.sh
bratpiorka: I think that this way of running processes is closer to the requirements for UMF IPC functionality

examples/README.md
vinser52, line 1: We should also update online documentation. If we plan to do it in a separate PR lets create an issue for that task.

examples/README.md
```diff
@@ -38,3 +38,32 @@ and build this example Level Zero development package should be installed.
...
+- Consumer starts
+- Consumer creates a socket
+- Consumer listens for incoming connections
+- Producer starts
+- Producer's shared memory contains a number: 140722582213392
+- Producer gets the IPC handle
+- Producer creates a socket
+- Producer connects to the consumer
+- Consumer connects at IP 127.0.0.1 and port 47770 to the producer
+- Producer sents the IPC handle to the consumer (24 bytes)
```
vinser52, line 55: `sents` -> `sends`

vinser52, line 55: `(24 bytes)` is implementation details that might be changed. Let's omit such details.

examples/README.md
```diff
@@ -38,3 +38,32 @@ and build this example Level Zero development package should be installed.
...
+- Consumer receives the IPC handle from the producer (24 bytes)
+- Consumer opens the IPC handle received from the producer
+- Consumer reads the number from the producer's shared memory: 140722582213392
+- Consumer writes a new number directly to the producer's shared memory: 70361291106696
+- Consumer sents a response message to the producer
```
vinser52, line 60: `sents` -> `sends`

examples/basic/ipc_shm_ipcapi_consumer.c
```diff
@@ -0,0 +1,229 @@
...
+        goto err_destroy_OS_memory_provider;
+    }
+
+    producer_socket = consumer_connect(port);
+    if (producer_socket < 0) {
+        goto err_destroy_scalable_pool;
+    }
+
+    // receive the IPC handle from the producer's
+    ssize_t len = recv(producer_socket, recv_buffer, RECV_BUFF_SIZE, 0);
```
vinser52, line 132: what if the size of the IPC handle is greater than `RECV_BUFF_SIZE`?

I think we need to send the size of the IPC handle first, and then receive the exact amount of bytes corresponding to IPC handle. It will make our example clearer and better demonstrate how to transfer IPC handle between processes.

examples/basic/ipc_shm_ipcapi_consumer.c
```diff
@@ -0,0 +1,229 @@
...
+
+    fprintf(
+        stderr,
+        "[consumer] Received the IPC handle from the producer (%zi bytes)\n",
+        len);
+
+    void *SHM_ptr;
+    umf_result = umfOpenIPCHandle(scalable_pool, IPC_handle, &SHM_ptr);
+    if (umf_result == UMF_RESULT_ERROR_NOT_SUPPORTED) {
+        fprintf(stderr,
```
vinser52, line 148: why do we need special processing for the `UMF_RESULT_ERROR_NOT_SUPPORTED` ? 
Furthermore, this block of code is not reachable, because if `umfOpenIPCHandle` is not supported then `umfGetIPCHandle` is not supported as well by the same memory provider on the producer side (the producer is not able to send the IPC handle in that case).

examples/basic/ipc_shm_ipcapi_producer.c
```diff
@@ -0,0 +1,240 @@
...
+                                         &OS_memory_provider);
+    if (umf_result != UMF_RESULT_SUCCESS) {
+        fprintf(stderr,
+                "[producer] ERROR: creating OS memory provider failed\n");
+        return -1;
+    }
+
+    size_t page_size;
+    umf_result =
+        umfMemoryProviderGetMinPageSize(OS_memory_provider, NULL, &page_size);
```
vinser52, line 89: Why do we need to use `page_size`? I think the allocation size does not matter for the IPC example.

examples/basic/ipc_shm_ipcapi_producer.c
```diff
@@ -0,0 +1,240 @@
...
+    umf_result = umfPoolCreate(umfScalablePoolOps(), OS_memory_provider, NULL,
+                               0, &scalable_pool);
+    if (umf_result != UMF_RESULT_SUCCESS) {
+        fprintf(stderr,
+                "[producer] ERROR: creating the UMF jemalloc pool failed\n");
+        goto err_destroy_OS_memory_provider;
+    }
+
+    // Make 3 allocations of size: 1 page, 2 pages and 3 pages
+    void *ptr1, *ptr2, *IPC_shared_memory;
```
vinser52, line 106: for what reason we need `ptr1` and `ptr2`? It just blurs the attention.

vinser52 (changes requested):

examples/basic/ipc_shm_ipcapi_consumer.c
ldorau: We need this to correctly handle the `UMF_RESULT_ERROR_NOT_SUPPORTED` error, because `umfOpenIPCHandle()` can return `UMF_RESULT_ERROR_NOT_SUPPORTED` when `! defined(__NR_pidfd_open) || ! defined(__NR_pidfd_getfd)`.
It is reachable on Ubuntu 20.04 where the `! defined(__NR_pidfd_open) || ! defined(__NR_pidfd_getfd)` condition is true:
https://github.com/ldorau/unified-memory-framework/actions/runs/9002739451/job/24731880245

vinser52: Ah, I see.

examples/basic/ipc_shm_ipcapi_producer.c
vinser52: Now I understand the reason. The `pool_proxy` implementation does not automatically round up the size to be multiple of the page size. But as I described in #475, I think the `pool_proxy` should be extended to support arbitrary allocation sizes.

examples/README.md
ldorau: Done

ldorau: Done

examples/README.md
ldorau: Done

examples/basic/ipc_shm_ipcapi_consumer.c
ldorau: Done

examples/basic/ipc_shm_ipcapi_producer.c
ldorau: Removed

examples/basic/ipc_shm_ipcapi_producer.c
ldorau: It was a test for correctness of `fd_offset` - it is done in the IPC test.
Removed.

examples/basic/ipc_shm_ipcapi_producer.c
ldorau: A size can be any now, because we use `umfPoolCalloc(scalable_pool)` to allocate this memory.

examples/README.md
ldorau: Done: https://github.com/oneapi-src/unified-memory-framework/issues/477

vinser52 (approved):

bratpiorka (approved):

bratpiorka:
@vinser52 please close all conversations if you think the changes are ok :)

vinser52:
> @vinser52 please close all conversations if you think the changes are ok :)

Done

Issue #461, opened 2024-04-26
Fix all valgrind-memcheck errors in the tests


lukaszstolarczuk:
Fix all multi-numa tests run under memcheck. They are currently filtered-out.

Ref. #459

PR #464, opened 2024-04-30
Fix valgrind multi numa tests

lukaszstolarczuk:
Clean up multi numa tests up a little, move "free" function up in the code, where possible, and use `EXPECT_` rather than `ASSERT_` to allow proper freeing.

While these tests seem clean from any valgrind issues now, they still are incompatibile with valgrind, hence I've left the gtest filters (but I've set the more specific ones).

Closes: #461


szadam (approved):

bratpiorka:
The changes look good but after merging I think we shouldn't close #461 yet

bratpiorka (approved):

PR #468, opened 2024-05-06
Enable IPC tests for Level Zero provider

vinser52:
### Description
Enable IPC tests for the Level Zero provider.
The main challenge was to unify work with host and device memory.

### Checklist

- [x] All tests pass locally

### DIFF
--- a/test/ipcFixtures.hpp
+++ b/test/ipcFixtures.hpp
@@ -19,10 +19,39 @@
 #include <numeric>
 #include <tuple>
 
-using testParams = std::tuple<umf_memory_pool_ops_t *, void *,
-                              umf_memory_provider_ops_t *, void *>;
+class MemoryAccessor {
+  public:
+    virtual void fill(void *ptr, size_t size, const void *pattern,
+                      size_t pattern_size) = 0;
+    virtual void copy(void *dst_ptr, void *src_ptr, size_t size) = 0;
+};
 
-struct umfIpcTest : umf_test::test, ::testing::WithParamInterface<testParams> {
+class HostMemoryAccessor : public MemoryAccessor {
+  public:
+    void fill(void *ptr, size_t size, const void *pattern,
+              size_t pattern_size) override {
+        assert(ptr != nullptr);
+        assert(pattern != nullptr);
+        assert(pattern_size > 0);
+        while (size) {
+            size_t copy_size = std::min(size, pattern_size);
+            std::memcpy(ptr, pattern, copy_size);
+            ptr = static_cast<char *>(ptr) + copy_size;
+            size -= copy_size;
+        }
+    }
+
+    void copy(void *dst_ptr, void *src_ptr, size_t size) override {
+        std::memcpy(dst_ptr, src_ptr, size);
+    }
+};
+
+using ipcTestParams =
+    std::tuple<umf_memory_pool_ops_t *, void *, umf_memory_provider_ops_t *,
+               void *, MemoryAccessor *>;
+
+struct umfIpcTest : umf_test::test,
+                    ::testing::WithParamInterface<ipcTestParams> {
     umfIpcTest() : pool(nullptr, nullptr) {}
     void SetUp() override {
         test::SetUp();
@@ -36,7 +65,7 @@ struct umfIpcTest : umf_test::test, ::testing::WithParamInterface<testParams> {
         //       from memoryPool.hpp
         umf_memory_provider_handle_t hProvider;
         umf_memory_pool_handle_t hPool;
-        auto [pool_ops, pool_params, provider_ops, provider_params] =
+        auto [pool_ops, pool_params, provider_ops, provider_params, accessor] =
             this->GetParam();
 
         auto ret =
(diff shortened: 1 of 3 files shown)


test/providers/provider_level_zero.cpp
```diff
@@ -326,21 +326,66 @@ struct umfLevelZeroProviderTest : umf_test::test {
...
+    void SetUp() override {
+        test::SetUp();
+
+        auto [l0_params, accessor] = this->GetParam();
+        params = l0_params;
+        hDevice = (ze_device_handle_t)params.level_zero_device_handle;
+        hContext = (ze_context_handle_t)params.level_zero_context_handle;
+
+        if (nullptr == hDevice) {
+            GTEST_SKIP() << "Test skipped, no GPU devices found";
```
lukaszstolarczuk, line 350: I think we may fail the test actually - we have a specific CMake flag to enable these tests (`UMF_BUILD_GPU_TESTS`, OFF by default). I believe we prefer not to have a "silent skips" (gtest doesn't say a word about skips).

test/providers/provider_level_zero.cpp
```diff
@@ -58,178 +61,175 @@ struct libze_ops {
...
+        }
     }
+};
 
-    void DestroyLevelZeroOps() {
-        // Just close the handle here
-        dlclose(zeDlHandle);
-    }
+std::unique_ptr<void, DlHandleCloser> zeDlHandle = nullptr;
+//void *zeDlHandle = nullptr;
```
lukaszstolarczuk, line 74: I guess you can remove this line

test/providers/provider_level_zero.cpp
```diff
@@ -5,9 +5,12 @@
 #include <dlfcn.h>
 #include <level_zero/ze_api.h>
 
+#include "ipcFixtures.hpp"
 #include "pool.hpp"
 #include "umf/providers/provider_level_zero.h"
 
+#include <mutex>
```
lukaszstolarczuk, line 12: is it needed...?

test/providers/provider_level_zero.cpp
```diff
@@ -326,21 +326,66 @@ struct umfLevelZeroProviderTest : umf_test::test {
...
+            zeMemoryTypeExpected = ZE_MEMORY_TYPE_DEVICE;
+            break;
+        case UMF_MEMORY_TYPE_SHARED:
+            zeMemoryTypeExpected = ZE_MEMORY_TYPE_SHARED;
+            break;
+        case UMF_MEMORY_TYPE_HOST:
+            zeMemoryTypeExpected = ZE_MEMORY_TYPE_HOST;
+            break;
+        default:
+            GTEST_SKIP() << "Test skipped, unsupported memory type";
```
lukaszstolarczuk, line 364: fail here as well...?

kswiecicki (approved):

test/providers/provider_level_zero.cpp
vinser52: The `InitLevelZero` function uses `std::call_once` which requires this header.

test/providers/provider_level_zero.cpp
vinser52: Done

test/providers/provider_level_zero.cpp
vinser52: Done

test/providers/provider_level_zero.cpp
vinser52: Done. I have removed `default` case and added `ASSERT_NE(zeMemoryTypeExpected, ZE_MEMORY_TYPE_UNKNOWN);` after `switch` block.

lukaszstolarczuk (approved):
thx!

bratpiorka (changes requested):

test/ipcFixtures.hpp
```diff
@@ -19,10 +19,36 @@
...
+
+class HostMemoryAccessor : public MemoryAccessor {
+  public:
+    void fill(void *ptr, size_t size, const void *pattern,
+              size_t pattern_size) override {
+        while (size) {
+            size_t copy_size = std::min(size, pattern_size);
+            std::memcpy(ptr, pattern, copy_size);
+            ptr = static_cast<char *>(ptr) + copy_size;
+            size -= copy_size;
```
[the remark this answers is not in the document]
vinser52: Strictly speaking, it should be `size >= copy_size`. And on L34 we choose minimum between `size` and `pattern_size`, so `copy_size` cannot be greater than `size`.

test/ipcFixtures.hpp
```diff
@@ -19,10 +19,36 @@
...
+    virtual void fill(void *ptr, size_t size, const void *pattern,
+                      size_t pattern_size) = 0;
+    virtual void copy(void *dst_ptr, void *src_ptr, size_t size) = 0;
+};
+
+class HostMemoryAccessor : public MemoryAccessor {
+  public:
+    void fill(void *ptr, size_t size, const void *pattern,
+              size_t pattern_size) override {
+        while (size) {
```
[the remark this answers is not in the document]
vinser52: Done.
Added `assert`

PR #471, opened 2024-05-07
Implement umfPoolCreateFromMemspace() with proxy pool

ldorau:
### Description

Implement `umfPoolCreateFromMemspace()` with proxy pool.

### Checklist
- [x] All tests pass locally


igchor:
Why proxy and not scalable pool? (or even disjoint pool if we scalable is not available)

ldorau:
> Why proxy and not scalable pool? (or even disjoint pool if we scalable is not available)

It was just the first step. I will convert it to the scalable pool with a fallback to the disjoint pool.

ldorau:
> Why proxy and not scalable pool? (or even disjoint pool if we scalable is not available)

@igchor It would require linking `libumf` with `libscalable_pool`.

See: https://github.com/oneapi-src/unified-memory-framework/actions/runs/8998833427/job/24719757805?pr=471
```
/usr/bin/ld: ../lib/libumf.a(memory_target_numa.c.o): in function `numa_pool_create_from_memspace':
memory_target_numa.c:(.text+0x1c5): undefined reference to `umfScalablePoolOps'
```

vinser52:
> > Why proxy and not scalable pool? (or even disjoint pool if we scalable is not available)
> 
> @igchor It would require linking `libumf` with `libscalable_pool`.
> 
> See: https://github.com/oneapi-src/unified-memory-framework/actions/runs/8998833427/job/24719757805?pr=471
> 
> ```
> /usr/bin/ld: ../lib/libumf.a(memory_target_numa.c.o): in function `numa_pool_create_from_memspace':
> memory_target_numa.c:(.text+0x1c5): undefined reference to `umfScalablePoolOps'
> ```

But we agreed in the past that we do not need separate library for scalable pool (the same like with L0 provider). We do not need additional proxy between libumf and libtbbmalloc - libumf can directly try dlopen(libtbbmalloc).

Regarding proxy pool, I think it is useless by default to create a proxy pool. We need some real heap manager by default.

PR #479, opened 2024-05-10
Remove assert from tracking provider

igchor:
This assert would be triggered every time a memory is leaked by the application. Currently this happens in certain SYCL tests.
Ref: https://github.com/intel/llvm/actions/runs/9024914892/job/24800360426?pr=13343

### Checklist

- [ ] All tests pass locally

### DIFF
--- a/src/provider/provider_tracking.c
+++ b/src/provider/provider_tracking.c
@@ -405,7 +405,6 @@ static void check_if_tracker_is_empty(umf_memory_tracker_handle_t hTracker,
                         "left)",
                         n_items);
             }
-            assert(n_items == 0);
         }
     }
 }


bratpiorka:
Igor, please rebase. I think you need https://github.com/oneapi-src/unified-memory-framework/pull/473

vinser52:
But it is triggered only in debug build.
Does llvm uses debug version of UMF?

igchor:
> But it is triggered only in debug build. Does llvm uses debug version of UMF?

I believe the build type is inherited from llvm.

vinser52:
> > But it is triggered only in debug build. Does llvm uses debug version of UMF?
> 
> I believe the build type is inherited from llvm.

Is it because the UMF sources today are just fetched by UR and UR is built as a part of compiler build?

Anyway I am not against removing this assert

igchor:
> > > But it is triggered only in debug build. Does llvm uses debug version of UMF?
> > 
> > 
> > I believe the build type is inherited from llvm.
> 
> Is it because the UMF sources today are just fetched by UR and UR is built as a part of compiler build?
> 
> Anyway I am not against removing this assert

Yes, that's correct.

igchor:
> Igor, please rebase. I think you need #473

I am rebased: see my branch https://github.com/igchor/unified-memory-framework/tree/remove_assert

igchor:
@bratpiorka restarting the CI solved the issue

src/provider/provider_tracking.c
```diff
@@ -405,7 +405,6 @@ static void check_if_tracker_is_empty(umf_memory_tracker_handle_t hTracker,
                         "left)",
                         n_items);
             }
-            assert(n_items == 0);
```
bratpiorka, line 408: what do you think about logging a message here?

igchor: We already do that (see just above this assert).

bratpiorka (approved):

ldorau (approved):

Issue #481, opened 2024-05-10
Disjoint pool causes memory leak calling umfMemoryProviderFree() with size equal 0


ldorau:
Disjoint pool causes memory leak calling `umfMemoryProviderFree()` with `size` equal 0:

https://github.com/oneapi-src/unified-memory-framework/blob/6fc3e56b4870184dec70613359c4db2ef7f1180a/src/pool/pool_disjoint.cpp#L407
```c
static void memoryProviderFree(umf_memory_provider_handle_t hProvider,
                               void *ptr) {
    auto ret = umfMemoryProviderFree(hProvider, ptr, 0);
    if (ret != UMF_RESULT_SUCCESS) {
        throw MemoryProviderError{ret};
    }
}
```

what can cause an error (which is ignored now).

## Environment Information

- UMF version (hash commit or a tag): 6fc3e56b4870184dec70613359c4db2ef7f1180a

ldorau:
@igchor @vinser52 @bratpiorka

vinser52:
More details just FYI.
Initially disjoint pool was designed for L0 provider.  L0 provider does not need the size when memory is freed (the size parameter is ignored).
The similar story is with the OS provider implementation on Windows - the size parameter is ignored when memory is freed.

igchor:
When we are deallocating Slabs then it's easy, we can just pass the Slab size here: https://github.com/oneapi-src/unified-memory-framework/blob/6fc3e56b4870184dec70613359c4db2ef7f1180a/src/pool/pool_disjoint.cpp#L441

But for allocations that bypass the pooling we do not know the size. Perhaps we could leverage memory tracking there?

vinser52:
> But for allocations that bypass the pooling we do not know the size. Perhaps we could leverage memory tracking there?

Exactly, the plan is to use tracking capabilities. @ldorau already prepared such a fix for the `pool_proxy`: Red #487.

ldorau:
@vinser52 @igchor The fix is ready: #490 - please review :-)

PR #482, opened 2024-05-10
Do not ignore error in os_free() when size == 0

ldorau:
### Description

Do not ignore error in `os_free()` when size == 0.

Ref: #481

### Checklist
- [x] All tests pass locally


benchmark/ubench.c
```diff
@@ -196,6 +196,7 @@ UBENCH_EX(simple, proxy_pool_with_os_memory_provider) {
     }
 
     umf_memory_pool_handle_t proxy_pool;
+    // the proxy pool requires the tracking provider to work correctly
```
vinser52, line 199: The tracking provider is always required to work correctly :)
Would you like to say that tracking capabilities should be ON?

My general thoughts: the amount of features that depend on tracking capabilities is increasing. We should consider making the tracking provider always ON (remove the option that allows to disable it) or we need to check it at build/run time and return appropriate error. I vote for the first option (always unconditional ON).

benchmark/ubench.c
```diff
@@ -238,8 +239,10 @@ UBENCH_EX(simple, disjoint_pool_with_os_memory_provider) {
     disjoint_memory_pool_params.MinBucketSize = DISJOINT_POOL_MIN_BUCKET_SIZE;
 
     umf_memory_pool_handle_t disjoint_pool;
-    umf_result = umfPoolCreate(umfDisjointPoolOps(), os_memory_provider,
-                               &disjoint_memory_pool_params, 0, &disjoint_pool);
+    // disable the tracking provider to make it work faster (UMF_POOL_CREATE_FLAG_DISABLE_TRACKING)
```
vinser52, line 242: Are we interested in performance without a tracking provider? But in real workloads tracking capabilities will be ON?

src/provider/provider_tracking.c
```diff
@@ -343,6 +348,15 @@ static umf_result_t trackingFree(void *hProvider, void *ptr, size_t size) {
         umf_ba_global_free(value);
     }
 
+    // umfMemoryProviderFree() should not be called with size == 0,
```
vinser52, line 351: But according to this PR  `umfMemoryProviderFree()` for tracking provider can be called with `size == 0`.

Maybe a particular pool implementation should explicitly get the size from the tracker and call `umfMemoryProviderFree()` with the right size?

vinser52 (changes requested):

benchmark/ubench.c
ldorau: I agree. Done: https://github.com/oneapi-src/unified-memory-framework/pull/485

benchmark/ubench.c
ldorau: Right. Removed.
Done.

src/provider/provider_tracking.c
ldorau: Right. New fix: https://github.com/oneapi-src/unified-memory-framework/pull/487

ldorau:
Replaced with #487

Issue #488, opened 2024-05-10
umf-ipc_os_prov silenty change security permissions


lplewa:
User by running ```ctest``` command in terminal where sudo works without password (if you recently used sudo, next prompts do not ask for password) this test permanently changes security setting for this machine (to make this worse we even never change it back). 

This is a security issue:
- We should never change any settings on the platform, without user consent.
- We should never use escalate permissions in tests, by exploiting sudo. 

How this should be solved:
if platform is miss-configured this test should be SKIP in ctest. 
In CI we should configure platforms to enable this feature. 
It would be good if there is extra option in CI to fail on skip - so we do not have silent skips, when we are expecting tests to run.

lplewa:
@wlemkows

PR #495, opened 2024-05-14
[Tests] skip test with improper ptrace value

lukaszstolarczuk:
Add skipping of IPC test, which require a specific value of ptrace (set in the system). Enable it in the CI in all workflows.

Fixes: #488

### DIFF
--- a/.github/workflows/sanitizers.yml
+++ b/.github/workflows/sanitizers.yml
@@ -52,6 +52,7 @@ jobs:
         -DUSE_UBSAN=${{matrix.sanitizers.ubsan}}
         -DUSE_TSAN=${{matrix.sanitizers.tsan}}
         -DUMF_BUILD_EXAMPLES=ON
+        -DUMF_TESTS_FAIL_ON_SKIP=ON
 
     - name: Build UMF
       run: cmake --build ${{env.BUILD_DIR}} -j $(nproc)
@@ -96,6 +97,7 @@ jobs:
         -DUSE_UBSAN=${{matrix.sanitizers.ubsan}}
         -DUSE_TSAN=${{matrix.sanitizers.tsan}}
         -DUMF_BUILD_EXAMPLES=ON
+        -DUMF_TESTS_FAIL_ON_SKIP=ON
 
     - name: Build UMF
       run: cmake --build ${{env.BUILD_DIR}} -j $(nproc)
@@ -157,6 +159,7 @@ jobs:
         -DUSE_ASAN=${{matrix.sanitizers.asan}}
         -DUMF_BUILD_EXAMPLES=ON
         -DUMF_BUILD_LEVEL_ZERO_PROVIDER=OFF
+        -DUMF_TESTS_FAIL_ON_SKIP=ON
 
     - name: Build UMF
       run: cmake --build ${{env.BUILD_DIR}} --config Debug -j $Env:NUMBER_OF_PROCESSORS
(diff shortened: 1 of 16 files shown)


bratpiorka (approved):

.github/workflows/sanitizers.yml
```diff
@@ -34,6 +34,9 @@ jobs:
         apt-get update
         apt-get install -y cmake libnuma-dev libjemalloc-dev libtbb-dev libhwloc-dev sudo
 
+    - name: Set ptrace value for IPC test
```
ldorau, line 37: The same should be done also in line no. 83 below.

ldorau (changes requested):

test/ipc_os_prov.sh
```diff
@@ -20,19 +20,8 @@ VAL=0
 if [ -f $PTRACE_SCOPE_FILE ]; then
 	PTRACE_SCOPE_VAL=$(cat $PTRACE_SCOPE_FILE)
 	if [ $PTRACE_SCOPE_VAL -ne $VAL ]; then
```
ldorau, line 22: I think this test should NOT skip silently in CI, but it should always error out in CI:

```sh
	if [ $PTRACE_SCOPE_VAL -ne $VAL ]; then
		if [ "$CI" == "true" ]; then
			echo "ERROR: ptrace_scope is not set to 0 (classic ptrace permissions)"
			exit 1 # error out in case of CI
		else
			echo "SKIP: ptrace_scope is not set to 0 (classic ptrace permissions) - skipping the test"
			exit 125 # skip code defined in CMakeLists.txt
		fi
	fi
```

ldorau (changes requested):

test/ipc_os_prov.sh
lukaszstolarczuk: Done.

CMakeLists.txt
```diff
@@ -44,6 +44,8 @@ option(UMF_DEVELOPER_MODE "Enable developer checks, treats warnings as errors"
 option(UMF_FORMAT_CODE_STYLE
        "Add clang, cmake, and black -format-check and -format-apply targets"
        OFF)
+# Only part of skips are treated as fail, now. TODO: extend to all tests
```
ldorau, line 47: `Only a part of skips is treated as fail now.` (AFAIK ;-))

CMakeLists.txt
```diff
@@ -59,6 +61,16 @@ set(UMF_PROXY_LIB_BASED_ON_POOL
 set_property(CACHE UMF_PROXY_LIB_BASED_ON_POOL
              PROPERTY STRINGS ${KNOWN_PROXY_LIB_POOLS})
 
+if(UMF_BUILD_TESTS
+   AND DEFINED ENV{CI}
```
ldorau, line 65: We can verify if `"$CI" == "true"` exactly.

ldorau (approved):

CMakeLists.txt
lukaszstolarczuk: tbh, I couldn't make it work - I'm not sure how CMake is treating env vars...

anyway, this shouldn't be a big problem since you'd have to have exactly `CI` variable defined in your OS, for this to fail, and even so, it's just a matter of adding extra param to the CMake command.

lplewa (approved):

CMakeLists.txt
ldorau: It is OK.

Issue #514, opened 2024-05-22
Introduce init/teardown functions instead of relying on library ctor/dtor


igchor:
## Rationale

UMF creates and destroys global state on linux using `__attribute__((constructor))` and  `__attribute__((destructor))` respectively. This is problematic for libraries that do cleanup in it's own library destructor. For example, llvm, might in some cases, try to free memory to UMF in it's destructor after UMF has already been destroyed. 

Because llvm uses dlopen to (indirectly) load UMF, setting constructor/destructor priority is not enough to guarantee proper ordering.

## Description

UMF should expose explicit init/teardownAPI, similarly to UR.

The UMF API should implement ref counting, similarly, to: https://oneapi-src.github.io/unified-runtime/core/api.html?highlight=teardown#uradapterrelease as there might be multiple users (adapters) initializing and deinitializing UMF - the resources should only be released once the last user is gone.

## API Changes

umf_result_t umfInit(/* TBD */);
umf_result_t umfTeardown(void);

## Implementation details

Move all global state initialization and destruction to the new functions. For proxy_library we can still have library ctor and dtor that would just call init/teardown function. Since those function would implement refcounting it should be safe to use proxy_lib with applications that use UMF directly.

ldorau:
Implementation is ready: #530 - @igchor please review.

bratpiorka:
@igchor in #530 Łukasz added required functions, but they work slightly differently from urLoaderInit/Teardown - in UMF user must call these functions only when UMF is loaded using dlopen(). In the scenario where UMF is linked dynamically in CMake they are not needed.

pbalcer:
> Łukasz added required functions, but they work slightly differently from urLoaderInit/Teardown - in UMF user must call these functions only when UMF is loaded using dlopen(). In the scenario where UMF is linked dynamically in CMake they are not needed.

They might be needed though, in UR. Where UMF itself is *not* dlopen'ed, but UR is. Anyway, I agree the library should work without these calls, but calling them, no matter how umf is linked, should not break things and cause the state destruction happen in the umfTeardown.

PR #522, opened 2024-05-27
Do not run umfPoolDestroy() nor umfMemoryProviderDestroy() if umf has already been destroyed

ldorau:
### Description

Add `umf_ba_is_destroyed()` and `umf_is_destroyed()`.

Do not run `umfPoolDestroy()` nor `umfMemoryProviderDestroy()` if umf has already been destroyed.

### Checklist
- [x] All tests pass locally

### DIFF
--- a/src/libumf_linux.c
+++ b/src/libumf_linux.c
@@ -33,6 +33,8 @@ void __attribute__((destructor)) umfDestroy(void) {
     umfDestroyTopology();
 }
 
+int umf_is_destroyed(void) { return (TRACKER == NULL); }
+
 void libumfInit(void) {
     // do nothing, additional initialization not needed
 }
--- a/src/base_alloc/base_alloc_global.c
+++ b/src/base_alloc/base_alloc_global.c
@@ -47,6 +47,8 @@ void umf_ba_destroy_global(void) {
     }
 }
 
+int umf_ba_is_destroyed(void) { return (BASE_ALLOC.ac[0] == NULL); }
+
 static void umf_ba_create_global(void) {
     for (int i = 0; i < NUM_ALLOCATION_CLASSES; i++) {
         // allocation classes need to be powers of 2
--- a/src/base_alloc/base_alloc_global.h
+++ b/src/base_alloc/base_alloc_global.h
@@ -19,6 +19,7 @@ void umf_ba_global_free(void *ptr);
 void umf_ba_destroy_global(void);
 size_t umf_ba_global_malloc_usable_size(void *ptr);
 void *umf_ba_global_aligned_alloc(size_t size, size_t alignment);
+int umf_ba_is_destroyed(void);
 
 #ifdef __cplusplus
 }
--- a/src/libumf.h
+++ b/src/libumf.h
@@ -17,6 +17,8 @@ extern "C" {
 // initializes runtime state needed by the library (needed mostly for static libaries on windows)
 void libumfInit(void);
 
+int umf_is_destroyed(void);
+
 #ifdef __cplusplus
 }
 #endif
--- a/src/libumf_windows.c
+++ b/src/libumf_windows.c
@@ -23,8 +23,11 @@ static void umfCreate(void) {
 
 static void umfDestroy(void) {
     umfMemoryTrackerDestroy(TRACKER);
+    TRACKER = NULL;
     umf_ba_destroy_global();
 }
+
+int umf_is_destroyed(void) { return (TRACKER == NULL); }
 
 #if defined(UMF_SHARED_LIBRARY)
 BOOL APIENTRY DllMain(HINSTANCE hinstDLL, DWORD fdwReason, LPVOID lpvReserved) {
--- a/src/memory_pool.c
+++ b/src/memory_pool.c
@@ -71,6 +71,10 @@ err_provider_create:
 }
 
 void umfPoolDestroy(umf_memory_pool_handle_t hPool) {
+    if (umf_is_destroyed() || umf_ba_is_destroyed()) {
+        return;
+    }
+
     hPool->ops.finalize(hPool->pool_priv);
     if (hPool->flags & UMF_POOL_CREATE_FLAG_OWN_PROVIDER) {
         // Destroy associated memory provider.
--- a/src/memory_provider.c
+++ b/src/memory_provider.c
@@ -195,6 +195,10 @@ umf_result_t umfMemoryProviderCreate(const umf_memory_provider_ops_t *ops,
 }
 
 void umfMemoryProviderDestroy(umf_memory_provider_handle_t hProvider) {
+    if (umf_is_destroyed() || umf_ba_is_destroyed()) {
+        return;
+    }
+
     hProvider->ops.finalize(hProvider->provider_priv);
     umf_ba_global_free(hProvider);
 }


src/libumf_linux.c
```diff
@@ -33,6 +33,8 @@ void __attribute__((destructor)) umfDestroy(void) {
     umfDestroyTopology();
 }
 
+int umf_is_destroyed(void) { return (TRACKER == NULL); }
```
vinser52, line 36: I did not get the idea. If UMF is destroyed we cannot call any function from the library.

vinser52 (changes requested):

src/libumf_linux.c
ldorau: It happens that OpenMP destroys UMF before `umfPoolDestroy()` and `umfMemoryProviderDestroy()` were called.

vinser52: My question is the following: if UMF is destroyed, how adding `umf_is_destroyed` can help?

What "destroy" means in the OpenMP scenario? UMF library is unloaded? If so, no UMF functions can be called.

ldorau: "Destroy" in the OpenMP scenario means that the UMF's destructor has been called, but we do not if it was called or not and we want to call `umfPoolDestroy()` and `umfMemoryProviderDestroy()` in the destructor of OpenMP's allocator.

ldorau: Adding this patch helps - it prevents from a "use-after-free" segmentation fault.

ldorau:
> @ldorau @vinser52 would the approach proposed in #514 solve the issue ? This aims to solve a similar problem for llvm, but more generally.

No, because the UMF's constructor and destructor are called only once in this case before a segmentation fault:

1) UMF's constructor is called (I call dlopen(libumf.so))
2) UMF's destructor is called incorrectly by OpenMP on its own (I do not want that)
3) I call umfPoolDestroy() to free memory <<<--- it cause a "use-after-free" segmentation fault
4) UMF's destructor would be called correctly here

src/libumf_linux.c
ldorau: > Ok, but why OpenMP allocator calls `umfPoolDestroy` after UMF's destructor is called. 

1) UMF's constructor is called (I call dlopen(libumf.so))
2) UMF's destructor is called incorrectly by OpenMP on its own (I do not know why and I do not want that)
3) I call umfPoolDestroy() to free memory <<<--- it cause a "use-after-free" segmentation fault
4) UMF's destructor (dlclose()) would be called correctly here

> Does it happen because UMF is used via `dlopen` by OpenMP 

Yes.

> and OpenMP calls `dlclose` in its destructor?

Probably, I am almost sure that's why.

ldorau:
So it is just a WA for an issue in OpenMP. I suppose it is jus a bug.

igchor:
> > @ldorau @vinser52 would the approach proposed in #514 solve the issue ? This aims to solve a similar problem for llvm, but more generally.
> 
> No, because the UMF's constructor and destructor are called only once in this case before a segmentation fault:
> 
> 1. UMF's constructor is called (I call dlopen(libumf.so))
> 2. UMF's destructor is called incorrectly by OpenMP on its own (I do not want that)
> 3. I call umfPoolDestroy() to free memory <<<--- it cause a "use-after-free" segmentation fault
> 4. UMF's destructor would be called correctly here

If the issue is because OpenMP calls dloclose in it's destructor as you wrote in a previous comment then the approach in #514 would actually help - that's the same problem as in llvm. My understanding is that when dlclose is called on UMF in other library destructor, the destructor of UMF can actually be called BEFORE that other one.

With the init/teardown API OpenMP would just call teardown once all other resources from UMF (pools, providers, ...) are freed and there would be no use-after-free.

ldorau:
> > > @ldorau @vinser52 would the approach proposed in #514 solve the issue ? This aims to solve a similar problem for llvm, but more generally.
> > 
> > No, because the UMF's constructor and destructor are called only once in this case before a segmentation fault:
> > 
> > 1. UMF's constructor is called (I call dlopen(libumf.so))
> > 2. UMF's destructor is called incorrectly by OpenMP on its own (I do not want that)
> > 3. I call umfPoolDestroy() to free memory <<<--- it cause a "use-after-free" segmentation fault
> > 4. UMF's destructor would be called correctly here
> 
> If the issue is because OpenMP calls dloclose in it's destructor as you wrote in a previous comment then the approach in #514 would actually help - that's the same problem as in llvm. My understanding is that when dlclose is called on UMF in other library destructor, the destructor of UMF can actually be called BEFORE that other one.
> 
> With the init/teardown API OpenMP would just call teardown once all other resources from UMF (pools, providers, ...) are freed and there would be no use-after-free.

Yes, this would help indeed.

vinser52:
> > > > @ldorau @vinser52 would the approach proposed in #514 solve the issue ? This aims to solve a similar problem for llvm, but more generally.
> > > 
> > > 
> > > No, because the UMF's constructor and destructor are called only once in this case before a segmentation fault:
> > > 
> > > 1. UMF's constructor is called (I call dlopen(libumf.so))
> > > 2. UMF's destructor is called incorrectly by OpenMP on its own (I do not want that)
> > > 3. I call umfPoolDestroy() to free memory <<<--- it cause a "use-after-free" segmentation fault
> > > 4. UMF's destructor would be called correctly here
> > 
> > 
> > If the issue is because OpenMP calls dloclose in it's destructor as you wrote in a previous comment then the approach in #514 would actually help - that's the same problem as in llvm. My understanding is that when dlclose is called on UMF in other library destructor, the destructor of UMF can actually be called BEFORE that other one.
> > With the init/teardown API OpenMP would just call teardown once all other resources from UMF (pools, providers, ...) are freed and there would be no use-after-free.
> 
> Yes, this would help indeed.

Guys, I still do not quite understand how the proposed changes might fix the issue. If `dlclose` is called by OpenMP so why OpenMP calls UMF API after that? In that case, it should be fixed on the OpenMP side. If UMF is unloaded by the loader and OpenMP does not know about that, then I do not understand how the proposed functions might help. If UMF is unloaded already calling UMF function might cause crash because memory used by UMF library can be reused for something else.

bratpiorka:
> > Yes, this would help indeed.
> 
> Guys, I still do not quite understand how the proposed changes might fix the issue. If `dlclose` is called by OpenMP so why OpenMP calls UMF API after that? In that case, it should be fixed on the OpenMP side. If UMF is unloaded by the loader and OpenMP does not know about that, then I do not understand how the proposed functions might help. If UMF is unloaded already calling UMF function might cause crash because memory used by UMF library can be reused for something else.

I think if it was a bug in some important closed-source application, then we could consider such a WA, but if it comes from an Intel component that we have access to, then we should try to solve it by fixing the source of the problem first.

ldorau:
> Guys, I still do not quite understand how the proposed changes might fix the issue. If `dlclose` is called by OpenMP so why OpenMP calls UMF API after that? In that case, it should be fixed on the OpenMP side. If UMF is unloaded by the loader and OpenMP does not know about that, then I do not understand how the proposed functions might help. If UMF is unloaded already calling UMF function might cause crash because memory used by UMF library can be reused for something else.

@vinser52 @bratpiorka @igchor I have debugged it more and it turned out that only in one group ("Offload") of tests the UMF's destructor is called **by the OS program loader** before the destructor of the OpenMP library which uses UMF. 
So I think that the solution proposed by @igchor would solve this issue. I have no idea now how we could fix it in another way.

ldorau:
Replaced with https://github.com/oneapi-src/unified-memory-framework/pull/530

PR #525, opened 2024-05-28
cleanup include ze_api.h

bratpiorka:
### Description

ze_api.h should be always:
* included from the deps (build/_deps/level-zero-loader-***) for all internal code including tests
* included from the system paths in the examples

### Checklist
- [x] All tests pass locally

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -52,9 +52,8 @@ if(UMF_BUILD_GPU_EXAMPLES
         LIBS disjoint_pool ze_loader umf)
 
     target_include_directories(
-        ${EXAMPLE_NAME}
-        PRIVATE ${LEVEL_ZERO_INCLUDE_DIRS} ${UMF_CMAKE_SOURCE_DIR}/src/utils
-                ${UMF_CMAKE_SOURCE_DIR}/include)
+        ${EXAMPLE_NAME} PRIVATE ${UMF_CMAKE_SOURCE_DIR}/src/utils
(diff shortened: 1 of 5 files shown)


kswiecicki (approved):

KFilipek (changes requested):
If you plan to reorder headers, please, use Coding style guidelines. 
As I remember, external headers should be in a separate block. Currently, they are mixed.

ldorau (approved):

bratpiorka:
> If you plan to reorder headers, please, use Coding style guidelines. As I remember, external headers should be in a separate block. Currently, they are mixed. E.g. ze_api.h or mutex

We don't have strict coding style rules - we only have what's here: https://github.com/oneapi-src/unified-memory-framework/blob/main/CONTRIBUTING.md#code-style
But regarding the order - I keep the "standard" includes fist (<mutex>), then UMF interface that is supposed to be installed (<umf/*>) and local includes at the end. In "test/providers/provider_level_zero.cpp" I threat "ze_api.h" as a local header, because it is downloaded by UMF to the _deps folder and I intentionally keep it with the other local files

kswiecicki:
> If you plan to reorder headers, please, use Coding style guidelines. As I remember, external headers should be in a separate block. Currently, they are mixed. E.g. ze_api.h or mutex

From what I see, the `ze_api.h` header is treated as external in examples, and as internal in umf sources/tests. In case of examples, @bratpiorka mentioned it should come from system paths from the user.

KFilipek (approved):

PR #531, opened 2024-06-05
[CI] Bump MSVC version to more generic one

lukaszstolarczuk:
Apparently, the more generic version works properly (even though it uses precisely the same version, as before).

- [x] All tests pass locally


PatKamin (approved):

lukaszstolarczuk:
hmm.. I think it works now without this fix: https://github.com/oneapi-src/unified-memory-framework/actions/runs/9391053862

should we still merge this?

bratpiorka:
> should we still merge this?

if it make the CI more stable - yes

PR #532, opened 2024-06-05
Fix memory leak if critnib_new fails

lplewa:
translate_params() allocates bitmaps which have to freed in case of error

### DIFF
--- a/src/provider/provider_os_memory.c
+++ b/src/provider/provider_os_memory.c
@@ -331,6 +331,13 @@ static umf_result_t os_initialize(void *params, void **provider) {
         goto err_destroy_hwloc_topology;
     }
 
+    os_provider->fd_offset_map = critnib_new();
+    if (!os_provider->fd_offset_map) {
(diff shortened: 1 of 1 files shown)


bratpiorka (approved):

src/provider/provider_os_memory.c
```diff
@@ -331,6 +331,13 @@ static umf_result_t os_initialize(void *params, void **provider) {
         goto err_destroy_hwloc_topology;
     }
 
+    os_provider->fd_offset_map = critnib_new();
+    if (!os_provider->fd_offset_map) {
+        LOG_ERR("creating file descriptor offset map failed");
+        ret = UMF_RESULT_ERROR_OUT_OF_HOST_MEMORY;
+        goto err_destroy_hwloc_topology;
+    }
+
```
lukaszstolarczuk, line 340: what will happen now, the other way around, if e.g. `translate_params` fail. Shouldn't we call `critnib_delete`...?

lplewa:
it will be fixed in #538

PR #533, opened 2024-06-06
Handle default kernel behavior for CPU access to all NUMA nodes

bratpiorka:
### Description

For the default CPU memspace (MemspaceHostAll) with all NUMA nodes, we should follow the default kernel behavior, and we shouldn't call any kind of mbind() that could negatively impact the performance

### Checklist

- [x] All tests pass locally

### DIFF
--- a/src/memory_targets/memory_target_numa.c
+++ b/src/memory_targets/memory_target_numa.c
@@ -87,21 +87,33 @@ static umf_result_t numa_memory_provider_create_from_memspace(
             return UMF_RESULT_ERROR_INVALID_ARGUMENT;
         }
     } else {
-        params.numa_mode = UMF_NUMA_MODE_BIND;
+        if (memspace == umfMemspaceHostAllGet()) {
+            // For the default memspace, we use the default mode without any
+            // call to mbind
+            params.numa_mode = UMF_NUMA_MODE_DEFAULT;
+        } else {
+            params.numa_mode = UMF_NUMA_MODE_BIND;
+        }
     }
 
-    params.numa_list =
-        umf_ba_global_alloc(sizeof(*params.numa_list) * numNodesProvider);
+    if (memspace == umfMemspaceHostAllGet() && policy == NULL) {
+        // For default memspace with default policy we use all numa nodes so
+        // simply left numa list empty
+        params.numa_list_len = 0;
+        params.numa_list = NULL;
+    } else {
+        params.numa_list =
+            umf_ba_global_alloc(sizeof(*params.numa_list) * numNodesProvider);
 
-    if (!params.numa_list) {
-        return UMF_RESULT_ERROR_OUT_OF_HOST_MEMORY;
+        if (!params.numa_list) {
+            return UMF_RESULT_ERROR_OUT_OF_HOST_MEMORY;
+        }
+
+        for (size_t i = 0; i < numNodesProvider; i++) {
+            params.numa_list[i] = numaTargets[i]->physical_id;
+        }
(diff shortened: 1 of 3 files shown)


kswiecicki (approved):

src/memory_targets/memory_target_numa.c
```diff
@@ -87,22 +87,34 @@ static umf_result_t numa_memory_provider_create_from_memspace(
             return UMF_RESULT_ERROR_INVALID_ARGUMENT;
         }
     } else {
-        params.numa_mode = UMF_NUMA_MODE_BIND;
+        if (memspace == umfMemspaceHostAllGet()) {
```
lplewa, line 90: This is bad . What you are doing here is a pointer comparison.  While what you want to do is check if memspace contains all numanodes in system. For now it is very difficult to get all numanodes in system from different source then memspaceHostAllGet() - but really soon it will be possible. 

What you should have here is a helper function which checks if memspace contains all nodes, or function to compare namespaces.

src/provider/provider_os_memory.c
```diff
@@ -574,28 +574,32 @@ static umf_result_t os_alloc(void *provider, size_t size, size_t alignment,
...
-            LOG_PERR("binding memory to NUMA node failed");
-            // TODO: (errno == 0) when hwloc_set_area_membind() fails on Windows,
-            // ignore this temporarily
-            if (errno != ENOSYS &&
-                errno != 0) { // ENOSYS - Function not implemented
-                // Do not error out if memory binding is not implemented at all
-                // (like in case of WSL on Windows).
-                goto err_unmap;
+    // Bind memory to NUMA nodes if numa_policy is other than DEFAULT
+    if (os_provider->numa_policy != HWLOC_MEMBIND_DEFAULT) {
```
lplewa, line 578: Do we need this if?
Calling once mbind it should not be an issue worth extra if. 
If we need it please move it few lines up - there is a code related to mbind

lplewa (changes requested):

test/memspaces/memspace_host_all.cpp
```diff
@@ -87,6 +89,69 @@ TEST_F(memspaceHostAllProviderTest, allocFree) {
     UT_ASSERTeq(ret, UMF_RESULT_SUCCESS);
 }
 
+TEST_F(memspaceHostAllProviderTest, hostAllDefaults) {
+    // This testcase checks if the allocations made uising the provider with
```
lukaszstolarczuk, line 93: uising (misspell)

src/memory_targets/memory_target_numa.c
bratpiorka: ok yes, this is a ptr comparison, but the ‘memspace’ I'm comparing here is the handle returned from umfMemspaceHostAllGet(), and this function always returns the same handle for each call. Furthermore, comparing the contents of memspace is not the same as comparing memspace itself - and here I want to check exactly whether the user is using the default (host all) memspace

src/provider/provider_os_memory.c
bratpiorka: Yes I want to avoid calling mbind at all and this can't be done without "if"

test/memspaces/memspace_host_all.cpp
bratpiorka: done

src/provider/provider_os_memory.c
bratpiorka: done - moved

lplewa (approved):

Issue #513, opened 2024-05-22
Add option to link statically with hwloc (on windows)


igchor:
## Rationale

UMF is currently being used by UR and Intel/llvm. Those projects use the `/DEPENDENTLOADFLAG:0x800` linker flag on Windows when building .dlls, which results in the loader only looking for .dll dependencies in System32. Since we currently do not have an installer for UMF, we cannot make UR depend on umf.dll or libwloc.dll. Instead, we need to link with UMF and hwloc statically.

## Description

Add CMake option to link with hwloc statically. This may be supported only on Windows.

## API Changes

None.

## Implementation details

I'm not exactly sure how static linking works on windows, we might need to do something like this: https://stackoverflow.com/questions/2157629/linking-static-libraries-to-other-static-libraries or in addition to changes in UMF, also link to hwloc when building the application (UR adapter / llvm plugin).

## Meta

Related PRs:
- https://github.com/oneapi-src/unified-runtime/pull/1430
- https://github.com/intel/llvm/pull/13343

igchor:
Once this is implemented and a new flag for static linking is added, we will need to:
- update https://github.com/intel/llvm/pull/13343 and set this newly added flag on windows
- (optional) update https://github.com/oneapi-src/unified-runtime/pull/1430 and add a CI job for windows to test the static linking

lukaszstolarczuk:
@igchor, does https://github.com/oneapi-src/unified-memory-framework/pull/534 fix this issue?

igchor:
> @igchor, does #534 fix this issue?

Yes, the build on https://github.com/intel/llvm/pull/13343 passes with this change

KFilipek:
> > @igchor, does #534 fix this issue?
> 
> Yes, the build on [intel/llvm#13343](https://github.com/intel/llvm/pull/13343) passes with this change

PR #534 was merged so you can rebase your changes.

lukaszstolarczuk:
closing then, thx!

PR #534, opened 2024-06-06
Add static linking option

KFilipek:
### Checklist

- [x] All tests pass locally

### DIFF
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -228,6 +228,76 @@ jobs:
         --umf-version ${{env.UMF_VERSION}}
         ${{ matrix.shared_library == 'ON' && '--shared-library' || ''}}
 
+  windows-dynamic_build_hwloc:
+    name: "Windows dynamic UMF + static hwloc"
+    env:
+      BUILD_DIR : "${{github.workspace}}/build/${{matrix.build_type}}"
+    strategy:
+      matrix:
+        build_type: [Release]
+
+    runs-on: 'windows-2022'
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+
+    - name: Configure build
+      run: >
+        cmake
+        -B ${{env.BUILD_DIR}}
+        -DUMF_BUILD_SHARED_LIBRARY=ON
+        -DUMF_BUILD_EXAMPLES=OFF
+        -DUMF_FORMAT_CODE_STYLE=OFF
+        -DUMF_DEVELOPER_MODE=ON
+        -DUMF_BUILD_LIBUMF_POOL_DISJOINT=ON
+        -DUMF_BUILD_LIBUMF_POOL_JEMALLOC=OFF
+        -DUMF_BUILD_LEVEL_ZERO_PROVIDER=ON
+        -DUMF_TESTS_FAIL_ON_SKIP=ON
+        -DUMF_LINK_HWLOC_STATICALLY=ON
+
+    - name: Build UMF
+      run: cmake --build ${{env.BUILD_DIR}} --config ${{matrix.build_type}} -j $Env:NUMBER_OF_PROCESSORS
+
+    - name: Run tests
+      working-directory: ${{env.BUILD_DIR}}
+      run: ctest -C ${{matrix.build_type}} --output-on-failure --test-dir test
+
+  windows-static_build_hwloc:
+    name: "Windows static UMF + static hwloc"
+    env:
+      BUILD_DIR : "${{github.workspace}}/build/${{matrix.build_type}}"
+    strategy:
+      matrix:
+        build_type: [Release]
+
+    runs-on: 'windows-2022'
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+
+    - name: Configure build
+      run: >
+        cmake
+        -B ${{env.BUILD_DIR}}
+        -DUMF_BUILD_SHARED_LIBRARY=OFF
+        -DUMF_BUILD_EXAMPLES=OFF
+        -DUMF_FORMAT_CODE_STYLE=OFF
+        -DUMF_DEVELOPER_MODE=ON
+        -DUMF_BUILD_LIBUMF_POOL_DISJOINT=ON
+        -DUMF_BUILD_LIBUMF_POOL_JEMALLOC=OFF
+        -DUMF_BUILD_LEVEL_ZERO_PROVIDER=ON
+        -DUMF_TESTS_FAIL_ON_SKIP=ON
+        -DUMF_LINK_HWLOC_STATICALLY=ON
+
+    - name: Build UMF
+      run: cmake --build ${{env.BUILD_DIR}} --config ${{matrix.build_type}} -j $Env:NUMBER_OF_PROCESSORS
+
+    - name: Run tests
+      working-directory: ${{env.BUILD_DIR}}
+      run: ctest -C ${{matrix.build_type}} --output-on-failure --test-dir test
+
   macos-build:
     name: MacOS
     strategy:
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,6 +14,11 @@ set(UMF_CMAKE_SOURCE_DIR ${CMAKE_CURRENT_SOURCE_DIR})
 list(APPEND CMAKE_MODULE_PATH "${UMF_CMAKE_SOURCE_DIR}/cmake")
 include(${UMF_CMAKE_SOURCE_DIR}/cmake/helpers.cmake)
 
+include(CTest)
+include(CMakePackageConfigHelpers)
+include(GNUInstallDirs)
+find_package(PkgConfig)
+
 # CMAKE_PROJECT_VERSION[_MAJOR|_MINOR|_PATCH] variables are set via 'project'
 # command. They cannot contain any "pre-release" part, though. We use custom
 # "UMF_SRC_VERSION" to store more accurate (source) version - this var should be
@@ -39,6 +44,8 @@ option(UMF_BUILD_EXAMPLES "Build UMF examples" ON)
 option(UMF_BUILD_FUZZTESTS "Build UMF fuzz tests" OFF)
 option(UMF_BUILD_GPU_EXAMPLES "Build UMF GPU examples" OFF)
 option(UMF_DEVELOPER_MODE "Enable additional developer checks" OFF)
+option(UMF_LINK_HWLOC_STATICALLY
+       "Link UMF with HWLOC library statically (Windows+Release only)" OFF)
 option(UMF_FORMAT_CODE_STYLE
        "Add clang, cmake, and black -format-check and -format-apply targets"
        OFF)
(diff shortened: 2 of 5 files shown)


cmake/FindLIBHWLOC.cmake
```diff
@@ -4,6 +4,33 @@
 
 message(STATUS "Checking for module 'libhwloc' using find_library()")
 
+# When user choose to 
+if (UMF_LINK_HWLOC_STATICALLY)
+    message("Add hwloc as a external project")
+    include(ExternalProject)
+
+    set(HWLOC_URL "https://github.com/open-mpi/hwloc/archive/refs/heads/master.zip")
+    ExternalProject_Add(hwloc
```
bratpiorka, line 13: why this way instead of doing this similar to level-zero? 
we could have something like:

```
if (UMF_LINK_HWLOC_STATICALLY) 
    include(FetchContent)
    set(HWLOC_REPO "....git")
    set(HWLOC_TAG ...)

    FetchContent_Declare(..)
    ...
else()
    pkg_check_modules(LIBHWLOC hwloc>=2.3.0)
    if(NOT LIBHWLOC_FOUND)
        find_package(LIBHWLOC 2.3.0 REQUIRED hwloc)
    endif()
    # add PATH to DLL on Windows
    set(DLL_PATH_LIST
        "${DLL_PATH_LIST};PATH=path_list_append:${LIBHWLOC_LIBRARY_DIRS}/../bin")
endif()
```

and IMO this code should be added to CMakeLists.txt, not FindLIBHWLOC.cmake

CMakeLists.txt
```diff
@@ -37,6 +42,7 @@ option(UMF_BUILD_BENCHMARKS "Build UMF benchmarks" OFF)
 option(UMF_BUILD_BENCHMARKS_MT "Build UMF multithreaded benchmarks" OFF)
 option(UMF_BUILD_EXAMPLES "Build UMF examples" ON)
 option(UMF_BUILD_GPU_EXAMPLES "Build UMF GPU examples" OFF)
+option(UMF_LINK_HWLOC_STATICALLY "Link UMF with HWLOC library statically" OFF)
```
bratpiorka, line 45: please add static linking to the windows build matrix in CI workflow

CMakeLists.txt
```diff
@@ -83,6 +89,41 @@ else()
...
+    # add PATH to DLL on Windows
+    set(DLL_PATH_LIST
+        "${DLL_PATH_LIST};PATH=path_list_append:${LIBHWLOC_LIBRARY_DIRS}/../bin"
+    )
+else()
+    include(FetchContent)
+    FetchContent_Declare(
+        hwloc_targ
+        GIT_REPOSITORY "https://github.com/open-mpi/hwloc.git"
+        GIT_TAG master
```
PatKamin, line 106: It's better to stick with a specific tag than the current master branch which changes.

CMakeLists.txt
```diff
@@ -83,6 +89,41 @@ else()
...
+        FIND_PACKAGE_ARGS)
+
+    FetchContent_GetProperties(hwloc_targ)
+    if(NOT hwloc_targ_POPULATED)
+        FetchContent_MakeAvailable(hwloc_targ)
+    endif()
+
+    set(hwloc_cmake ${hwloc_targ_SOURCE_DIR}/contrib/windows-cmake)
```
PatKamin, line 114: Is this directory available both on Linux and Windows?

CMakeLists.txt
```diff
@@ -83,6 +89,41 @@ else()
...
+
+    add_subdirectory(${hwloc_cmake} ${hwloc_targ_SOURCE_DIR})
+
+    set(LIBHWLOC_INCLUDE_DIRS ${hwloc_targ_SOURCE_DIR}/include)
+    set(LIBHWLOC_LIBRARY_DIRS ${hwloc_targ_SOURCE_DIR}/Debug)
```
PatKamin, line 119: Does `Debug` means that hwloc is built in a debug mode?

.github/workflows/pr_push.yml
```diff
@@ -28,6 +28,13 @@ jobs:
             # Tests' building is off for a pure C build
             build_tests: 'OFF'
             simple_cmake: 'OFF'
+          # Enable static linking of HWLOC
+          - os: windows-latest
```
lukaszstolarczuk, line 32: does this job need to be in the FastBuild workflow?

CMakeLists.txt
lukaszstolarczuk, line 45: pls add this new option to README

lukaszstolarczuk, line 45: you could describe somewhere (e.g. as a comment here?) why would someone want to link hwloc statically

src/CMakeLists.txt
```diff
@@ -131,6 +131,10 @@ else()
         LIBS ${UMF_LIBS})
 endif()
 
+if(UMF_LINK_HWLOC_STATICALLY)
+    add_dependencies(umf hwloc)
```
lukaszstolarczuk, line 135: can't this be done right after the fetch (in the top-level CMakeLists)?

CMakeLists.txt
KFilipek: Done.

CMakeLists.txt
KFilipek: Done.

src/CMakeLists.txt
KFilipek: `umf` target is created under src/, but fetch is performed level up, where the target doesn't exist.

bratpiorka:
Please remove hwloc from your vcpackage installation if it is fetched and built by us.

CMakeLists.txt
bratpiorka: this path is only to be used on Windows - I don't think we even need to test this for Linux. @KFilipek could you add an additional "AND WINDOWS" to L#92 with an appropriate comment?

CMakeLists.txt
bratpiorka: done? I still see "master"

igchor:
@KFilipek could you add a test for this? I would like to see if having a binary that links to UMF statically works fine with `UMF_LINK_HWLOC_STATICALLY`  set to `1`. As I wrote in https://github.com/oneapi-src/unified-memory-framework/issues/513 I'm not sure if this will work straight away. If there are some tweaks needed, it would be good to know that.

CMakeLists.txt
KFilipek: I've marked here, what I've done locally, but not pushed until everything was fixed :)

CMakeLists.txt
KFilipek: We discussed it offline, only valid way to use it is Windows with release build, because of the requirement to not add extra CMake step with dependency preparation.
Done.

CMakeLists.txt
KFilipek: Corrected to Release

CMakeLists.txt
KFilipek: Done

KFilipek: I've talked with Rafał and it will be used in the very specific case, and probably nobody wants to use this feature.

.github/workflows/basic.yml
```diff
@@ -228,6 +228,57 @@ jobs:
         --umf-version ${{env.UMF_VERSION}}
         ${{ matrix.shared_library == 'ON' && '--shared-library' || ''}}
 
+  windows-build_hwloc:
+    name: "Windows + static hwloc"
+    env:
+      VCPKG_PATH: "${{github.workspace}}/build/vcpkg/packages/tbb_x64-windows;${{github.workspace}}/build/vcpkg/packages/jemalloc_x64-windows"
+      BUILD_DIR : "${{github.workspace}}/build/${{matrix.build_type}}"
+      INSTL_DIR : "${{github.workspace}}/../install-dir"
```
PatKamin, line 236: Remove, it's not used.

.github/workflows/basic.yml
```diff
@@ -228,6 +228,57 @@ jobs:
         --umf-version ${{env.UMF_VERSION}}
         ${{ matrix.shared_library == 'ON' && '--shared-library' || ''}}
 
+  windows-build_hwloc:
+    name: "Windows + static hwloc"
+    env:
+      VCPKG_PATH: "${{github.workspace}}/build/vcpkg/packages/tbb_x64-windows;${{github.workspace}}/build/vcpkg/packages/jemalloc_x64-windows"
```
PatKamin, line 234: Does jemalloc pool require hwloc?

.github/workflows/basic.yml
```diff
@@ -228,6 +228,57 @@ jobs:
...
+      BUILD_DIR : "${{github.workspace}}/build/${{matrix.build_type}}"
+      INSTL_DIR : "${{github.workspace}}/../install-dir"
+    strategy:
+      matrix:
+        os: ['windows-2019', 'windows-2022']
+        build_type: [Release]
+        shared_library: ['ON']
+        level_zero_provider: ['OFF']
```
PatKamin, line 242: I find it easier to read jobs when matrix includes only variables that actually have multiple values instead of having constant variables defined here.

.github/workflows/basic.yml
```diff
@@ -228,6 +228,57 @@ jobs:
...
+
+    - name: Install dependencies
+      run: vcpkg install
+      shell: pwsh # Specifies PowerShell as the shell for running the script.
+
+    - name: Configure build
+      run: >
+        cmake
+        -B ${{env.BUILD_DIR}}
+        -DCMAKE_PREFIX_PATH="${{env.VCPKG_PATH}}"
```
PatKamin, line 265: Not needed as UMF is not installed.

CMakeLists.txt
```diff
@@ -83,6 +90,47 @@ else()
...
+    pkg_check_modules(LIBHWLOC hwloc>=2.3.0)
+    if(NOT LIBHWLOC_FOUND)
+        find_package(LIBHWLOC 2.3.0 REQUIRED hwloc)
+    endif()
+    # add PATH to DLL on Windows
+    set(DLL_PATH_LIST
+        "${DLL_PATH_LIST};PATH=path_list_append:${LIBHWLOC_LIBRARY_DIRS}/../bin"
+    )
+else()
+    if(NOT WIN32)
```
PatKamin, line 103: We use WINDOWS instead of WIN32 everywhere in CMake files.

src/CMakeLists.txt
lukaszstolarczuk: heh, make sense, I've mistaken umf lib with umf project

CMakeLists.txt
```diff
@@ -83,6 +90,47 @@ else()
...
+        message(FATAL_ERROR "hwloc can be static linked only on Windows")
+    endif()
```
lukaszstolarczuk, line 105: as this only works on Release and we don't have an easy way to check this on Windows, you could add a message, similarly to proxy_lib (`"The proxy library will be built, however it is supported only in the Release build on Windows"`).

I'm not sure what are the consequences in Debug build - it won't work...?

igchor:
> @KFilipek could you add a test for this? (probably adding a job with `-DUMF_LINK_HWLOC_STATICALLY=1 -DUMF_BUILD_SHARED_LIBRARY=OFF` would be enough) I would like to see if having a binary that links to UMF statically works fine with `UMF_LINK_HWLOC_STATICALLY` set to `1`. As I wrote in #513 I'm not sure if this will work straight away. If there are some tweaks needed, it would be good to know that.

@KFilipek  It seems like this test scenario was removed in the latest push. Can you enable this again?

CMakeLists.txt
KFilipek: It won't build due to many workarounds to force it to build during configure+build time.
The only way is to add an extra step for building dependencies using ExternalProject_Add(), then the whole change will be limited to one function and Debug/Release will work like a charm. If you prefer such a solution, please, talk to Rafał and I can make a separate PR the same day.

PatKamin (approved):

bratpiorka (approved):

.github/workflows/basic.yml
```diff
@@ -228,6 +228,57 @@ jobs:
...
 
+  windows-build_hwloc:
+    name: "Windows + static hwloc"
+    env:
+      VCPKG_PATH: "${{github.workspace}}/build/vcpkg/packages/tbb_x64-windows;${{github.workspace}}/build/vcpkg/packages/jemalloc_x64-windows"
+      BUILD_DIR : "${{github.workspace}}/build/${{matrix.build_type}}"
+      INSTL_DIR : "${{github.workspace}}/../install-dir"
+    strategy:
+      matrix:
+        os: ['windows-2019', 'windows-2022']
```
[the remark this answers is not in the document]
lukaszstolarczuk: I don't think we need 4 extra jobs for static hwloc, but we can adjust this later on

lukaszstolarczuk (approved):

igchor (approved):

src/CMakeLists.txt
```diff
@@ -174,6 +178,6 @@ install(TARGETS umf EXPORT ${PROJECT_NAME}-targets)
 
 add_subdirectory(pool)
 
-if(UMF_PROXY_LIB_ENABLED)
+if(UMF_PROXY_LIB_ENABLED AND NOT UMF_LINK_HWLOC_STATICALLY)
```
lukaszstolarczuk, line 181: soo.. now if hwloc is linked statically the proxy_lib won't build...? we should add a message with such info (e.g. around the place where `UMF_PROXY_LIB_ENABLED` is set; or just here)

bratpiorka: when both proxy lib and UMF are linked statically with hwloc, we have a problem with doubling the hwloc symbols. But in the scenario that we want to enable the proxy lib wouldn't be used

lukaszstolarczuk: status message shouldn't hurt, I guess

also, we can write what you just wrote as a comment in this line

CMakeLists.txt
```diff
@@ -83,6 +90,44 @@ else()
...
+
+    FetchContent_GetProperties(hwloc_targ)
+    if(NOT hwloc_targ_POPULATED)
+        FetchContent_MakeAvailable(hwloc_targ)
+    endif()
+
+    set(LIBHWLOC_INCLUDE_DIRS
+        ${hwloc_targ_SOURCE_DIR}/include;${hwloc_targ_BINARY_DIR}/include)
+    set(LIBHWLOC_LIBRARY_DIRS
+        ${hwloc_targ_SOURCE_DIR}/Release;${hwloc_targ_SOURCE_DIR}/Debug)
```
PatKamin, line 124: It's the `Release` directory we're looking for. Perhaps you meant:
```
${hwloc_targ_SOURCE_DIR}/Release;${hwloc_targ_BINARY_DIR}/Release)
```

PatKamin (approved):

PatKamin (approved):

src/CMakeLists.txt
KFilipek: I've added earlier

lukaszstolarczuk (approved):

PR #544, opened 2024-06-12
Minor fix in util_atomic_store_release

vinser52:
I am not sure why it was not detected during the compilation.

### Checklist

- [x] All tests pass locally


lukaszstolarczuk (approved):

bratpiorka:
it was not detected because the only place where this macro is used is:
```
static void store(void *dst, void *src) {
    util_atomic_store_release((word *)dst, (word)src);
}
```
so there is a "dst" in the scope. BTW - this is why I hate macros.

bratpiorka (approved):

Issue #543, opened 2024-06-12
Link error when building umf_test-init_teardown test


vinser52:
The issue happens when UMF is built as a shared library.

We already had a similar issue #517. It was fixed by PR #518. I think we need to extend our CI to catch such issues.

## Environment Information

- UMF version (hash commit or a tag): main branch
- OS(es) version(s): Ubuntu 20.04
- kernel version(s): 5.4.0-146-generic
- compiler, libraries, and other related tools version(s): gcc (Ubuntu 9.4.0-1ubuntu1~20.04.2) 9.4.0, GLIBC 2.31-0ubuntu9.15

## Please provide a reproduction of the bug:

```
cmake .. -DUMF_BUILD_SHARED_LIBRARY=ON
make
```

## How often bug is revealed:

always

## Actual behavior:

link error:
```
[ 97%] Linking CXX executable umf_test-init_teardown
/usr/bin/ld: CMakeFiles/umf_test-init_teardown.dir/test_init_teardown.c.o: in function `load_symbol':
test_init_teardown.c:(.text+0xb): undefined reference to `dlsym'
/usr/bin/ld: CMakeFiles/umf_test-init_teardown.dir/test_init_teardown.c.o: in function `run_test':
test_init_teardown.c:(.text+0xcd): undefined reference to `dlopen'
/usr/bin/ld: test_init_teardown.c:(.text+0x446): undefined reference to `dlclose'
/usr/bin/ld: test_init_teardown.c:(.text+0x4ea): undefined reference to `dlclose'
collect2: error: ld returned 1 exit status
make[2]: *** [test/CMakeFiles/umf_test-init_teardown.dir/build.make:101: test/umf_test-init_teardown] Error 1
make[1]: *** [CMakeFiles/Makefile2:1809: test/CMakeFiles/umf_test-init_teardown.dir/all] Error 2
make: *** [Makefile:146: all] Error 2
```

## Expected behavior:

Build successes.

## Details

The issue happens because umf_utils uses dlopen. Starting GLIBC 2.34 the libpthread and the libdl are parts of glibc library (no explicit linking with libpthread and libdl is needed). But on the systems with GLIBC < 2.34 binary that uses dlopen/dlsym/dlclose should explicitly link with libdl

PR #545, opened 2024-06-13
Link the init_teardown test with libdl

ldorau:
### Description

Link the init_teardown test with libdl explicitly
for systems with glibc < 2.34.

The init_teardown test uses dlopen(), dlsym() and dlclose().

Fixes: #543

### Checklist
- [x] All tests pass locally

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -287,7 +287,10 @@ endif()
 if(LINUX
    AND UMF_BUILD_SHARED_LIBRARY
    AND UMF_POOL_SCALABLE_ENABLED)
-    add_umf_test(NAME init_teardown SRCS test_init_teardown.c)
+    add_umf_test(
+        NAME init_teardown
+        SRCS test_init_teardown.c
+        LIBS dl)
     # append LD_LIBRARY_PATH to the libumf
     set_property(
         TEST umf-init_teardown


kswiecicki (approved):

lukaszstolarczuk:
perhaps to address the issue, we should also switch "simplest CMake" in FastBuild to run on ubuntu 20.04 (instead of current ubuntu latest)

ldorau:
> perhaps to address the issue, we should also switch "simplest CMake" in FastBuild to run on ubuntu 20.04 (instead of current ubuntu latest)

@lukaszstolarczuk @bratpiorka Done

test/CMakeLists.txt
```diff
@@ -287,7 +287,10 @@ endif()
 if(LINUX
    AND UMF_BUILD_SHARED_LIBRARY
    AND UMF_POOL_SCALABLE_ENABLED)
-    add_umf_test(NAME init_teardown SRCS test_init_teardown.c)
+    add_umf_test(
+        NAME init_teardown
+        SRCS test_init_teardown.c
+        LIBS dl)
```
vinser52, line 293: When we had a similar issue in the past, @bratpiorka fixed it by adding `dl` dependency to utils static library. Why a similar fix does not work here?
If `init-teardown` test uses `utils`, should it be linked with `utils`?

ldorau: The `init-teardown` test does not use `utils`.

vinser52 (approved):

ldorau:
> perhaps to address the issue, we should also switch "simplest CMake" in FastBuild to run on ubuntu 20.04 (instead of current ubuntu latest)

I am afraid it is not so simple - Ubuntu 20.04 has too old HWLOC ... :-(
https://github.com/oneapi-src/unified-memory-framework/actions/runs/9505967029/job/26202100344
```
-- Checking for module 'hwloc>=2.3.0'
--   Requested 'hwloc >= 2.3.0' but version of hwloc is 2.1.0
-- Checking for module 'libhwloc' using find_library()
--   Found libhwloc using find_library()
--     LIBHWLOC_LIBRARIES = /usr/lib/x86_64-linux-gnu/libhwloc.so
--     LIBHWLOC_INCLUDE_DIRS = /usr/include
--     LIBHWLOC_LIBRARY_DIRS = /usr/lib/x86_64-linux-gnu
--     LIBHWLOC_API_VERSION = 2.1.0
CMake Error at cmake/FindLIBHWLOC.cmake:61 (message):
      Required version: 2.3.0, found 2.1.0
```

lukaszstolarczuk (approved):
as disccussed, we'll use `install_hwloc.sh`, but to be done on a sep. PR.

PR #550, opened 2024-06-14
Print UMF version in umfInit() when env var UMF_VERSION=1

ldorau:
### Description

Print UMF version in `umfInit()` when env var `UMF_VERSION=1`

### Checklist
- [x] All tests pass locally

### DIFF
--- a/src/libumf.c
+++ b/src/libumf.c
@@ -45,3 +45,5 @@ void umfTearDown(void) {
         umf_ba_destroy_global();
     }
 }
+
+int umfGetCurrentVersion(void) { return UMF_VERSION_CURRENT; }
--- a/src/libumf.c
+++ b/src/libumf.c
@@ -9,10 +9,13 @@
 
 #include <stddef.h>
 
+#include "umf.h"
+
 #include "base_alloc_global.h"
 #include "memspace_internal.h"
 #include "provider_tracking.h"
 #include "topology.h"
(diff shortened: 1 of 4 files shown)


src/libumf.c
```diff
@@ -21,6 +24,11 @@ static unsigned long long umfRefCount = 0;
 
 int umfInit(void) {
     if (util_fetch_and_add64(&umfRefCount, 1) == 0) {
+        if (util_env_var_has_str("UMF_VERSION", "1")) {
+            int umf_ver = umfGetCurrentVersion();
+            printf("UMF_VERSION: %i.%i\n", UMF_MAJOR_VERSION(umf_ver),
```
bratpiorka, line 29: why we can't use our logging mechanism here? please do not use printf.

bratpiorka:
as I understand it, here we want to behave in the same way as OpenMP with their environment variable KMP_VERSION and TCM with TCM_VERSION etc. ?

src/libumf.c
ldorau: Done: https://github.com/oneapi-src/unified-memory-framework/pull/552

ldorau:
> as I understand it, here we want to behave in the same way as OpenMP with their environment variable KMP_VERSION and TCM with TCM_VERSION etc. ?

Yes, but I try to move it to OpenMP, so closing it for now ...

PR #551, opened 2024-06-16
load DLLs only from safe load list

bratpiorka:
Set /DEPENDENTLOADFLAG to 2000 (LOAD_LIBRARY_SAFE_CURRENT_DIRS) in Windows umf.dll builds (MSVC only). 
With this flag, loading a DLL for execution from the current directory is only allowed if it is under a directory in the Safe load list.

In addition, there is a new test written in Power Shell that uses dumpbin.exe to check that this flag is set correctly.

Detailed information about the flag can be found at: https://learn.microsoft.com/en-us/cpp/build/reference/dependentloadflag?view=msvc-170
and: https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-loadlibraryexa

- [x] requires #534

### DIFF
--- a/.github/scripts/check_dll_flags.ps1
+++ b/.github/scripts/check_dll_flags.ps1
@@ -1,0 +1,42 @@
+# Copyright (C) 2023-2024 Intel Corporation
+# Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# Invoke-CmdScript runs a command script and updates the current environment 
+# with any flag changes set by the script.
+function Invoke-CmdScript {
+  param(
+    [String] $scriptName
+  )
+  $cmdLine = """$scriptName"" $args & set"
+  & $Env:SystemRoot\system32\cmd.exe /c $cmdLine |
+  select-string '^([^=]*)=(.*)$' | foreach-object {
+    $varName = $_.Matches[0].Groups[1].Value
+    $varValue = $_.Matches[0].Groups[2].Value
+    set-item Env:$varName $varValue
+  }
+}
+
+# Get the path to vcvarsall.bat
+$vsPath = & "${env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer\vswhere.exe" `
+  -latest -products * -requires Microsoft.VisualStudio.Component.VC.Tools.x86.x64 `
+  -property installationPath
+$vsPath = "$vsPath\VC\Auxiliary\Build"
+$vcvarsall = "${vsPath}\vcvarsall.bat"
+echo "Visual Studio path: $vsPath"
+echo "vcvarsall.bat path: $vcvarsall"
+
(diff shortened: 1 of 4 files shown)


.github/scripts/check_dll_flags.ps1
```diff
@@ -0,0 +1,46 @@
...
+$vcvarsall = "${vsPath}\vcvarsall.bat"
+echo "Visual Studio path: $vsPath"
+echo "vcvarsall.bat path: $vcvarsall"
+
+# Call vcvarsall.bat so we can run MSVC commands
+echo "Setting up MSVC environment..."
+Invoke-CmdScript "$vcvarsall" x86
+
+echo "$args"
+echo "$($args[0] | Out-String)"
```
PatKamin, line 34: Don't we have to set and check this flag also for the umf_proxy.dll?

bratpiorka: done

cmake/helpers.cmake
```diff
@@ -136,11 +136,17 @@ function(add_umf_target_link_options name)
             endif()
         endif()
     elseif(MSVC)
+        message(STATUS "C_COMPILER_ID: ${C_COMPILER_ID}")
+        message(STATUS "CXX_COMPILER_ID: ${CXX_COMPILER_ID}")
+        message(STATUS "CMAKE_C_COMPILER_ID: ${CMAKE_C_COMPILER_ID}")
+        message(STATUS "CMAKE_CXX_COMPILER_ID: ${CMAKE_CXX_COMPILER_ID}")
```
PatKamin, line 142: nit: These messages potentially can be useful for other CMake generators. Could you move them to some common space, maybe in the main CMakeLists.txt?

PatKamin (approved):

.github/scripts/check_dll_flags.ps1
```diff
@@ -0,0 +1,42 @@
...
+echo "vcvarsall.bat path: $vcvarsall"
+
+# Call vcvarsall.bat so we can run MSVC commands
+echo "Setting up MSVC environment..."
+Invoke-CmdScript "$vcvarsall" x86
+
+# Get umf.dll configuration flags and check if DEPENDENTLOADFLAG is set to 0x2000
+$flags = & "${env:VCToolsInstallDir}\bin\Hostx64\x64\dumpbin.exe" /LOADCONFIG "$args"
+if (($flags | Where-Object { $_ -match '(\d+).*Dependent' } | ForEach-Object { $matches[1] } ) -eq "2000") { 
+  echo "The DEPENDENTLOADFLAG is correclty set to 0x2000."
```
lplewa, line 36: correctly

bratpiorka: done

cmake/helpers.cmake
bratpiorka: these were only debug prints - removed

lplewa (approved):

PR #563, opened 2024-06-24
Print UMF source version in util_log_init() if it is defined

ldorau:
### Description

Add `UMF_SRC_VERSION` to `UMF_PRIVATE_COMPILE_DEFINITIONS`.
Print UMF source version (`UMF_SRC_VERSION`) in `util_log_init()` if it is defined.

It looks like:
```
[PID:21521  TID:21521  INFO  UMF] util_log_init: Logger enabled (umf version: 0.9, src version: 0.1.0.git.d1e2930, level: DEBUG, flush: DEBUG, pid: yes, timestamp: no)
```

### Checklist
- [x] All tests pass locally

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -76,7 +76,7 @@ set(UMF_SOURCES_WINDOWS libumf_windows.c)
 # Compile definitions for UMF library.
 #
 # TODO: Cleanup the compile definitions across all the CMake files
-set(UMF_PRIVATE_COMPILE_DEFINITIONS "")
+set(UMF_PRIVATE_COMPILE_DEFINITIONS "-DUMF_SRC_VERSION=${UMF_SRC_VERSION}")
 
 set(UMF_SOURCES_COMMON_LINUX_MACOSX
     provider/provider_os_memory.c
--- a/src/utils/utils_log.c
+++ b/src/utils/utils_log.c
@@ -293,10 +293,19 @@ void util_log_init(void) {
         loggerConfig.flushLevel = LOG_FATAL;
     }
 
+#ifdef UMF_SRC_VERSION
+// convert a define to a C string
+#define STR_(X) #X
+#define STR(X) STR_(X)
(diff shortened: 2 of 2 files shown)


lplewa:
Do we need separate src version ?- For me it is strange. IMHO for not release version we should have umf version defined according to https://semver.org/ 

0.9.<NR_commits from this tag>+<src version/git describe/whatever>

ldorau:
> Do we need separate src version ?- For me it is strange. IMHO for not release version we should have umf version defined according to https://semver.org/ i.e.:
> 
> 0.9.0-<NR_commits from this tag>+<src version/git describe/whatever>

As I understand the "UMF version" can be for example "1.0" and may not contain a commit hash, while the "src version" should always contain a commit hash.

ldorau:
> Do we need separate src version ?

They both already are present on the main branch, see an output of the `cmake ..` command:
```
-- UMF version: 0.1.0 (source version: 16d6673)
...
```

ldorau:
> > Do we need separate src version ?- For me it is strange. IMHO for not release version we should have umf version defined according to https://semver.org/ i.e.:
> > 0.9.0-<NR_commits from this tag>+<src version/git describe/whatever>
> 
> As I understand the "UMF version" can be for example "1.0" and may not contain a commit hash, while the "src version" should always contain a commit hash.

But I think that we should unify that and use only one version of version ;-) according to https://semver.org/ i.e.:

> 0.9.0-<NR_commits from this tag>+<src version/git describe/whatever>

lukaszstolarczuk:
> > 0.9.0-<NR_commits from this tag>+<src version/git describe/whatever>
> 
> To be able to do that, we need a tag, I suppose ;-)

I'm working on it 😉 

Anyway, we can use SRCVERSION now and @lplewa - you can file an issue to fix this (for sure before 1.0.0 version)

lukaszstolarczuk (approved):

PR #588, opened 2024-07-02
optimize qemu workflow

lplewa:
In short this PR merges all qemu workflow in one. We can reuse the same image and build project only one time. 
We reduce Runner utilization from ~55min to 25 min (+ we gain time wasted for runner to pickup work and start, which is not counted)
As a downside we reduce parallelize - we had 11 -  5 min jobs  now we have a single 25 min long job. 
It should not be a problem, as we are blocked by runner availability, so qemu workflow should finish anyway before all other jobs. But to limit negative influence of this change, i changed qemu workflow dependency, to start it sooner, and do not block other workflows.

### DIFF
--- a/.github/workflows/qemu.yml
+++ b/.github/workflows/qemu.yml
@@ -4,7 +4,7 @@ name: Qemu
 on: workflow_call
 
 env:
-  CI_BRANCH : "${{ github.head_ref || github.ref_name }}"
+  CI_BRANCH: "${{ github.head_ref || github.ref_name }}"
 
 permissions:
   contents: read
@@ -13,19 +13,6 @@ jobs:
   qemu-build:
     name: Qemu
     runs-on: ubuntu-22.04
-    strategy:
-      matrix:
-        config: [{name: 'default', hmat: 'on'},
-                 { name: 'sock_2_var1', hmat: 'off'},
-                 { name: 'sock_2_var1_hmat', hmat: 'on'},
-                 { name: 'sock_2_var2', hmat: 'off'},
-                 { name: 'sock_2_var2_hmat', hmat: 'on'},
-                 { name: 'sock_2_var3', hmat: 'off'},
-                 { name: 'sock_2_var3_hmat', hmat: 'on'},
-                 { name: 'sock_4_var1', hmat: 'off'},
-                 { name: 'sock_4_var1_hmat', hmat: 'on'},
-                 { name: 'sock_4_var2', hmat: 'off'},
-                 { name: 'sock_4_var2_hmat', hmat: 'on'}]
 
     steps:
     - name: Checkout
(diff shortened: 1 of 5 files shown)


scripts/qemu/run-tests.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+# Copyright (C) 2024 Intel Corporation
+# Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+set -x
+set -e
+
```
lukaszstolarczuk, line 8: pls move `numactl -H` in here, or into `start_qemu.sh`

lukaszstolarczuk, line 8: same for:
```
# Set ptrace value for IPC test
echo password | sudo bash -c "echo 0 > /proc/sys/kernel/yama/ptrace_scope"
```

it resets to default value after shutdown

scripts/qemu/run-build.sh
```diff
@@ -34,17 +34,3 @@ cmake .. \
     -DUMF_BUILD_EXAMPLES=ON
```
lukaszstolarczuk, line 35: pls add `-DUMF_TESTS_FAIL_ON_SKIP=ON` - right now we missed the skipped ipc test

.github/workflows/qemu.yml
```diff
@@ -83,38 +70,39 @@ jobs:
...
+        scp -P 2222 ${{github.workspace}}/scripts/qemu/run-tests.sh cxltest@127.0.0.1:/home/cxltest
         ssh cxltest@127.0.0.1 -p 2222 -t "bash /home/cxltest/run-build.sh https://github.com/$CI_REPO ${{env.CI_BRANCH}}"
+
+        ssh cxltest@127.0.0.1 -p 2222 -t "sudo shutdown -h now"
+
+    - name: Run tests
+      run: |
+        for config_file in scripts/qemu/configs/*.xml; do
+          config_name=$(basename $config_file .xml)
+
```
lukaszstolarczuk, line 93: pls add an empty line print and some "header" like
```

### <config_name> START ###
```
right now, the log is quite long and it's a little hard to find exact configuration you're looking for

scripts/qemu/run-tests.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+# Copyright (C) 2024 Intel Corporation
+# Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+set -x
```
lukaszstolarczuk, line 6: not sure if `-x` is needed here

scripts/qemu/run-tests.sh
lplewa: done

lplewa: done

scripts/qemu/run-build.sh
lplewa: done

.github/workflows/qemu.yml
lplewa: done

scripts/qemu/run-tests.sh
lplewa: done

lukaszstolarczuk (approved):

KFilipek (approved):
I have mixed emotions with such big log, but I can close my eyes a little bit and give "looks good to me".

PR #591, opened 2024-07-02
do not open symlinks

bratpiorka:
### Description

For security reasons, never open symbolic links.

### Checklist

- [ ] All tests pass locally

### DIFF
--- a/src/utils/utils_common.h
+++ b/src/utils/utils_common.h
@@ -29,8 +29,10 @@ extern "C" {
         expression;                                                            \
     } while (0)
 
-#define ALIGN_UP(value, align) (((value) + (align)-1) & ~((align)-1))
-#define ALIGN_DOWN(value, align) ((value) & ~((align)-1))
+#ifndef ALIGN_UP
+#define ALIGN_UP(value, align) (((value) + (align) - 1) & ~((align) - 1))
+#define ALIGN_DOWN(value, align) ((value) & ~((align) - 1))
+#endif
 
 #define VALGRIND_ANNOTATE_NEW_MEMORY(p, s) DO_WHILE_EMPTY
 #define VALGRIND_HG_DRD_DISABLE_CHECKING(p, s) DO_WHILE_EMPTY
@@ -84,6 +86,9 @@ int utils_close_fd(int fd);
 
 // obtain a duplicate of another process's file descriptor
 umf_result_t utils_duplicate_fd(int pid, int fd_in, int *fd_out);
+
+// checks whether the path points to a symlink
+int util_is_symlink(const char *path);
 
 #ifdef __cplusplus
 }
--- a/src/utils/utils_log.c
+++ b/src/utils/utils_log.c
@@ -254,6 +254,16 @@ void util_log_init(void) {
 
         memcpy(file, arg, len);
         file[len] = '\0';
+
+        // for security reasons check if we don't open a symlink file
+        if (util_is_symlink(file)) {
+            loggerConfig.output = stderr;
+            LOG_ERR("Output file %s exists and is a symlink - logging disabled",
+                    file);
+            loggerConfig.output = NULL;
+            return;
+        }
+
         loggerConfig.output = fopen(file, "w+");
         if (!loggerConfig.output) {
             loggerConfig.output = stderr;
--- a/src/utils/utils_posix_common.c
+++ b/src/utils/utils_posix_common.c
@@ -10,7 +10,9 @@
 #include <errno.h>
 #include <stdlib.h>
 #include <string.h>
+#include <sys/stat.h>
 #include <sys/syscall.h>
+#include <sys/types.h>
 #include <unistd.h>
 
 #include "utils_common.h"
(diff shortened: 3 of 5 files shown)


src/utils/utils_common.h
```diff
@@ -29,8 +29,10 @@ extern "C" {
         expression;                                                            \
     } while (0)
 
-#define ALIGN_UP(value, align) (((value) + (align)-1) & ~((align)-1))
-#define ALIGN_DOWN(value, align) ((value) & ~((align)-1))
+#ifndef ALIGN_UP
+#define ALIGN_UP(value, align) (((value) + (align) - 1) & ~((align) - 1))
+#define ALIGN_DOWN(value, align) ((value) & ~((align) - 1))
+#endif
```
lplewa, line 35: This looks like random change not related to this PR

src/utils/utils_log.c
```diff
@@ -254,6 +254,16 @@ void util_log_init(void) {
...
+
+        // for security reasons check if we don't open a symlink file
+        if (util_is_symlink(file)) {
+            loggerConfig.output = stderr;
+            LOG_ERR("Output file %s exists and is a symlink - logging disabled",
+                    file);
+            loggerConfig.output = NULL;
+            return;
+        }
+
```
lplewa, line 266: We open for write only here - i'm pretty sure that is it not an security issue.

src/utils/utils_windows_common.c
```diff
@@ -46,3 +46,56 @@ umf_result_t utils_duplicate_fd(int pid, int fd_in, int *fd_out) {
...
+        BYTE buffer[MAXIMUM_REPARSE_DATA_BUFFER_SIZE];
+        DWORD returnedLength;
+
+        // Query the reparse data
+        BOOL result = DeviceIoControl(
+            fileHandle, FSCTL_GET_REPARSE_POINT, NULL, 0, buffer,
+            MAXIMUM_REPARSE_DATA_BUFFER_SIZE, &returnedLength, NULL);
+        if (!result) {
+            // Handle error, could not read reparse point data
+            CloseHandle(fileHandle);
```
lplewa, line 80: you can CloseHandle before if - and remove all other closeHandle in other parts of the code.

src/utils/utils_posix_common.c
```diff
@@ -104,3 +106,8 @@ umf_result_t utils_duplicate_fd(int pid, int fd_in, int *fd_out) {
     return UMF_RESULT_SUCCESS;
 #endif
 }
+
+int util_is_symlink(const char *path) {
+    struct stat st;
+    return lstat(path, &st) == 0 && S_ISLNK(st.st_mode);
```
lplewa, line 112: please create consistent return values in this function.
On windows it is -1 in case of error, and 1 in case of symlink and 0 otherwise 
on linux error in not distinguished from not a symlink

src/utils/utils_windows_common.c
```diff
@@ -46,3 +46,56 @@ umf_result_t utils_duplicate_fd(int pid, int fd_in, int *fd_out) {
...
+
+int util_is_symlink(const char *path) {
+    DWORD attributes = GetFileAttributesA(path);
+    if (attributes == INVALID_FILE_ATTRIBUTES) {
+        // Handle error, could not get file attributes
+        return -1;
+    }
+
+    // Check if the file is a reparse point
+    if (attributes & FILE_ATTRIBUTE_REPARSE_POINT) {
```
lplewa, line 58: you can reverse this if condition, as else cause has only return.

lplewa (changes requested):
Please refactor and improve error logs. util_open_library fails user get error about TBB not found. It is important to LOG that we couldn't open it as it was a symlink vs that library was not found.

lplewa:
Also i'm 99% sure that it will not work on linux - dlopen() works on file name (not on path unless it contains '/') - so if you want to call stat() on file opened you have to find a path to if first and that call stat() on full file path. 

I guess on windows it might be the same.

src/utils/utils_windows_common.c
```diff
@@ -46,3 +46,56 @@ umf_result_t utils_duplicate_fd(int pid, int fd_in, int *fd_out) {
     (void)fd_out; // unused
     return UMF_RESULT_ERROR_NOT_SUPPORTED;
 }
+
+int util_is_symlink(const char *path) {
+    DWORD attributes = GetFileAttributesA(path);
```
lplewa, line 51: we do not use A sufixed functions in this project.

src/utils/utils_common.h
bratpiorka: we define ALIGN_UP in two places - here and in test/common/test_helpers.h

src/utils/utils_log.c
bratpiorka: according to security guidelines we should never open symlinks

pbalcer (commented):
I disagree with this approach. Symlinks are an OS feature and we can't just say they are "unsafe". Their use may be unsafe (just like the use of everything else), but symlinks themselves are a very commonly used feature that we cannot simply disable.

In PMDK, we used 'realpath' to resolve user-provided paths if we needed to store them.

PR #602, opened 2024-07-05
[CI] Clean workspace on self-hosted runners on start

lukaszstolarczuk:
We need to clean the workspace on self-hosted runners before executing build, because it may contain old builds, which poison the current build.


.github/workflows/pr_push.yml
```diff
@@ -15,87 +15,5 @@ permissions:
   contents: read
 
 jobs:
-  CodeStyle:
```
lplewa, line 18: Do you intend this?

lukaszstolarczuk: haha, rather so😄

I'm testing the GPU workflow only

lukaszstolarczuk:
cleaning workspace does not solve the issue, it still occurs from time to time, with error:

```
CUSTOMBUILD : CMake error : Cannot restore timestamp "C:/Users/Administrator/repo/actions-runner/_work/unified-memory-framework/unified-memory-framework/build/test/CMakeFiles/generate.stamp": Access is denied. [C:\Users\Administrator\repo\actions-runner\_work\unified-memory-framework\unified-memory-framework\build\test\umf_test-jemalloc_pool.vcxproj]
```

we'd have to investigate this further to avoid these random fails

KFilipek:
As we talked few weeks ago, it should be done only via: https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/running-scripts-before-or-after-a-job 
Hooks will always clean the working directory before/after.
Additionally, this solution is done purely in infrastructure, nothing to do with PR.

lukaszstolarczuk:
This change does not fix the issue as we expected. Closing it for now.

PR #599, opened 2024-07-04
Add standalone tests of examples

ldorau:
### Description

Add standalone tests of examples.

Fixes: #427

### Checklist
- [x] Code compiles without errors locally
- [x] All tests pass locally
- [x] CI workflows execute properly

PR #611, opened 2024-07-15
Do not run cmake inside ctest tests

ldorau:
### Description

1) Set `CMAKE_INSTALL_PREFIX` in all CI builds.
2) Clean up setting and using of `env.BUILD_DIR`.
3) Do not run cmake inside ctest tests.

Ref: https://github.com/oneapi-src/unified-memory-framework/pull/599/files#r1670260245

### Checklist
- [x] All tests pass locally

### DIFF
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -9,6 +9,8 @@ permissions:
 env:
   # for installation testing - it should match with version set in CMake
   UMF_VERSION: 0.1.0
+  BUILD_DIR : "${{github.workspace}}/build"
+  INSTL_DIR : "${{github.workspace}}/../install-dir"
 
 jobs:
   ubuntu-build:
@@ -67,9 +69,6 @@ jobs:
             shared_library: 'ON'
             level_zero_provider: 'ON'
             install_tbb: 'OFF'
-    env:
-      BUILD_DIR : "${{github.workspace}}/build/"
-      INSTL_DIR : "${{github.workspace}}/../install-dir"
     runs-on: ${{matrix.os}}
 
     steps:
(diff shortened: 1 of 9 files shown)


.github/workflows/basic.yml
```diff
@@ -148,8 +150,6 @@ jobs:
     name: Windows
     env:
       VCPKG_PATH: "${{github.workspace}}/build/vcpkg/packages/hwloc_x64-windows;${{github.workspace}}/build/vcpkg/packages/tbb_x64-windows;${{github.workspace}}/build/vcpkg/packages/jemalloc_x64-windows"
```
bratpiorka, line 152: we could use ${{env.BUILD_DIR}} here

.github/workflows/fast.yml
```diff
@@ -6,6 +6,10 @@ on: workflow_call
 permissions:
   contents: read
 
+env:
+  BUILD_DIR : "${{github.workspace}}/build"
```
bratpiorka, line 10: we could use BUILD_DIR in VCPKG paths

bratpiorka (changes requested):

.github/workflows/basic.yml
```diff
@@ -131,6 +130,9 @@ jobs:
         ${{ matrix.compiler.cxx == 'icpx' && '. /opt/intel/oneapi/setvars.sh &&' || ''}} 
         ctest --output-on-failure --test-dir test
 
+    - name: Remove the installation directory
```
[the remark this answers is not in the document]
ldorau: We do not have to, because the `umf_standalone_examples` test is run only on Linux.

.github/workflows/basic.yml
ldorau: We cannot, it would be a syntax error, because GHA does not allow to use env vars inside a definition of another env var.

.github/workflows/fast.yml
ldorau: We cannot, it would be a syntax error, because GHA does not allow to use env vars inside a definition of another env var.

bratpiorka: ok I see, thanks

bratpiorka (approved):

lukaszstolarczuk (approved):

PR #615, opened 2024-07-16
[CI] Add branch context to the Coverity

KFilipek:
### Description

Since this PR Coverity can also be executed
for another branch. Previous solutions always
cloned the main branch.

### DIFF
--- a/.github/workflows/coverity.yml
+++ b/.github/workflows/coverity.yml
@@ -22,6 +22,8 @@ jobs:
     steps:
     - name: Checkout repository
       uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      with: 
+        ref: ${{ github.ref }}
 
     - name: Install apt packages
       run: |


bratpiorka (approved):

.github/workflows/coverity.yml
```diff
@@ -22,6 +22,8 @@ jobs:
     steps:
     - name: Checkout repository
       uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      with: 
+        ref: ${GITHUB_REF_NAME}
```
lukaszstolarczuk, line 26: hmm... I just pushed this change to my repo and it seems not to work - perhaps a wrong variable in this context...?

https://github.com/lukaszstolarczuk/unified-memory-framework/actions/runs/9954564151/job/27500462789#step:2:60

lukaszstolarczuk (changes requested):

.github/workflows/coverity.yml
KFilipek: You're right, the `${GITHUB_REF_NAME}` can be read only during execution, in workflow we should use another approach. Fixed.

lukaszstolarczuk (approved):

Issue #565, opened 2024-06-25
UMF versioning is doubled (and not compatibile with semver.)


lplewa:
**Current State:**

We are currently using two different versioning schemes in our project:
- Main Version: This version is hardcoded in the CMake project command.
- Git Version (called "source version"): This version, compliant with Semantic Versioning (SemVer), is generated separately based on git-describe.

**Expected state:**

We should have a single versioning scheme, based on Git version which is compliant with SemVer

ref: https://semver.org/

lukaszstolarczuk:
When fixing this issue we should also take into considereration this issue: https://github.com/oneapi-src/unified-memory-framework/issues/466

lukaszstolarczuk:
Proposed solution to this problem is in: https://github.com/oneapi-src/unified-memory-framework/pull/616

We set CMake version just to use it for SOVERSION and to generate CMake FindUMF.cmake/config file.
`UMF_VERSION` after the change will be aligned to semver and used solely in the project.

PR #616, opened 2024-07-16
[CMake] Update versioning

lukaszstolarczuk:
Properly set version based on our git tags.
For CMake's sake we use major.minor.patch
For Windows dll metadata we use major.minor.build.revision and additional variables.

Print the dll's metadata, to check correctness.

Fixes: https://github.com/oneapi-src/unified-memory-framework/issues/565


bratpiorka:
shouldn't you change "UMF_VERSION" in basic.yml line 11?

lukaszstolarczuk:
> shouldn't you change "UMF_VERSION" in basic.yml line 11?

heh, it's the other way around - I forgot that by default in CI, we have a shallow clone. It caused no git history, so no tags, so the version was improperly set to "0.0.0" (instead of expected 0.1.0).

CMakeLists.txt
```diff
@@ -3,32 +3,43 @@
...
+    LANGUAGES C)
+# BUGFIX = "true" if Patch version > 0
+if(CMAKE_PROJECT_VERSION_PATCH GREATER 0)
+    set(UMF_VERSION_BUGFIX TRUE)
+endif()
+
+message(
+    STATUS
+        "UMF CMake version: ${UMF_CMAKE_VERSION} (via CMake: ${CMAKE_PROJECT_VERSION})\n"
+        "UMF version: ${UMF_VERSION}\n"
```
ldorau, line 32: Please print each line in a separate `message(STATUS ...` because now it looks like:
```
-- GIT_VERSION: v0.1.0-28-g4ab61a7
-- UMF CMake version: 0.1.0 (via CMake: 0.1.0)
UMF version: 0.1.0-git28.g4ab61a7
UMF_VERSION_REVISION: 28
UMF_VERSION_PRERELEASE: FALSE
UMF_VERSION_PRIVATE: TRUE
UMF_VERSION_BUGFIX: FALSE
```

ldorau (changes requested):
What about `UMF_VERSION_CURRENT` ? It is still `0.9` ...
```
util_log_init: Logger enabled (umf version: 0.9, UMF version: 0.1.0-git28.g4ab61a7,
```

src/utils/utils_log.c
```diff
@@ -305,18 +305,18 @@ void util_log_init(void) {
...
-#endif /* !UMF_SRC_VERSION */
+#define STR_UMF_VERSION "UMF version: " STR(UMF_VERSION) ", "
+#else /* !UMF_VERSION */
+#define STR_UMF_VERSION ""
+#endif /* !UMF_VERSION */
 
     int umf_ver = umfGetCurrentVersion();
     LOG_INFO(
-        "Logger enabled (umf version: %i.%i, " STR_UMF_SRC_VERSION
+        "Logger enabled (umf version: %i.%i, " STR_UMF_VERSION
```
ldorau, line 319: It looks very strange now:
```
util_log_init: Logger enabled (umf version: 0.9, UMF version: 0.1.0-git28.g4ab61a7,
```

util_log_init: Logger enabled (**umf version**: 0.9, **UMF version**: 0.1.0-git28.g4ab61a7,

ldorau (changes requested):

CMakeLists.txt
lukaszstolarczuk: we shouldn't care - these messages are temporary, for debugging purpose, to check if all vars are set properly.

src/utils/utils_log.c
lukaszstolarczuk: from the 0.9.0 release this will align - I updated RELEASE_STEPS to update this manually for each major and minor release

ldorau: OK, but what about "`umf version`" and the second "`UMF version`"

lukaszstolarczuk: we can rename the first one to "API version" or simply remove the first part (leaving just the full UMF_VERSION) - what do you think...?

src/utils/utils_log.c
```diff
@@ -305,18 +305,18 @@ void util_log_init(void) {
...
 // convert a define to a C string
 #define STR_(X) #X
 #define STR(X) STR_(X)
-#define STR_UMF_SRC_VERSION "src version: " STR(UMF_SRC_VERSION) ", "
-#else /* !UMF_SRC_VERSION */
-#define STR_UMF_SRC_VERSION ""
-#endif /* !UMF_SRC_VERSION */
+#define STR_UMF_VERSION "UMF version: " STR(UMF_VERSION) ", "
+#else /* !UMF_VERSION */
+#define STR_UMF_VERSION ""
```
ldorau, line 314:
```c
#else /* !UMF_VERSION */
#error UMF_VERSION not defined!
#endif /* !UMF_VERSION */
```

src/utils/utils_log.c
ldorau: I would remove the first part, but the second one should be required then:
```c
#else /* !UMF_VERSION */
#error UMF_VERSION not defined!
#endif /* !UMF_VERSION */
```

ldorau (changes requested):

RELEASE_STEPS.md
```diff
@@ -40,7 +40,7 @@ Do changes for a release:
 - Add an entry to ChangeLog, remember to change the day of the week in the release date
   - For major releases mention API and ABI compatibility with the previous release
 - Update project's version in a few places:
-  - Set the new $VERSION in `project` function in the top-level `CMakeLists.txt`
+  - For major and minor releases: `UMF_VERSION_CURRENT` in `include/umf/base.h` (the API version)
```
ldorau, line 43: What about `UMF_VERSION_CURRENT` ? Shouldn't it be equal to `UMF_VERSION` ?
I think that `umfGetCurrentVersion()` should return the new `UMF_VERSION`.

ldorau (dismissed):

src/utils/utils_log.c
lukaszstolarczuk: make sense, thx! Done.

RELEASE_STEPS.md
ldorau: I see. OK. Thanks for the explanation!

bratpiorka (approved):

lukaszstolarczuk:
I just added last few changes in the comments, plus removed the redundant messages - now we only print the UMF version (one and only 😉 )

ldorau (approved):

PR #618, opened 2024-07-17
Add disable_provider_free parameter to pool jemalloc

ldorau:
### Description

Add the `disable_provider_free ` parameter to pool jemalloc. It should be set to true if `umfMemoryProviderFree()` should never be called.

This option will be needed by the future memory providers that will not provide the free() op.

### Checklist
- [x] All tests pass locally

### DIFF
--- a/include/umf/pools/pool_jemalloc.h
+++ b/include/umf/pools/pool_jemalloc.h
@@ -14,7 +14,14 @@
 extern "C" {
 #endif
 
+#include <stdbool.h>
 #include <umf/memory_pool_ops.h>
+
+/// @brief Configuration of Jemalloc Pool
+typedef struct umf_jemalloc_pool_params_t {
+    /// Set to true if umfMemoryProviderFree() should never be called.
+    bool disable_provider_free;
+} umf_jemalloc_pool_params_t;
 
 umf_memory_pool_ops_t *umfJemallocPoolOps(void);
 
--- a/src/pool/pool_jemalloc.c
+++ b/src/pool/pool_jemalloc.c
@@ -37,6 +37,8 @@
 typedef struct jemalloc_memory_pool_t {
     umf_memory_provider_handle_t provider;
     unsigned int arena_index; // index of jemalloc arena
+    // set to true if umfMemoryProviderFree() should never be called
+    bool disable_provider_free;
 } jemalloc_memory_pool_t;
 
 static __TLS umf_result_t TLS_last_allocation_error;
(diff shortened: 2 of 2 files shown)


include/umf/pools/pool_jemalloc.h
```diff
@@ -14,8 +14,17 @@
...
 
+#include <stdbool.h>
 #include <umf/memory_pool_ops.h>
 
+/// @brief Configuration of Jemalloc Pool
+typedef struct umf_jemalloc_pool_params_t {
+    /// Set to true if umfMemoryProviderFree()
+    /// of the chosen provider always fails.
+    /// It is used to turn of the error messages.
+    bool provider_free_always_fails;
```
bratpiorka, line 25: If it is used to turn of the errors then I think "no_error_on_fail" or similar would be better. Additionally, if we know that the free will always fail, then we could add a flag "never_free" and don't call umfMemoryProviderFree at all.

include/umf/pools/pool_jemalloc.h
```diff
@@ -14,8 +14,17 @@
...
 #endif
 
+#include <stdbool.h>
 #include <umf/memory_pool_ops.h>
 
+/// @brief Configuration of Jemalloc Pool
+typedef struct umf_jemalloc_pool_params_t {
+    /// Set to true if umfMemoryProviderFree()
+    /// of the chosen provider always fails.
+    /// It is used to turn of the error messages.
```
bratpiorka, line 24: of -> off

include/umf/pools/pool_jemalloc.h
ldorau: Done

include/umf/pools/pool_jemalloc.h
ldorau: Done

bratpiorka (approved):

src/pool/pool_jemalloc.c
```diff
@@ -37,6 +37,8 @@
 typedef struct jemalloc_memory_pool_t {
     umf_memory_provider_handle_t provider;
     unsigned int arena_index; // index of jemalloc arena
+    // set to true if umfMemoryProviderFree() should never be called
+    bool disable_provider_free;
```
KFilipek, line 41: Can we replace this bool (similar to pool_jemalloc.h:23) with:
`umf_jemalloc_pool_params_t` params or sth?

lukaszstolarczuk (commented):
👍

PR #622, opened 2024-07-19
Add lock for updating file size in OS memory provider

ldorau:
### Description

Add lock for updating file size in OS memory provider, because `umfMemoryProviderAlloc()` has to be MT-safe.

### Checklist
- [x] All tests pass locally

### DIFF
--- a/src/provider/provider_os_memory.c
+++ b/src/provider/provider_os_memory.c
@@ -525,6 +525,14 @@ static umf_result_t os_initialize(void *params, void **provider) {
         goto err_destroy_bitmaps;
     }
 
+    if (os_provider->fd > 0) {
+        if (util_mutex_init(&os_provider->lock_fd) == NULL) {
+            LOG_ERR("initializing the file size lock failed");
+            ret = UMF_RESULT_ERROR_UNKNOWN;
+            goto err_destroy_bitmaps;
+        }
+    }
+
     os_provider->nodeset_str_buf = umf_ba_global_alloc(NODESET_STR_BUF_LEN);
     if (!os_provider->nodeset_str_buf) {
         LOG_INFO("allocating memory for printing NUMA nodes failed");
@@ -561,6 +569,10 @@ static void os_finalize(void *provider) {
     }
 
     os_memory_provider_t *os_provider = provider;
+
+    if (os_provider->fd > 0) {
+        util_mutex_destroy_not_free(&os_provider->lock_fd);
+    }
 
     critnib_delete(os_provider->fd_offset_map);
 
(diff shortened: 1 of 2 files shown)


src/provider/provider_os_memory.c
```diff
@@ -641,13 +653,20 @@ static int os_mmap_aligned(void *hint_addr, size_t length, size_t alignment,
...
+
         if (*fd_size + extended_length > max_fd_size) {
+            util_mutex_unlock(lock_fd);
             LOG_ERR("cannot grow a file size beyond %zu", max_fd_size);
             return -1;
         }
 
         fd_offset = *fd_size;
         *fd_size += extended_length;
+        util_mutex_unlock(lock_fd);
```
lplewa, line 669: What about something like this? It probably should have better performance, and we don't need extra locks. 
I'm not very proficient with atomics, so it would be good if @pbalcer or @igchor check if my proposal is valid . 
```suggestion
  do {
  size_t size = *fd_size;
  if (size + extended_length > max_fd_size) {
           LOG_ERR("cannot grow a file size beyond %zu", max_fd_size);
           return -1;
  }
  fd_offset = *fd_size;
  } while(!__sync_bool_compare_and_swap(fd_size, size, size + extended_lenght));
```

lplewa (changes requested):

src/provider/provider_os_memory.c
ldorau: It does not work - another thread can update `fd_size` with exactly the same value ...

lplewa: It should not be a problem - as compare_and_swap checks for old value, and them updates it to new value. So even if other tread updates this value, second thread will detect it

ldorau: Right, it should work. 
But how would you do that on Windows? I do not know any equivalent of `__sync_bool_compare_and_swap` on Windows ...

lplewa: All atomic functions are available on windows - just they have stupid names. 
https://learn.microsoft.com/en-us/windows/win32/api/winnt/nf-winnt-interlockedcompareexchange64

pbalcer: my 2 cents: since this isn't perf critical, I'd just leave it as is with the mutex. Atomics can be tricky to get right and make things more difficult to modify.

ldorau: @lplewa ?

bratpiorka: I agree with Piotr - since this is not the allocation path we should use the simplest solution

lplewa: > I agree with Piotr - since this is not the allocation path we should use the simplest solution

This is an allocation path

bratpiorka (approved):

PR #631, opened 2024-07-23
Add umfMemspaceMemtarget[Num|Get]

lplewa:


test/memspaces/memspace_numa.cpp
```diff
@@ -56,6 +57,10 @@ TEST_F(numaNodesTest, createDestroy) {
         nodeIds.data(), nodeIds.size(), &hMemspace);
     ASSERT_EQ(ret, UMF_RESULT_SUCCESS);
     ASSERT_NE(hMemspace, nullptr);
+    EXPECT_EQ(umfMemspaceMemtargetNum(hMemspace), nodeIds.size());
+    for (size_t i = 0; i < umfMemspaceMemtargetNum(hMemspace); ++i) {
+        EXPECT_NE(umfMemspaceMemtargetGet(hMemspace, i), nullptr);
+    }
```
lukaszstolarczuk, line 63: what about testing other memspaces...?

test/memspaces/memspace_numa.cpp
```diff
@@ -6,6 +6,7 @@
 #include "memspace_fixtures.hpp"
 #include "memspace_helpers.hpp"
 #include "memspace_internal.h"
+#include "umf/memspace.h"
```
bratpiorka, line 9: nit: #include <umf/memspace.h> and move <> includes up

test/memspaces/memspace_numa.cpp
bratpiorka: imo this is fine, as these functions are trivial and do not depend on the memspace type

lplewa: +1 it's not test for memspaces, but test for this function.

test/memspaces/memspace_numa.cpp
lplewa: Done

bratpiorka (approved):

test/memspaces/memspace_numa.cpp
lukaszstolarczuk: it would take you the same amount of time to add these few extra lines in the code, as you just wasted on this review 😄 
but being serious - since we don't have coverage check on each PR, we may change this test, or remove it, or something else and we won't even notice.

If you would have a seperate function or even a sep. file for this test then ok, I'd agree. But putting this into memspace test make it disappear.

lukaszstolarczuk (approved):

PR #656, opened 2024-07-31
[CMake] Don't use CMAKE_SOURCE_DIR during install

ProGTX:
Referring to an installed file using `CMAKE_SOURCE_DIR` only works when UMF is built and installed standalone, but not when it's included by other projects.

Only `LICENSE.TXT` and `licensing/third-party-programs.txt` still had these issue, while other files already used `PROJECT_SOURCE_DIR`.

- [x] All tests pass locally

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -599,10 +599,10 @@ endif()
 # --------------------------------------------------------------------------- #
 # Configure make install/uninstall and packages
 # --------------------------------------------------------------------------- #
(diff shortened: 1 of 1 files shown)


ProGTX:
Note this affects both https://github.com/oneapi-src/unified-runtime and https://github.com/intel/llvm .

pbalcer (approved):

vinser52 (approved):

bratpiorka (approved):

PR #667, opened 2024-08-08
Minor fixes around hwloc in CMake and CI

lukaszstolarczuk:
A small fixes in CMake and CI around hwloc options.

- [x] All tests pass locally

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,6 +52,10 @@ option(
     UMF_LINK_HWLOC_STATICALLY
     "Link UMF with HWLOC library statically (supported for Linux, MacOS and Release build on Windows)"
     OFF)
+option(
+    UMF_DISABLE_HWLOC
+    "Disable features that requires hwloc (OS provider, memory targets, topology discovery)"
+    OFF)
 option(UMF_FORMAT_CODE_STYLE
        "Add clang, cmake, and black -format-check and -format-apply targets"
        OFF)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -414,12 +414,8 @@ if(UMF_BUILD_BENCHMARKS)
...
 endif()
 
-if(UMF_BUILD_EXAMPLES)
-    if(NOT UMF_DISABLE_HWLOC)
-        add_subdirectory(examples)
-    else()
-        message(WARNING "Examples cannot be build - hwloc disabled")
-    endif()
+if(UMF_BUILD_EXAMPLES AND NOT UMF_DISABLE_HWLOC)
+    add_subdirectory(examples)
```
PatKamin, line 418: I think you could leave a STATUS message indicating that no examples will be built. This way, if someone sets UMF_BUILD_EXAMPLES and disables hwloc at the same time, they will understand why the examples weren't built.

PatKamin (approved):

CMakeLists.txt
```diff
@@ -100,16 +100,19 @@ else()
     message(FATAL_ERROR "Unknown OS type")
 endif()
 
-if(NOT DEFINED UMF_HWLOC_REPO)
-    set(UMF_HWLOC_REPO "https://github.com/open-mpi/hwloc.git")
-endif()
+if(UMF_DISABLE_HWLOC)
+    message(WARNING "hwloc is disabled, hence OS provider, memtargets, "
```
lukaszstolarczuk, line 104: hmm... I'm not sure if this should be a warning - this option is used in llvm project now, so they will see this warning all the time, right...?

PatKamin: I think that STATUS suffices just to notify the user. We shouldn't warn on using an option that is legitimate.

lukaszstolarczuk: yeah, but on the other hand it's disabling half of our functionality 😉

bratpiorka (approved):

CMakeLists.txt
lukaszstolarczuk: I've changed to STATUS

ldorau (approved):
Rebase please.

Maybe move `UMF_DISABLE_HWLOC` before ` UMF_LINK_HWLOC_STATICALLY` in CMake options?

CMakeLists.txt
```diff
@@ -52,6 +52,10 @@ option(
     UMF_LINK_HWLOC_STATICALLY
     "Link UMF with HWLOC library statically (supported for Linux, MacOS and Release build on Windows)"
     OFF)
+option(
```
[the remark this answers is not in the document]
lukaszstolarczuk: Done.

PatKamin (approved):

PR #672, opened 2024-08-09
[CI] Add Ubuntu 24.04 build

lukaszstolarczuk:
Add Ubuntu 24.04 build in basic workflow


PatKamin (approved):

lukaszstolarczuk:
> can we instead of adding additional build just repurpose one to be run on this OS. We have too many build and adding extra one do not help with it

Done.

ldorau (approved):

PR #677, opened 2024-08-13
docs: add details in CONTRIBUTING.md

EuphoricThinking:
Add black installation details in CONTRIBUTING.md
as in README.md

### Description

Added information about `black` installation in CONTRIBUTING.md, as it is in README.md. A possible contributor (who could focus on CONTRIBUTING.md) might have problems with formatting, since targets `format-apply` and `format-check` are not built if `black` is not installed (issue based on a true story).
### Checklist

- [ ] All tests pass locally

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -97,6 +97,7 @@ source files, respectively.
 **NOTE**: We use specific versions of formatting tools to ensure consistency across the project. The required versions are:
 - clang-format version **15.0**, which can be installed with the command: `python -m pip install clang-format==15.0.7`.
 - cmake-format version **0.6**, which can be installed with the command: `python -m pip install cmake-format==0.6.13`.
+- black (no specific version required), which can be installed with the command: `python -m pip install black`.
 
 Please ensure you have these specific versions installed before contributing to the project.
 


KFilipek (approved):

CONTRIBUTING.md
```diff
@@ -97,6 +97,7 @@ source files, respectively.
 **NOTE**: We use specific versions of formatting tools to ensure consistency across the project. The required versions are:
 - clang-format version **15.0**, which can be installed with the command: `python -m pip install clang-format==15.0.7`.
 - cmake-format version **0.6**, which can be installed with the command: `python -m pip install cmake-format==0.6.13`.
+- black version **24.3**, which can be installed with the command: `python -m pip install black==24.3.0`.
```
lukaszstolarczuk, line 100: hmm.. this section is about `specific required versions` of these tools. We don't require a specific version of `black` (it has rather good compatibility between versions (in comparison to clang-format, which may change too much between versions).

So... I would not add this info here...?


//
General information how to install black (including suggested version) is in the top-level Readme - https://github.com/oneapi-src/unified-memory-framework?tab=readme-ov-file#build

lukaszstolarczuk (changes requested):

CONTRIBUTING.md
EuphoricThinking: Do you wish to leave the suggested version number? Simply `python -m pip install black` should suffice.

I have removed the version number in the updated version.

EuphoricThinking: Or maybe you mean that you don't want to leave the info about black at all?

lukaszstolarczuk: I mean this change does not introduce anything new in here, I believe. The same info is in the top-level README.

This section, useful for contributors, says `The required versions are: ...` and `black` does not have any specific version required (at least at the moment).

lplewa (approved):

CONTRIBUTING.md
lukaszstolarczuk: ok, I see your point now. Then perhaps we could write something like this:

```
- black - no specific version required (can be installed with the command: `python -m pip install black==24.3.0)`.
```

EuphoricThinking: Done

lukaszstolarczuk (approved):

PR #678, opened 2024-08-14
Disable pciaccess library in hwloc

PatKamin:
### Checklist

- [ ] All tests pass locally

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -173,7 +173,8 @@ else()
             COMMAND
                 ./configure --prefix=${hwloc_targ_BINARY_DIR}
                 --enable-static=yes --enable-shared=no --disable-libxml2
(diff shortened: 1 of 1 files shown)


lukaszstolarczuk (approved):

lplewa:
can you provide why we are doing it in commit mesage?

igchor:
Similarly to https://github.com/oneapi-src/unified-memory-framework/pull/660, on some machines hwloc seems to link with some pci management library and since UMF does not, we get undefined symbol errors. (It would be good to check what exact library it is). Also, we should make sure that we don't actually require any pci-related features from hwloc. 

Additionally, since this is the second time, that hwloc links with something unexpected, perhaps we could add some CMake option to allow specyfing custom configure flags for hwloc to avoid similar problems in future?

bratpiorka (approved):

PR #684, opened 2024-08-15
Add option to control fetching L0

igchor:
Turning UMF_BUILD_LEVEL_ZERO_PROVIDER on in UR
results in configuration errors as both projects try to fetch and build L0 loader.

- [ ] All tests pass locally

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -4,12 +4,17 @@
 
 include(${UMF_CMAKE_SOURCE_DIR}/cmake/helpers.cmake)
 
+set(UMF_LEVEL_ZERO_INCLUDE_DIR
+    ""
+    CACHE FILEPATH "Directory containing the Level Zero Headers")
+
 # Compile definitions for UMF library.
 #
 # TODO: Cleanup the compile definitions across all the CMake files
 set(UMF_COMMON_COMPILE_DEFINITIONS UMF_VERSION=${UMF_VERSION})
 
-if(UMF_BUILD_LEVEL_ZERO_PROVIDER)
+if(UMF_BUILD_LEVEL_ZERO_PROVIDER AND (UMF_BUILD_GPU_TESTS
+                                      OR (NOT UMF_LEVEL_ZERO_INCLUDE_DIR)))
     include(FetchContent)
 
     set(LEVEL_ZERO_LOADER_REPO "https://github.com/oneapi-src/level-zero.git")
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -9,7 +9,7 @@ include(${UMF_CMAKE_SOURCE_DIR}/cmake/helpers.cmake)
 # TODO: Cleanup the compile definitions across all the CMake files
 set(UMF_COMMON_COMPILE_DEFINITIONS UMF_VERSION=${UMF_VERSION})
 
-if(UMF_BUILD_LEVEL_ZERO_PROVIDER)
```
lukaszstolarczuk, line 12: actually I can see we set here `LEVEL_ZERO_INCLUDE_DIRS` and UR is setting `LEVEL_ZERO_INCLUDE_DIR` which may result in error in UR (after we'd disable that new option).

Perhaps we could:
- rename `LEVEL_ZERO_INCLUDE_DIRS` in this repo to `LEVEL_ZERO_INCLUDE_DIR`
- check here (instead of this new option) if `LEVEL_ZERO_INCLUDE_DIR` is set
- set UMF target dependent of level_zero adapter, or more precisely of the `level_zero_loader` target (if that's possible)

igchor: We could just link with `LevelZeroLoader-Headers` target that is provided by UR (or create that target ourselves if UMF_FETCH_LEVEL_ZERO is set.

bratpiorka:
@igchor is this needed for 2025.0? AFAIK UR would not use L0 provider from UMF in 2025.0?

igchor:
> @igchor is this needed for 2025.0? AFAIK UR would not use L0 provider from UMF in 2025.0?

No, it's not for 2025.0. I'm working on L0 v2 adapter (https://github.com/oneapi-src/unified-runtime/tree/main/source/adapters/level_zero/v2) and I'd rather use UMF instead of duplicating the original logic for USM. We would still need to implement missing features (at least support for making the memory resident) but that shouldn't be too difficult.

src/CMakeLists.txt
bratpiorka: I agree that we should add an option that disables fetching L0 headers but for providing existing ones to UMF I would rather use CMAKE_C_FLAGS and not introduce any UR-specific code that makes our cmake files more complex

src/CMakeLists.txt
```diff
@@ -9,7 +9,7 @@ include(${UMF_CMAKE_SOURCE_DIR}/cmake/helpers.cmake)
...
+if(UMF_BUILD_LEVEL_ZERO_PROVIDER AND UMF_FETCH_LEVEL_ZERO)
```
bratpiorka, line 12: I think we should add the "else" path that checks if cmake include path contains L0 headers. 
Something like:

`
if (UMF_BUILD_LEVEL_ZERO_PROVIDER) 
  if (UMF_FETCH_LEVEL_ZERO) 
    // fetch
  else
    // check and print errors
`

igchor:
@bratpiorka  @lukaszstolarczuk  Ok, I did it slightly differently. Since we only need loader for tests (and example, but they have separate fetch_content) we can just add option to set `UMF_LEVEL_ZERO_INCLUDE_DIR`.  If this option is set and we don't need the loader, we can skip fetching L0. UR actually have a similar mechanism with `UR_LEVEL_ZERO_INCLUDE_DIR`

igchor:
@lukaszstolarczuk do you know why the GPU tests are failing on CI? it works for me locally

lukaszstolarczuk:
> @lukaszstolarczuk do you know why the GPU tests are failing on CI? it works for me locally

I think it's the issue with the machine, it should be fixed by now. I've hit re-run.

igchor:
@bratpiorka @lukaszstolarczuk are you OK with this solution? I've made a draft PR to UR to test this: https://github.com/oneapi-src/unified-runtime/pull/2012

bratpiorka (approved):

lukaszstolarczuk (approved):

PR #690, opened 2024-08-21
Refactor IPC tests and enable them for OS provider

vinser52:
### Description
Minor refactoring of IPC. Enabled IPC tests for the OS provider.

### Checklist
- [x] All tests pass locally

### DIFF
--- a/src/ipc.c
+++ b/src/ipc.c
@@ -51,6 +51,11 @@ umf_result_t umfPoolGetIPCHandleSize(umf_memory_pool_handle_t hPool,
 
 umf_result_t umfGetIPCHandle(const void *ptr, umf_ipc_handle_t *umfIPCHandle,
                              size_t *size) {
+    if (ptr == NULL || umfIPCHandle == NULL || size == NULL) {
+        LOG_ERR("invalid argument.");
+        return UMF_RESULT_ERROR_INVALID_ARGUMENT;
+    }
+
     size_t ipcHandleSize = 0;
     umf_alloc_info_t allocInfo;
     umf_result_t ret = umfMemoryTrackerGetAllocInfo(ptr, &allocInfo);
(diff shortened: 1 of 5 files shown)


lukaszstolarczuk:
> @bratpiorka As I understand, CI fails because the `mmap` is failed inside the `os_open_ipc_handle` function. According to the log, it failed not immediately but after some iterations. I guess it is because the IPC cache is not implemented yet. But locally tests are passed. Is there a way to change CI runner parameters to allow more virtual memory to be used by the process?

The tests that failed are running on the VMs delivered to us, not defined by us. If you can change it in the system - just update our CI workflow; if you can't change it we have to update the test, I guess.

Perhaps, we could also enable this specific test only in a selected workflow - e.g. multi-numa (which has our own machine used for CI).

vinser52:
> > @bratpiorka As I understand, CI fails because the `mmap` is failed inside the `os_open_ipc_handle` function. According to the log, it failed not immediately but after some iterations. I guess it is because the IPC cache is not implemented yet. But locally tests are passed. Is there a way to change CI runner parameters to allow more virtual memory to be used by the process?
> 
> The tests that failed are running on the VMs delivered to us, not defined by us. If you can change it in the system - just update our CI workflow; if you can't change it we have to update the test, I guess.
> 
> Perhaps, we could also enable this specific test only in a selected workflow - e.g. multi-numa (which has our own machine used for CI).

Ok, thank you, @lukaszstolarczuk. 
For now I enabled only OS provider with Disjoint pool combination. We can enable tests with a scalable pool once IPC caching is done.

src/ipc.c
```diff
@@ -51,6 +51,11 @@ umf_result_t umfPoolGetIPCHandleSize(umf_memory_pool_handle_t hPool,
 
 umf_result_t umfGetIPCHandle(const void *ptr, umf_ipc_handle_t *umfIPCHandle,
                              size_t *size) {
+    if (ptr == NULL || umfIPCHandle == NULL || size == NULL) {
+        LOG_ERR("invalid arguments.");
```
ldorau, line 55: `"invalid argument"`

ldorau (approved):

src/ipc.c
vinser52: Done

ldorau (approved):

lukaszstolarczuk (approved):

PR #692, opened 2024-08-21
Extend L0 provider to match UR capabilities

igchor:
Extend L0 provider with:
- ability to make memory resident on devices
- ability to allocate memory exceeding maxMemAllocSize

to match L0 provider capabilities from UR.

### Checklist

- [ ] All tests pass locally

### DIFF
--- a/include/umf/providers/provider_level_zero.h
+++ b/include/umf/providers/provider_level_zero.h
@@ -14,6 +14,9 @@
 extern "C" {
 #endif
 
+typedef struct _ze_context_handle_t *ze_context_handle_t;
+typedef struct _ze_device_handle_t *ze_device_handle_t;
+
 /// @brief USM memory allocation type
 typedef enum umf_usm_memory_type_t {
     UMF_MEMORY_TYPE_UNKNOWN = 0, ///< The memory pointed to is of unknown type
@@ -24,9 +27,17 @@ typedef enum umf_usm_memory_type_t {
 
 /// @brief Level Zero Memory Provider settings struct
 typedef struct level_zero_memory_provider_params_t {
-    void *level_zero_context_handle;   ///< Handle to the Level Zero context
-    void *level_zero_device_handle;    ///< Handle to the Level Zero device
+    ze_context_handle_t
+        level_zero_context_handle; ///< Handle to the Level Zero context
+    ze_device_handle_t
+        level_zero_device_handle; ///< Handle to the Level Zero device
+
     umf_usm_memory_type_t memory_type; ///< Allocation memory type
+
+    ze_device_handle_t *
+        resident_device_handles; ///< Array of devices for which the memory should be made resident
+    uint32_t
+        resident_device_count; ///< Number of devices for which the memory should be made resident
 } level_zero_memory_provider_params_t;
 
 umf_memory_provider_ops_t *umfLevelZeroMemoryProviderOps(void);
--- a/src/provider/provider_level_zero.c
+++ b/src/provider/provider_level_zero.c
@@ -63,6 +63,25 @@ typedef struct ze_ops_t {
 static ze_ops_t g_ze_ops;
 static UTIL_ONCE_FLAG ze_is_initialized = UTIL_ONCE_FLAG_INIT;
 static bool Init_ze_global_state_failed;
+static __TLS ze_result_t TLS_last_native_error;
+
+static void store_last_native_error(int32_t native_error) {
+    TLS_last_native_error = native_error;
+}
+
+umf_result_t ze2umf_result(ze_result_t result) {
+    switch (result) {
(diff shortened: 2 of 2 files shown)


include/umf/providers/provider_level_zero.h
```diff
@@ -14,6 +14,9 @@
 extern "C" {
 #endif
 
+typedef struct _ze_context_handle_t *ze_context_handle_t;
```
bratpiorka, line 17: I'm not we should create typedefs defined in ze_api.h - there would be no conflict if user include this file?

src/provider/provider_level_zero.c
```diff
@@ -168,8 +201,8 @@ static umf_result_t ze_memory_provider_alloc(void *provider, size_t size,
             .flags = 0};
         ze_device_mem_alloc_desc_t dev_desc = {
             .stype = ZE_STRUCTURE_TYPE_DEVICE_MEM_ALLOC_DESC,
-            .pNext = NULL,
-            .flags = 0,
+            .flags = NULL,
```
bratpiorka, line 204: please keep the stype, pNext, and flags initialization order. Also please set flags to 0 instead of NULL

src/provider/provider_level_zero.c
```diff
@@ -114,6 +130,13 @@ umf_result_t ze_memory_provider_initialize(void *params, void **provider) {
     ze_provider->device = ze_params->level_zero_device_handle;
     ze_provider->memory_type = (ze_memory_type_t)ze_params->memory_type;
 
+    ze_result_t ret = g_ze_ops.zeDeviceGetProperties(
+        ze_provider->device, &ze_provider->device_properties);
+    if (ret != ZE_RESULT_SUCCESS) {
+        umf_ba_global_free(ze_provider);
+        return UMF_RESULT_ERROR_MEMORY_PROVIDER_SPECIFIC;
+    }
+
```
bratpiorka, line 139: could you add checks for resident_device_count and resident_device_handles? I would expect that the number of not-NULL resident_device_handles should match the count

src/provider/provider_level_zero.c
```diff
@@ -181,6 +214,15 @@ static umf_result_t ze_memory_provider_alloc(void *provider, size_t size,
         return UMF_RESULT_ERROR_MEMORY_PROVIDER_SPECIFIC;
     }
 
+    for (uint32_t i = 0; i < ze_provider->resident_device_count; i++) {
```
bratpiorka, line 217: I'm not sure but is calling zeContextMakeMemoryResident valid also for HOST allocations?

igchor: Yes, UR makes the allocation resident on all devices in the context in such case. My idea here was that the user (UR) would pass an appropriate list of devices depending on the memory type.

src/provider/provider_level_zero.c
igchor: I'm not sure if this is needed. If we call zeContextMakeMemoryResident we will get INVALID argument anyway in that case. I think we should just create a mapping from L0 error codes to UMF error codes.

include/umf/providers/provider_level_zero.h
igchor: There will be no problem with include, this is already what's happening in provider_level_zero.c. But I wonder if this typdef can change in the future - would that be considered L0 API breaking change or would this be allowed?

igchor: @vinser52 any thoughts? should we just keep void* here?

vinser52: hmm, the only concern I have is what if this header and L0 header are used simultaneously in the user code.

As @igchor mentioned above there is already such situation in `provider_level_zero.c` source file. It would be good also to check with C++ compiler. Our code is plain C.

src/provider/provider_level_zero.c
igchor: Done, it was a rebase conflict.

src/provider/provider_level_zero.c
igchor: Done.

include/umf/providers/provider_level_zero.h
igchor: This is a regular forward declaration, there's nothing here that could cause problems. I checked with the C++ compiler as well and everything works fine.

igchor:
@bratpiorka @lukaszstolarczuk any more comments/concern? I already have a PR that uses L0 provider from UMF (in L0 v2 adapter) but with features from this PR we could actually enable all test scenarios.

Ref: https://github.com/oneapi-src/unified-runtime/pull/2016

bratpiorka (approved):

lukaszstolarczuk (approved):

PR #694, opened 2024-08-22
add umfMemspace[New|TargetRemove|TargetAdd]

lplewa:


include/umf/memspace.h
vinser52, line 1: Do we already have use cases for such API? Does OpenMP need it? I am just curious to understand how it is used in real life.

lplewa: There is restrict_storage//subnumspace one. In their example they want to allocate from every second target(a.k.a. every second memory). We can achieve this also with filters (which I'm going to implement next), but such basic API as create empty memspace and add or remove targets from it, is quite useful too in my opinion. i.e when you want to create pool which allocates from single specific numa node. 

Currently we have  "umfMemspaceCreateFromNumaArray" to cover this openMP case, but i will consider to remove it(or to not export it to be precise), when this and filtering patch is merged.

vinser52: > but i will consider to remove it

that is what I am worried about. The first release of UMF is coming soon. And we should care about backward compatibility.

So we should be careful with introducing new API - it is not regarding this particular PR, but rather a general rule we should follow. Also, we need to consider an option to mark some API as experimental - it will give our users a notion that experimental API is not stable and might be removed/changed in the future.

bratpiorka: @vinser52 this would not be merged on v0.9.x branch (2025.0 oneAPI release)

lplewa: > > but i will consider to remove it
> 
> that is what I am worried about. The first release of UMF is coming soon. And we should care about backward compatibility.
> 
> So we should be careful with introducing new API - it is not regarding this particular PR, but rather a general rule we should follow. Also, we need to consider an option to mark some API as experimental - it will give our users a notion that experimental API is not stable and might be removed/changed in the future.

As far i know we are releasing 0.9 because we do not want to commit to stable API ([semver specyfication](https://semver.org/#spec-item-4))  There is few minor issues with our API they i did not have time to fix. Like wrong variable size, const, signed  vs unsigned

include/umf/memspace.h
```diff
@@ -85,6 +85,12 @@ umf_const_memspace_handle_t umfMemspaceHighestBandwidthGet(void);
 ///
 umf_const_memspace_handle_t umfMemspaceLowestLatencyGet(void);
 
+/// \brief Creates new empty memspace.
```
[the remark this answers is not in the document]
lplewa: done

include/umf/memspace.h
```diff
@@ -100,6 +106,30 @@ umf_const_memtarget_handle_t
 umfMemspaceMemtargetGet(umf_const_memspace_handle_t hMemspace,
                         unsigned targetNum);
 
+/// \brief Adds memory target to memspace.
+///
+/// \details
+/// This function duplicates the memory target and then adds its to the memspace
+/// This means that original memtarget handle and the handle of the duplicated memtarget are different.
```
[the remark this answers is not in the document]
lplewa: done

src/memtarget_internal.h
```diff
@@ -38,6 +38,10 @@ umf_result_t umfMemtargetGetLatency(umf_memtarget_handle_t srcMemoryTarget,
                                     umf_memtarget_handle_t dstMemoryTarget,
                                     size_t *latency);
 
+/// return 0 if memtargets are equal, -1 if they are different type,
```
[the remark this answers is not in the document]
lplewa: Done

src/memtarget.c
```diff
@@ -1,6 +1,6 @@
 /*
  *
- * Copyright (C) 2023-2024 Intel Corporation
+ * Copyright (C) 2022-2024 Intel Corporation
```
bratpiorka, line 3: ?

bratpiorka (approved):

PatKamin (approved):

src/memtarget.c
lplewa: Done

bratpiorka:
@lplewa - is this ready for merge?

PR #695, opened 2024-08-22
Remove umfTearDown call on windows

omarahmed1111:
### Description

windows calls `atexit` before it calls object destructors, this would create seg-fault problems where it will call `atexit(umfTearDown)` which will destroy umf, but the libraries that use umf will start calling the destructors after to destruct umf objects and as `umfTearDown` was already called, this umf objects will begin seg-faulting with errors, and memory leaks also happen. 

For example: Unified-runtime library is using umf pools in l0 contexts, when the destructor for the context will start being called, it will start calling umf pool destruction `umfPoolDestroy` on umf pool objects. This will error as before it have called these destructors, `umfTearDown` would have already been called. 

We should give the control of calling `umfTearDown()` to the calling library. for example: calling `umfTearDown` with `urLoaderTearDown`/`urAdapterRelease`. 

### Checklist

- [ ] All tests pass locally

### DIFF
--- a/src/libumf_windows.c
+++ b/src/libumf_windows.c
@@ -40,7 +40,6 @@ BOOL CALLBACK initOnceCb(PINIT_ONCE InitOnce, PVOID Parameter,
     (void)lpContext; // unused
 
     int ret = umfInit();
-    atexit(umfTearDown);
     return (ret == 0) ? TRUE : FALSE;
 }
 


igchor:
If I remember correctly we didn't want to force users to call umfTearDown() in the simple case (where there are no global objects that use UMF) so ideally we would leave this atexit call as is.

Since UMF does refcounting in umfInit/umfTearDown I think it should be enough to just call umfInit urAdapterGet/urLoaderInit and teardown in urAdapterRelease/urLoaderTearDown as you said (which we are not doing at the moment I think?). This atexit should not cause any problems then - it would be just one additional refcount.

omarahmed1111:
> If I remember correctly we didn't want to force users to call umfTearDown() in the simple case (where there are no global objects that use UMF) so ideally we would leave this atexit call as is.
> 
> Since UMF does refcounting in umfInit/umfTearDown I think it should be enough to just call umfInit urAdapterGet/urLoaderInit and teardown in urAdapterRelease/urLoaderTearDown as you said (which we are not doing at the moment I think?). This atexit should not cause any problems then - it would be just one additional refcount.

Make sense, I didn't notice that it uses RefCounting, that would do it. Thanks for the info!

Closing this now as it should be resolved on UR side.

Issue #569, opened 2024-06-26
Inconsistency in Usage of Assertion Methods in Tests


lplewa:
### Current Situation
Our test suite currently uses two different assertion frameworks:
1. **Google Test (GTEST) Assertions**: These assertions use `ASSERT` and `EXPECT` macros.
2. **UT_ASSERTs**: These assertions call `abort()` on failure, which undesirably terminates the execution of any subsequent tests within the same file.
3. Some test constructors utilize only EXPECT macro, and after initialization fails, it continue test execution, causing it fail in different place. 

### Problems Identified
- **Test Execution Interruption**: The use of `UT_ASSERT` can prematurely end test execution upon failure, preventing full test suite runs.
- **Incompatibility in Test Initialization**: Google Test's `ASSERT` macros are not suitable for use in initialization routines (like in helper functions or within test class constructors) because they employ `return` statements that is supposed to exit test "main function"

### Proposed Solution
1. **Standardize on GTEST Assertions**: Transition all assertions to GTest.

lplewa:
Idea with a base class might be not the best approach - i noticed it might not be needed. It is to be considered during refactoring

bratpiorka:
closing as we merged the last part of fixes

PR #704, opened 2024-09-02
fix: substitute UT_ASSERTs with asserts from GTEST part 4

EuphoricThinking:
The PR is a part of a major fix, which will be provided in multiple files.

Ref. #569 

canQuery*(...) are changed to void. The success, failure or the need to skip the tests are determined in the function body using GTEST asserts, which set the appropriate flags for the current test. The flags are checked then in the fixture. The solution is inspired by [GTEST docs.](https://google.github.io/googletest/advanced.html#checking-for-failures-in-the-current-test).

The condition using ::memspaceGetTest::IsSkipped() in memspaceProviderTest is replaced due to its ineffectiveness: it has used the method from superclass for itself, before skipping any tests.

EXPECT_* and ADD_FAILURE() interefere with GTEST_SKIP() when used in a non-void helper function - tests are run in spite of labelling as ,,skipped" by the fixture.

### Checklist

- [ ] All tests pass locally

### DIFF
--- a/test/memspaces/memspace_fixtures.hpp
+++ b/test/memspaces/memspace_fixtures.hpp
@@ -51,7 +51,7 @@ struct numaNodesTest : ::umf_test::test {
     unsigned long maxNodeId = 0;
 };
 
-using isQuerySupportedFunc = bool (*)(size_t);
+using isQuerySupportedFunc = void (*)(size_t);
 using memspaceGetFunc = umf_const_memspace_handle_t (*)();
 using memspaceGetParams = std::tuple<isQuerySupportedFunc, memspaceGetFunc>;
 
@@ -65,9 +65,10 @@ struct memspaceGetTest : ::numaNodesTest,
         }
 
         auto [isQuerySupported, memspaceGet] = this->GetParam();
+        isQuerySupported(nodeIds.front());
 
-        if (!isQuerySupported(nodeIds.front())) {
-            GTEST_SKIP();
+        if (IS_SKIPPED_OR_FAILED()) {
+            return;
         }
 
         hMemspace = memspaceGet();
@@ -81,8 +82,18 @@ struct memspaceProviderTest : ::memspaceGetTest {
     void SetUp() override {
         ::memspaceGetTest::SetUp();
 
-        if (::memspaceGetTest::IsSkipped()) {
-            GTEST_SKIP();
+        if (numa_available() == -1 || numa_all_nodes_ptr == nullptr) {
+            GTEST_SKIP() << "No available NUMA support; skipped";
+        }
+
+        auto [isQuerySupported, memspaceGet] = ::memspaceGetTest::GetParam();
+        isQuerySupported(nodeIds.front());
+
+        // The test has been marked as skipped in isQuerySupported,
+        // repeating GTEST_SKIP in fixture would only duplicate
+        // the output message
+        if (IS_SKIPPED_OR_FAILED()) {
+            return;
         }
 
         umf_result_t ret =
--- a/test/memspaces/memspace_highest_bandwidth.cpp
+++ b/test/memspaces/memspace_highest_bandwidth.cpp
@@ -9,16 +9,17 @@
 #include "memspace_internal.h"
 #include "test_helpers.h"
 
-static bool canQueryBandwidth(size_t nodeId) {
+static void canQueryBandwidth(size_t nodeId) {
     hwloc_topology_t topology = nullptr;
     int ret = hwloc_topology_init(&topology);
-    UT_ASSERTeq(ret, 0);
+    ASSERT_EQ(ret, 0);
+
     ret = hwloc_topology_load(topology);
-    UT_ASSERTeq(ret, 0);
+    ASSERT_EQ(ret, 0);
 
     hwloc_obj_t numaNode =
         hwloc_get_obj_by_type(topology, HWLOC_OBJ_NUMANODE, nodeId);
-    UT_ASSERTne(numaNode, nullptr);
+    ASSERT_NE(numaNode, nullptr);
 
     // Setup initiator structure.
     struct hwloc_location initiator;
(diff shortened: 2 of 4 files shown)


test/memspaces/memspace_highest_bandwidth.cpp
```diff
@@ -12,13 +12,23 @@
 static bool canQueryBandwidth(size_t nodeId) {
     hwloc_topology_t topology = nullptr;
     int ret = hwloc_topology_init(&topology);
-    UT_ASSERTeq(ret, 0);
+
+    if (!GTEST_OUT_EQ(ret, 0)) {
```
PatKamin, line 16: Can the `EXPECT_EQ` macro be used instead? What is the advantage of using the `GTEST_OUT_EQ` macro over `EXPECT_EQ`?

EuphoricThinking: I have tried EXPECT_EQ() and it interferes with GTEST_SKIP(): although the tests are labeled as ,,skipped" in the fixture, they are executed. For example, failed ASSERT_NE(sth, nullptr) are outputted, because sth has been never initialized since the fixture stopped the SetUp() execution after GTEST_SKIP(). 

 ADD_FAILURE() gives similar result to EXPECT_EQ() (tests are eventually run).

Therefore I have decided to display the message and to manipulate the program execution only in the given fixture.

PatKamin: If we need these GTEST checks, then I'm ok with that. Please, consider if we could get rid of the GTEST macros in this function and ie. return a `std::pair<bool, std::string>` with a message to pass to GTEST_SKIP instead. No need for introducing new workaround macros if this could work.

test/memspaces/memspace_highest_bandwidth.cpp
```diff
@@ -9,16 +9,20 @@
 #include "memspace_internal.h"
 #include "test_helpers.h"
 
-static bool canQueryBandwidth(size_t nodeId) {
+static void canQueryBandwidth(size_t nodeId) {
     hwloc_topology_t topology = nullptr;
     int ret = hwloc_topology_init(&topology);
-    UT_ASSERTeq(ret, 0);
+
```
lukaszstolarczuk, line 15: are these extra empty lines needed? (added in all places in this PR)

test/memspaces/memspace_highest_bandwidth.cpp
```diff
@@ -30,7 +34,10 @@ static bool canQueryBandwidth(size_t nodeId) {
                                   numaNode, &initiator, 0, &value);
 
     hwloc_topology_destroy(topology);
-    return (ret == 0);
+
+    if (ret != 0) {
+        GTEST_SKIP() << "ret is equal to " << ret << ", should be " << 0;
```
lukaszstolarczuk, line 39: can we make these messages a little more meaningful? this `ret` is bound specifically to the result of `hwloc_memattr_get_value`

test/memspaces/memspace_fixtures.hpp
```diff
@@ -81,10 +82,17 @@ struct memspaceProviderTest : ::memspaceGetTest {
     void SetUp() override {
         ::memspaceGetTest::SetUp();
 
-        if (::memspaceGetTest::IsSkipped()) {
+        if (numa_available() == -1 || numa_all_nodes_ptr == nullptr) {
             GTEST_SKIP();
```
lukaszstolarczuk, line 86: can you add the message (like you did in the other place in this PR)

test/memspaces/memspace_helpers.hpp
```diff
@@ -16,6 +16,8 @@
 #define SIZE_4K (4096UL)
 #define SIZE_4M (SIZE_4K * 1024UL)
 
+#define IS_SKIPPED_OR_FAILED() (HasFatalFailure() || IsSkipped())
```
lukaszstolarczuk, line 19: hmm you can change this later (with other PR or something) - but is this helper function only needed in `memspace_helpers.hpp`? no other tests will use it...?

test/memspaces/memspace_fixtures.hpp
```diff
@@ -51,7 +51,7 @@ struct numaNodesTest : ::umf_test::test {
     unsigned long maxNodeId = 0;
 };
 
-using isQuerySupportedFunc = bool (*)(size_t);
+using isQuerySupportedFunc = void (*)(size_t);
```
lukaszstolarczuk, line 54: can you describe in a commit message, in a sentence or two, why this change from void to bool. This seem like the most important part of this PR (it changes very little of UT_ASSERTs, as the title would suggest 😉 )

test/memspaces/memspace_highest_bandwidth.cpp
EuphoricThinking: No, not needed. I have probably added them on "aesthetic" purpose, I'll just delete them.

test/memspaces/memspace_fixtures.hpp
EuphoricThinking: Something like the first paragraph of the PR description? Without the sentence with GTEST docs.

test/memspaces/memspace_helpers.hpp
lukaszstolarczuk: that was just a thought... perhaps `base.hpp` if you wantto put it in hpp header

test/memspaces/memspace_fixtures.hpp
```diff
@@ -65,9 +65,10 @@ struct memspaceGetTest : ::numaNodesTest,
         }
 
         auto [isQuerySupported, memspaceGet] = this->GetParam();
+        isQuerySupported(nodeIds.front());
 
-        if (!isQuerySupported(nodeIds.front())) {
-            GTEST_SKIP();
+        if (IS_SKIPPED_OR_FAILED()) {
+            return;
```
lukaszstolarczuk, line 71: hmm... so what happens when we return in the SetUp function? will the test skip or what?

EuphoricThinking: The test has been marked as skipped in isQuerySupported() using GTEST_SKIP(), repeating GTEST_SKIP() would only duplicate the skip message in the output. I'll add a comment in the code

lukaszstolarczuk (approved):

PatKamin (approved):

PR #708, opened 2024-09-05
Run setvars.sh also before building for icpx compiler

ldorau:
### Description

Run setvars.sh also before building for icpx compiler. It fixes the following warnings:
```
/usr/bin/ld: warning: libsvml.so, needed by ../lib/libumf.so.0.9.0, not found (try using -rpath or -rpath-link)
/usr/bin/ld: warning: libirng.so, needed by ../lib/libumf.so.0.9.0, not found (try using -rpath or -rpath-link)
/usr/bin/ld: warning: libimf.so, needed by ../lib/libumf.so.0.9.0, not found (try using -rpath or -rpath-link)
/usr/bin/ld: warning: libintlc.so.5, needed by ../lib/libumf.so.0.9.0, not found (try using -rpath or -rpath-link)
```

See the build: https://github.com/oneapi-src/unified-memory-framework/actions/runs/10700901541/job/29666095891#logs

### Checklist
- [x] All tests pass locally

### DIFF
--- a/.github/workflows/basic.yml
+++ b/.github/workflows/basic.yml
@@ -158,12 +158,14 @@ jobs:
         -DUMF_LINK_HWLOC_STATICALLY=${{matrix.link_hwloc_statically}}
 
     - name: Build UMF
-      run: cmake --build ${{env.BUILD_DIR}} -j $(nproc)
+      run: |
+        ${{ matrix.compiler.cxx == 'icpx' && '. /opt/intel/oneapi/setvars.sh' || true }}
+        cmake --build ${{env.BUILD_DIR}} -j $(nproc)
 
     - name: Run tests
       working-directory: ${{env.BUILD_DIR}}
-      run: >
-        ${{ matrix.compiler.cxx == 'icpx' && '. /opt/intel/oneapi/setvars.sh &&' || ''}} 
+      run: |
+        ${{ matrix.compiler.cxx == 'icpx' && '. /opt/intel/oneapi/setvars.sh' || true }}
         ctest --output-on-failure --test-dir test
 
     - name: Remove the installation directory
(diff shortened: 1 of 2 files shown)


bratpiorka (approved):

.github/workflows/basic.yml
```diff
@@ -158,7 +158,9 @@ jobs:
         -DUMF_LINK_HWLOC_STATICALLY=${{matrix.link_hwloc_statically}}
 
     - name: Build UMF
-      run: cmake --build ${{env.BUILD_DIR}} -j $(nproc)
+      run: >
```
lukaszstolarczuk, line 161: nit:
you can use `|` instead of `>` here ([docs](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsrun))

and then you can omit the second `&&` and ` || ''` in the line below.

It will give a little clarity, I think. Each line would be a separate command then.

ldorau: But there is the same syntax used below:
```
    - name: Run tests
      run: > 
        ${{ matrix.compiler.cxx == 'icpx' && '. /opt/intel/oneapi/setvars.sh &&' || ''}}
        ctest --output-on-failure
```

so I would use the same syntax just for consistency.

lukaszstolarczuk: heh, we could change both places 😉

kswiecicki (approved):

.github/workflows/basic.yml
ldorau: Done

KFilipek (approved):

.github/workflows/basic.yml
lukaszstolarczuk: thx!

PR #709, opened 2024-09-05
Bump L0 loader

lukaszstolarczuk:
Bump L0 loader version, to align with UR - it's the latest version. Also, update comments and cache variable type.

### Checklist
- [x] All tests pass locally

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -20,7 +20,7 @@ if(UMF_BUILD_LEVEL_ZERO_PROVIDER AND (UMF_BUILD_GPU_TESTS
     include(FetchContent)
 
     set(LEVEL_ZERO_LOADER_REPO "https://github.com/oneapi-src/level-zero.git")
-    set(LEVEL_ZERO_LOADER_TAG v1.16.1)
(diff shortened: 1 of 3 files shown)


src/CMakeLists.txt
```diff
@@ -6,13 +6,15 @@ include(${UMF_CMAKE_SOURCE_DIR}/cmake/helpers.cmake)
 
 set(UMF_LEVEL_ZERO_INCLUDE_DIR
     ""
-    CACHE FILEPATH "Directory containing the Level Zero Headers")
```
KFilipek, line 9: It's not a part of the comment's commit.

lukaszstolarczuk: this commit contains two small changes, both described in the commit msg. They are not relevant to the major change here, which is bumping the version.

KFilipek: I mean, the change from "FILEPATH" to "FILE" is not just an update to the comment, as the commit message says. It can be confusing if something goes wrong after the commit.
Regardless, let's leave it as is.

KFilipek (approved):

PR #710, opened 2024-09-06
Fix hwloc patch application on windows

npmiller:
### Description

Windows `cmd.exe` doesn't have a `true` command which causes the DPC++ build to fail, `(exit 0)` will do the same thing but it will work on both linux and windows.

This issue only shows up on Windows and not on a first build, since when building the first time the patch applies cleanly so the `||` branch is not taken, but then when rebuilding the patch doens't apply and the command errors out in the `||` branch.

### Checklist

- [X] Code compiles without errors locally
- [X] All tests pass locally

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -135,7 +135,7 @@ else()
             apply
             ${PROJECT_SOURCE_DIR}/cmake/fix_coverity_issues.patch
             ||
-            true)
+            (exit 0))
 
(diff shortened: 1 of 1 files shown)


bratpiorka (approved):

bratpiorka:
@npmiller thanks for finding and fixing this!

ldorau (approved):

PR #733, opened 2024-09-16
Fix two testNuma tests

ldorau:
### Description

Fix two testNuma tests:
- `checkModeInterleave` and
- `checkModeInterleaveCustomPartSize`

It fixes the Coverity issue no. 468463.

### Checklist
- [x] All tests pass locally

### DIFF
--- a/test/provider_os_memory_multiple_numa_nodes.cpp
+++ b/test/provider_os_memory_multiple_numa_nodes.cpp
@@ -373,8 +373,19 @@ TEST_F(testNuma, checkModeInterleave) {
 
     // Test where each page will be allocated.
     // Get the first numa node for ptr; Each next page is expected to be on next nodes.
+    int node = -1;
+    ASSERT_NO_FATAL_FAILURE(getNumaNodeByPtr(ptr, &node));
+    ASSERT_GE(node, 0);
     int index = -1;
-    ASSERT_NO_FATAL_FAILURE(getNumaNodeByPtr(ptr, &index));
+    for (size_t i = 0; i < numa_nodes.size(); i++) {
+        if (numa_nodes[i] == (unsigned)node) {
+            index = i;
+            break;
+        }
+    }
+    ASSERT_GE(index, 0);
+    ASSERT_LT(index, numa_nodes.size());
+
     for (size_t i = 1; i < (size_t)pages_num; i++) {
         index = (index + 1) % numa_nodes.size();
         EXPECT_NODE_EQ((char *)ptr + page_size * i, numa_nodes[index]);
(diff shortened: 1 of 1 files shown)


test/provider_os_memory_multiple_numa_nodes.cpp
```diff
@@ -421,6 +421,8 @@ TEST_F(testNuma, checkModeInterleaveCustomPartSize) {
     ASSERT_NO_FATAL_FAILURE(getNumaNodeByPtr(ptr, &index));
     for (size_t i = 0; i < (size_t)part_num; i++) {
         for (size_t j = 0; j < part_size; j += page_size) {
+            ASSERT_GE(index, 0);
+            ASSERT_LT(index, numa_nodes.size());
```
PatKamin, line 425: This check seems to be wrong in some cases. Let's say that we have available `numa_nodes` vector looking like this: `0, 1, 4`. If memory is allocated on the node `#4`, then this check would fail the test, because the `numa_nodes` size is 3.

ldorau: But there is a line:
```c
index = (index + 1) % numa_nodes.size();
```
below inside this `for` loop, so it has to be true that: `index < numa_nodes.size();`

PatKamin: Hmm, looks like an issue to be resolved in some other PR then... I think the test should cover cases that I mentioned. We can merge this one to satisfy Coverity, though.

PatKamin (approved):

test/provider_os_memory_multiple_numa_nodes.cpp
ldorau: I do not get it. If `vector = [4]` then `index == 0 and vector[0] == 4` - am I wrong?

ldorau:
@lplewa please review

test/provider_os_memory_multiple_numa_nodes.cpp
PatKamin: `getNumaNodeByPtr` returns the actual node number where the memory has been allocated, so it will return in this case `index=4`.

ldorau: If  `index` in `numa_nodes[index]` can be greater than `numa_nodes.size()` there is a real bug in this test.
@lplewa could you fix that?

test/provider_os_memory_multiple_numa_nodes.cpp
```diff
@@ -421,6 +421,8 @@ TEST_F(testNuma, checkModeInterleaveCustomPartSize) {
     ASSERT_NO_FATAL_FAILURE(getNumaNodeByPtr(ptr, &index));
     for (size_t i = 0; i < (size_t)part_num; i++) {
```
lplewa, line 422:
```suggestion
    ASSERT_NO_FATAL_FAILURE(getNumaNodeByPtr(ptr, &index));
    ASSERT_LT(index, numa_nodes.size());
    for (size_t i = 0; i < (size_t)part_num; i++) {
```

test/provider_os_memory_multiple_numa_nodes.cpp
ldorau: Fixed.

Done

PR #744, opened 2024-09-19
fix: substitute UT_ASSERTs with asserts from GTEST part 5

EuphoricThinking:
The PR is a part of a major fix, which will be provided in multiple files.

Ref. #569 

### Checklist

- [ ] All tests pass locally

### DIFF
--- a/test/provider_os_memory_multiple_numa_nodes.cpp
+++ b/test/provider_os_memory_multiple_numa_nodes.cpp
@@ -42,7 +42,14 @@ std::vector<int> get_available_cpus() {
     CPU_ZERO(mask);
 
     int ret = sched_getaffinity(0, sizeof(cpu_set_t), mask);
-    UT_ASSERTeq(ret, 0);
+
+    if (ret != 0) {
+        available_cpus.emplace_back(-1);
+        CPU_FREE(mask);
+
+        return available_cpus;
+    }
+
     // Get all available cpus.
     printf("All CPUs: ");
     for (size_t i = 0; i < CPU_SETSIZE; ++i) {
@@ -88,13 +95,16 @@ struct testNuma : testing::Test {
         ASSERT_NE(os_memory_provider, nullptr);
     }
 
-    struct bitmask *retrieve_nodemask(void *addr) {
-        struct bitmask *retrieved_nodemask = numa_allocate_nodemask();
-        UT_ASSERTne(nodemask, nullptr);
-        int ret = get_mempolicy(nullptr, retrieved_nodemask->maskp,
+    void retrieve_nodemask(void *addr, bitmask **retrieved_nodemask) {
+        *retrieved_nodemask = numa_allocate_nodemask();
+
+        ASSERT_NE(nodemask, nullptr);
+        ASSERT_NE(*retrieved_nodemask, nullptr);
+
+        int ret = get_mempolicy(nullptr, (*retrieved_nodemask)->maskp,
                                 nodemask->size, addr, MPOL_F_ADDR);
-        UT_ASSERTeq(ret, 0);
-        return retrieved_nodemask;
+
+        ASSERT_EQ(ret, 0);
     }
 
     void TearDown() override {
@@ -241,7 +251,17 @@ TEST_P(testNumaOnEachNode, checkModeInterleaveSingleNode) {
     EXPECT_NODE_EQ(ptr, numa_node_number);
 }
 
-struct testNumaOnEachCpu : testNuma, testing::WithParamInterface<int> {};
+struct testNumaOnEachCpu : testNuma, testing::WithParamInterface<int> {
+    void SetUp() override {
+        ::testNuma::SetUp();
+
+        int cpuNumber = this->GetParam();
+
+        if (cpuNumber < 0) {
+            GTEST_FAIL() << "get_available_cpus() error";
+        }
+    }
+};
 
 INSTANTIATE_TEST_SUITE_P(testNumaNodesAllocationsAllCpus, testNumaOnEachCpu,
                          ::testing::ValuesIn(get_available_cpus()));
@@ -260,7 +280,7 @@ TEST_P(testNumaOnEachCpu, checkModePreferredEmptyNodeset) {
     int ret = sched_setaffinity(0, sizeof(cpu_set_t), mask);
     CPU_FREE(mask);
 
-    UT_ASSERTeq(ret, 0);
+    ASSERT_EQ(ret, 0);
 
     umf_os_memory_provider_params_t os_memory_provider_params =
         UMF_OS_MEMORY_PROVIDER_PARAMS_TEST;
@@ -275,7 +295,7 @@ TEST_P(testNumaOnEachCpu, checkModePreferredEmptyNodeset) {
 
     // Verify we're on the expected CPU
     int cpu_check = sched_getcpu();
-    UT_ASSERTeq(cpu, cpu_check);
+    ASSERT_EQ(cpu, cpu_check);
 
     int numa_node_number = numa_node_of_cpu(cpu);
     printf("Got CPU: %d, got numa node: %d\n", cpu, numa_node_number);
@@ -297,7 +317,7 @@ TEST_P(testNumaOnEachCpu, checkModeLocal) {
     int ret = sched_setaffinity(0, sizeof(cpu_set_t), mask);
     CPU_FREE(mask);
 
-    UT_ASSERTeq(ret, 0);
+    ASSERT_EQ(ret, 0);
 
     umf_os_memory_provider_params_t os_memory_provider_params =
         UMF_OS_MEMORY_PROVIDER_PARAMS_TEST;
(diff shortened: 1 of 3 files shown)


.github/workflows/basic.yml
```diff
@@ -8,7 +8,7 @@ permissions:
 
 env:
   # for installation testing - it should match with version set in CMake
-  UMF_VERSION: 0.9.0
+  UMF_VERSION: 0.10.0
```
lukaszstolarczuk, line 11: please rebase, Lukasz's commit should not belong here.

test/provider_os_memory_multiple_numa_nodes.cpp
```diff
@@ -42,7 +42,14 @@ std::vector<int> get_available_cpus() {
     CPU_ZERO(mask);
 
     int ret = sched_getaffinity(0, sizeof(cpu_set_t), mask);
-    UT_ASSERTeq(ret, 0);
+
```
lukaszstolarczuk, line 45: pls describe in commit message what's really changed here - it seems like changing UT_ASSERTs is again, just a small addition 😉

EuphoricThinking: done?

test/provider_os_memory_multiple_numa_nodes.cpp
```diff
@@ -42,7 +42,14 @@ std::vector<int> get_available_cpus() {
...
+    if (ret != 0) {
+        available_cpus.emplace_back(-__LINE__);
```
lukaszstolarczuk, line 47: why is `__LINE__` used (here and some other places)?

test/provider_os_memory_multiple_numa_nodes.cpp
lukaszstolarczuk: heh, now I'm reading the commit message and see that you do 2 changes (change functions and fix segfault) in one commit - usually we tend to do 1 change per commit - e.g. in case you need to revert one of this changes for some reason;

please split this into 2 commits

test/provider_os_memory_multiple_numa_nodes.cpp
EuphoricThinking: The line number is saved for debugging purposes.

Since get_available_cpus() returns vector<int>, but it does not emplace back negative values, a negative value as a vector element indicates an error.

In retrieve_nodemask, a non-zero value as the first element of the returned pair indicates an error. This non-zero value is an approximate line where the error has occurred.

Should I add these explanations as comments?
Should I change __LINE__ to other values, such as 0, -1 etc.? This is just an approximate line number (since the condition is checked a few lines after the possible error)

test/provider_os_memory_multiple_numa_nodes.cpp
EuphoricThinking: Done?

test/provider_os_memory_multiple_numa_nodes.cpp
EuphoricThinking: Changed to -1

test/provider_os_memory_multiple_numa_nodes.cpp
lukaszstolarczuk: nope, there's still a little confusion in the commits:
- they both edit the same place (one commit introduced some change and the other one overrides them with some fixes),
- the title and description should describe what's inside of the commit - more or less it should be one thing

that said, I believe we can leave this for dessert; When all changes are done you can split the commits (or, if that's easier, you can start from scratch).

test/provider_os_memory_multiple_numa_nodes.cpp
```diff
@@ -88,13 +95,21 @@ struct testNuma : testing::Test {
         ASSERT_NE(os_memory_provider, nullptr);
     }
 
-    struct bitmask *retrieve_nodemask(void *addr) {
+    std::pair<int, bitmask *> retrieve_nodemask(void *addr) {
```
lukaszstolarczuk, line 98: why return via `pair`? can't we return bitmask via argument and just return `int` here...?

test/provider_os_memory_multiple_numa_nodes.cpp
```diff
@@ -241,7 +256,18 @@ TEST_P(testNumaOnEachNode, checkModeInterleaveSingleNode) {
...
 
-struct testNumaOnEachCpu : testNuma, testing::WithParamInterface<int> {};
+struct testNumaOnEachCpu : testNuma, testing::WithParamInterface<int> {
+    void SetUp() override {
+        ::testNuma::SetUp();
+
+        int cpuNumber = this->GetParam();
+
+        if (cpuNumber < 0) {
+            GTEST_FAIL() << "get_available_cpus(): sched_getaffinity() return "
```
lukaszstolarczuk, line 266: I think this message is too precise - what if we add another ASSERT within `get_available_cpus`? I'd say that we ca write something super generic here, like `something wrong happened when getting CPU count`, and that's ok, because we can always debug the issue and get to the root cause's line (which, in this case, is `sched_getaffinity`)

test/provider_os_memory_multiple_numa_nodes.cpp
```diff
@@ -88,13 +95,21 @@ struct testNuma : testing::Test {
...
+
+        if (nodemask == nullptr) {
+            return std::make_pair(-1, nodemask);
+        }
+
         int ret = get_mempolicy(nullptr, retrieved_nodemask->maskp,
                                 nodemask->size, addr, MPOL_F_ADDR);
-        UT_ASSERTeq(ret, 0);
-        return retrieved_nodemask;
+
```
lukaszstolarczuk, line 107: why can't we use asserts here and just check it like in your other PRs (`IS_SKIPPED_OR_FAILED`)?

test/provider_os_memory_multiple_numa_nodes.cpp
```diff
@@ -391,7 +417,16 @@ TEST_F(testNuma, checkModeInterleave) {
         EXPECT_NODE_EQ((char *)ptr + page_size * i, numa_nodes[index]);
     }
 
-    bitmask *retrieved_nodemask = retrieve_nodemask(ptr);
+    auto [fileLine, retrieved_nodemask] = retrieve_nodemask(ptr);
```
lukaszstolarczuk, line 420: `fileLine`?

test/provider_os_memory_multiple_numa_nodes.cpp
```diff
@@ -391,7 +417,16 @@ TEST_F(testNuma, checkModeInterleave) {
...
+    if (fileLine != 0) {
+        if (retrieved_nodemask == nullptr) {
+            GTEST_FAIL() << "retrieve_nodemask(): numa_allocate_nodemask() "
+                            "returned nullptr";
+        } else {
+            GTEST_FAIL() << "retrieve_nodemask(): get_mempolicy() return value "
```
lukaszstolarczuk, line 426: again, I think this is overcomplicated - what if we add another place, where `retrieve_nodemask` could fail? some generic message should be okay, without splitting it into two cases.

test/provider_os_memory_multiple_numa_nodes.cpp
```diff
@@ -391,7 +417,16 @@ TEST_F(testNuma, checkModeInterleave) {
         EXPECT_NODE_EQ((char *)ptr + page_size * i, numa_nodes[index]);
     }
 
-    bitmask *retrieved_nodemask = retrieve_nodemask(ptr);
+    auto [fileLine, retrieved_nodemask] = retrieve_nodemask(ptr);
+    if (fileLine != 0) {
```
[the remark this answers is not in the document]
lukaszstolarczuk: haha, samesies 😄 

image

test/provider_os_memory_multiple_numa_nodes.cpp
EuphoricThinking: Done? 🥺

test/provider_os_memory_multiple_numa_nodes.cpp
EuphoricThinking: Changed as you wished, the output message is more precise with your suggestion

test/provider_os_memory_multiple_numa_nodes.cpp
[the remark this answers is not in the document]
EuphoricThinking: Changed

test/provider_os_memory_multiple_numa_nodes.cpp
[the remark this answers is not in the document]
EuphoricThinking: IS_SKIPPED_OR_FAILED works, therefore I have chosen this option for more precise diagnostic messages

EuphoricThinking: The helper function is changed to void with IS_SKIPPED_OR_FAILED() in the main test

test/provider_os_memory_multiple_numa_nodes.cpp
EuphoricThinking: Changed the message

test/provider_os_memory_multiple_numa_nodes.cpp
EuphoricThinking: Changed the text

lukaszstolarczuk (approved):

lplewa (approved):

PR #756, opened 2024-09-25
Fix test_valgrind.sh

ldorau:
### Description

Fix `test_valgrind.sh`:
If a test fails, but there is no "ERROR SUMMARY" string in the LOG, the LOG is removed and the whole test fails with the following error message:
```
ls: cannot access 'umf_test-*.log': No such file or directory
```
This patch fixes that.

### Checklist
- [x] All tests pass locally

### DIFF
--- a/test/test_valgrind.sh
+++ b/test/test_valgrind.sh
@@ -68,7 +68,7 @@ echo
 echo "Working directory: $(pwd)"
 echo "Running: \"valgrind $OPTION\" for the following tests:"
 
-FAIL=0
+ANY_TEST_FAILED=0
 rm -f umf_test-*.log umf_test-*.err
 
 for test in $(ls -1 umf_test-*); do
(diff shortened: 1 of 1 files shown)


test/test_valgrind.sh
```diff
@@ -121,7 +121,10 @@ for test in $(ls -1 umf_test-*); do
 
 	[ "$FILTER" != "" ] && echo -n "($FILTER) "
 
+	LAST_FAILED=0
+
 	if ! HWLOC_CPUID_PATH=./cpuid valgrind $OPTION $OPT_SUP --gen-suppressions=all ./$test $FILTER >$LOG 2>&1; then
+		LAST_FAILED=1
 		FAIL=1
```
lukaszstolarczuk, line 128: perhaps we should be more descriptive with the `FAIL` variable name now - perhaps `ANY_TEST_FAILED` or something...?

lplewa (approved):

test/test_valgrind.sh
ldorau: Done:
- `FAIL` replaced with `ANY_TEST_FAILED` and
- `LAST_FAILED` replaced with `LAST_TEST_FAILED`

lukaszstolarczuk (approved):

PR #760, opened 2024-09-26
Fix arguments checking in IPC API of OS and file memory providers

ldorau:
### Description

IPC API of OS memory provider requires the `UMF_MEM_MAP_SHARED` memory `visibility` mode (`UMF_RESULT_ERROR_INVALID_ARGUMENT` is returned otherwise).

IPC API of file memory provider requires the `UMF_MEM_MAP_SHARED`
or `UMF_MEM_MAP_SYNC` memory `visibility` mode
(`UMF_RESULT_ERROR_INVALID_ARGUMENT` is returned otherwise).

### Checklist
- [x] All tests pass locally

### DIFF
--- a/src/provider/provider_file_memory.c
+++ b/src/provider/provider_file_memory.c
@@ -8,6 +8,7 @@
 #include <assert.h>
 #include <errno.h>
 #include <limits.h>
+#include <stdbool.h>
 #include <stddef.h>
 #include <stdio.h>
 #include <stdlib.h>
@@ -40,6 +41,9 @@ typedef struct file_memory_provider_t {
     unsigned protection; // combination of OS-specific protection flags
     unsigned visibility; // memory visibility mode
     size_t page_size;    // minimum page size
+
+    // IPC is enabled only for UMF_MEM_MAP_SHARED or UMF_MEM_MAP_SYNC visibility
+    bool IPC_enabled;
 
     critnib *mmaps; // a critnib map storing mmap mappings (addr, size)
 
@@ -100,6 +104,10 @@ file_translate_params(umf_file_memory_provider_params_t *in_params,
         LOG_ERR("incorrect memory visibility flag: %u", in_params->visibility);
         return result;
     }
+
+    // IPC is enabled only for UMF_MEM_MAP_SHARED or UMF_MEM_MAP_SYNC visibility
+    provider->IPC_enabled = (in_params->visibility == UMF_MEM_MAP_SHARED ||
+                             in_params->visibility == UMF_MEM_MAP_SYNC);
 
     return UMF_RESULT_SUCCESS;
 }
@@ -545,6 +553,13 @@ static umf_result_t file_get_ipc_handle_size(void *provider, size_t *size) {
         return UMF_RESULT_ERROR_INVALID_ARGUMENT;
     }
 
+    file_memory_provider_t *file_provider = (file_memory_provider_t *)provider;
+    if (!file_provider->IPC_enabled) {
+        LOG_ERR("memory visibility mode is not UMF_MEM_MAP_SHARED nor "
+                "UMF_MEM_MAP_SYNC")
+        return UMF_RESULT_ERROR_INVALID_ARGUMENT;
+    }
+
     *size = sizeof(file_ipc_data_t);
 
     return UMF_RESULT_SUCCESS;
@@ -557,6 +572,11 @@ static umf_result_t file_get_ipc_handle(void *provider, const void *ptr,
     }
 
     file_memory_provider_t *file_provider = (file_memory_provider_t *)provider;
+    if (!file_provider->IPC_enabled) {
+        LOG_ERR("memory visibility mode is not UMF_MEM_MAP_SHARED nor "
+                "UMF_MEM_MAP_SYNC")
+        return UMF_RESULT_ERROR_INVALID_ARGUMENT;
+    }
 
     void *value = critnib_get(file_provider->fd_offset_map, (uintptr_t)ptr);
     if (value == NULL) {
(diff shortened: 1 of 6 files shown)


PatKamin (approved):

src/provider/provider_file_memory.c
```diff
@@ -41,6 +42,9 @@ typedef struct file_memory_provider_t {
     unsigned visibility; // memory visibility mode
     size_t page_size;    // minimum page size
 
+    /// IPC is enabled only for UMF_MEM_MAP_SHARED or UMF_MEM_MAP_SYNC visibility
```
lukaszstolarczuk, line 45: why using `///` instead of `//` (this is not public API, it will not show up on the web docs)

src/provider/provider_file_memory.c
```diff
@@ -101,6 +105,10 @@ file_translate_params(umf_file_memory_provider_params_t *in_params,
         return result;
     }
 
+    /// IPC is enabled only for UMF_MEM_MAP_SHARED or UMF_MEM_MAP_SYNC visibility
```
lukaszstolarczuk, line 108: I guess this comment is redundant - it doubles the simple information from the code below, I believe

src/provider/provider_file_memory.c
```diff
@@ -545,6 +553,14 @@ static umf_result_t file_get_ipc_handle_size(void *provider, size_t *size) {
         return UMF_RESULT_ERROR_INVALID_ARGUMENT;
     }
 
+    file_memory_provider_t *file_provider = (file_memory_provider_t *)provider;
+    if (!file_provider->IPC_enabled) {
+        /// IPC is enabled only for UMF_MEM_MAP_SHARED or UMF_MEM_MAP_SYNC visibility
```
lukaszstolarczuk, line 558: I guess these comments are redundant, they say pretty much the same as the LOG message (applies to all these LOGs)

src/provider/provider_os_memory_internal.h
```diff
@@ -29,8 +30,13 @@ extern "C" {
 typedef struct os_memory_provider_t {
     unsigned protection; // combination of OS-specific protection flags
     unsigned visibility; // memory visibility mode
+
+    /// IPC is enabled only if (in_params->visibility == UMF_MEM_MAP_SHARED)
```
lukaszstolarczuk, line 34: just `//`

src/provider/provider_file_memory.c
```diff
@@ -545,6 +553,14 @@ static umf_result_t file_get_ipc_handle_size(void *provider, size_t *size) {
         return UMF_RESULT_ERROR_INVALID_ARGUMENT;
     }
 
+    file_memory_provider_t *file_provider = (file_memory_provider_t *)provider;
+    if (!file_provider->IPC_enabled) {
```
vinser52, line 557: I am not sure that the `file_get_ipc_handle_size()` function should fail if the `IPC_enabled` flag is `false`.
I am thinking about the scenario when on the consumer side, the user creates a file provider without required flags (`UMF_MEM_MAP_SHARED` and `UMF_MEM_MAP_SYNC`) and just wants to use this provider to open IPC handles and wants to know expected IPC handle size (MPI uses our API in that way).

vinser52: More general question: do we want to allow to open IPC handles by the file provider instance that does not have the abovementioned flags set?

src/provider/provider_file_memory.c
ldorau: Done

src/provider/provider_os_memory_internal.h
ldorau: Done

src/provider/provider_file_memory.c
ldorau: Done

src/provider/provider_file_memory.c
ldorau: Just a mistake.
Done.

src/provider/provider_file_memory.c
ldorau: The consumer also has to have the correct visibility mode set, otherwise it does not work:
https://github.com/ldorau/unified-memory-framework/commit/7cd29a75c0dcb6f4e139a5250a1e0d28c5a28924
https://github.com/ldorau/unified-memory-framework/actions/runs/11056477461/job/30718352219

ldorau: > I am not sure that the `file_get_ipc_handle_size()` function should fail if the `IPC_enabled` flag is `false`.
> I am thinking about the scenario when on the consumer side, the user creates a file provider without required flags (`UMF_MEM_MAP_SHARED` and `UMF_MEM_MAP_SYNC`) and just wants to use this provider to open IPC handles and wants to know expected IPC handle size (MPI uses our API in that way).

When the consumer opens the IPC handle with the MAP_PRIVATE flag, it will not work as you can see in the example above.

vinser52: > When the consumer opens the IPC handle with the `MAP_PRIVATE` flag, it will not work:
> [ldorau@7cd29a7](https://github.com/ldorau/unified-memory-framework/commit/7cd29a75c0dcb6f4e139a5250a1e0d28c5a28924)
> https://github.com/ldorau/unified-memory-framework/actions/runs/11056477461/job/30718352219

That is exactly my question: do we want such behavior? Why cannot we open the handle in such a case?

ldorau: This build was run on the current main branch, without this patch.
We do open the handle and it does not work, because it was mapped with `MAP_PRIVATE`.

ldorau:
```
[consumer] Received the IPC handle from the producer (4112 bytes)
[PID:[165](https://github.com/ldorau/unified-memory-framework/actions/runs/11056477461/job/30718352219#step:12:166)72  TID:16572  DEBUG UMF] utils_mmap_file: file mapped with the MAP_PRIVATE flag
[consumer] Opened the IPC handle received from the producer
[consumer] Read the number from the producer's shared memory: 140722523604400
[consumer] Wrote a new number directly to the producer's shared memory: 70361261802200
[consumer] Sent a response message to the producer
[producer] Received the consumer's response: "This is the consumer. I just wrote a new number directly into your shared memory!"
[producer] ERROR: The consumer did NOT write the correct value (the old one / 2 = 70361261802200) to my shared memory: 140722523604400
[consumer] Closed the IPC handle received from the producer
[producer] Put the IPC handle
[producer] Shutting down (status ERROR) ...
[consumer] Shutting down (status OK) ...
```

ldorau: So the scenario, you want to use, does not work. The consumer also has to map with `MAP_SHARED`.

vinser52: I think I got the point: the `UMF_MEM_MAP_SHARED` and `UMF_MEM_MAP_SYNC` flags are used not only in the provider's alloc function implementation but also in the "open IPC Handle" implementation. In both cases, mmap should be called with these flags.

lukaszstolarczuk (approved):

Issue #757, opened 2024-09-25
Unescaped linker flags when building with the Intel C++ Compiler on Windows and CMake > 3.25


Maetveis:
## Environment Information

- UMF version (hash commit or a tag):                         27d2ce8e12fcf20580f8b82d0afcbf4433da1c9d
- OS(es) version(s):                                          Windows Server 2022 (10.0.20348 Build 20348)
- compiler, libraries, and other related tools version(s):    CMake 3.28.2 (versions >= 3.25 should be affected)

## Please provide a reproduction of the bug:

```powershell
cmake.exe -DCMAKE_CXX_COMPILER=icx -DCMAKE_C_COMPILER=icx -S . -B build -DCMAKE_EXE_LINKER_FLAGS_INIT="-Werror=unknown-argument" -DCMAKE_SHARED_LINKER_FLAGS_INIT="-Werror=unknown-argument" -DCMAKE_MODULE_LINKER_FLAGS_INIT="-Werror=unknown-argument" -DCMAKE_STATIC_LINKER_FLAGS_INIT="-Werror=unknown-argument" -DCMAKE_BUILD_TYPE=Debug -G Ninja -DUMF_DISABLE_HWLOC=ON
cmake --build build
```

## How often bug is revealed:

always

## Actual behavior:

Building fails with the following errors.

```plaintext
[57/65] Linking CXX executable test\umf_test-memoryProvider.exe
FAILED: test/umf_test-memoryProvider.exe
[...]

PR #772, opened 2024-10-02
Wrap linker flags on Windows for IntelLLVM

Maetveis:
The Intel C++ compiler requires linker flags to be wrapped, because CMake passes them through the compiler driver.
Prefix Linker options with `LINKER:`.
CMake will transform it to the appropriate flag for the compiler driver: (Nothing for MSVC, clang-cl and /Qoption,link for ICX), so this will work for the other compilers too, and for earlier versions of CMake.

Fixes: #757

### Checklist
- [ ] All tests pass locally -- (Only verified building)

### DIFF
--- a/cmake/helpers.cmake
+++ b/cmake/helpers.cmake
@@ -295,17 +295,17 @@ function(add_umf_target_link_options name)
         target_link_options(
             ${name}
             PRIVATE
-            /DYNAMICBASE
-            /HIGHENTROPYVA
+            LINKER:/DYNAMICBASE
+            LINKER:/HIGHENTROPYVA
             $<$<C_COMPILER_ID:MSVC>:/DEPENDENTLOADFLAG:0x2000>
             $<$<CXX_COMPILER_ID:MSVC>:/DEPENDENTLOADFLAG:0x2000>
-            /NXCOMPAT)
+            LINKER:/NXCOMPAT)
     endif()
 endfunction()
 
 function(add_umf_target_exec_options name)
     if(MSVC)
-        target_link_options(${name} PRIVATE /ALLOWISOLATION)
+        target_link_options(${name} PRIVATE LINKER:/ALLOWISOLATION)
     endif()
 endfunction()
 
@@ -362,7 +362,7 @@ function(add_umf_library)
 
         if(WINDOWS)
             target_link_options(${ARG_NAME} PRIVATE
-                                /DEF:${ARG_WINDOWS_DEF_FILE})
+                                LINKER:/DEF:${ARG_WINDOWS_DEF_FILE})
         elseif(LINUX)
             target_link_options(${ARG_NAME} PRIVATE
                                 "-Wl,--version-script=${ARG_LINUX_MAP_FILE}")


bratpiorka (approved):

lukaszstolarczuk (commented):
👍

lplewa:
I think we need reproduce this issue on CI - without it, we will break it again, as no one will remember that we have to do this workaround. 

Do we have build with icx on Windows in our CI? I see only a linux one. @lukaszstolarczuk  can you check this?

lukaszstolarczuk:
> I think we need reproduce this issue on CI - without it, we will break it again, as no one will remember that we have to do this workaround.
> 
> Do we have build with icx on Windows in our CI? I see only a linux one. @lukaszstolarczuk can you check this?

Yeah, of course we don't have such job. I tried to add it, but failed with the task 😞 
@Maetveis, could you take a look if I'm missing something...? CMake seems to be unable to compile a simple program with icx in my workflow...

Workflow: https://github.com/lukaszstolarczuk/unified-memory-framework/commit/e3dd2fe22c03db8227bc1f141f1e7a498687be5b

Example CI run: https://github.com/lukaszstolarczuk/unified-memory-framework/actions/runs/11181888055/job/31086892308#step:9:37

Maetveis:
> Yeah, of course we don't have such job. I tried to add it, but failed with the task 😞 @Maetveis, could you take a look if I'm missing something...? CMake seems to be unable to compile a simple program with icx in my workflow...

I don't see anything obviously wrong. I couldn't reproduce the error locally either, it probably has something to do with github's windows image. I wonder what version of cmake is on the path by default for example. https://github.com/mxschmitt/action-tmate might come handy for debugging, though I never used it myself.

ldorau (approved):

ldorau:
> > I think we need reproduce this issue on CI - without it, we will break it again, as no one will remember that we have to do this workaround.
> > Do we have build with icx on Windows in our CI? I see only a linux one. @lukaszstolarczuk can you check this?
> 
> Yeah, of course we don't have such job. I tried to add it, but failed with the task 😞 @Maetveis, could you take a look if I'm missing something...? CMake seems to be unable to compile a simple program with icx in my workflow...
> 
> Workflow: [lukaszstolarczuk@e3dd2fe](https://github.com/lukaszstolarczuk/unified-memory-framework/commit/e3dd2fe22c03db8227bc1f141f1e7a498687be5b)
> 
> Example CI run: https://github.com/lukaszstolarczuk/unified-memory-framework/actions/runs/11181888055/job/31086892308#step:9:37

@lukaszstolarczuk I suppose that the environment variables should be set (the `.\setvars-vcvarsall.bat` and `.\setvars.bat` scripts should be called) at the beginning of each of the following steps: `Configure build`, `Build UMF` and `Run tests` - like we do on Linux:
https://github.com/oneapi-src/unified-memory-framework/blob/main/.github/workflows/basic.yml#L153
https://github.com/oneapi-src/unified-memory-framework/blob/main/.github/workflows/basic.yml#L173
https://github.com/oneapi-src/unified-memory-framework/blob/main/.github/workflows/basic.yml#L179

lukaszstolarczuk:
> > > I think we need reproduce this issue on CI - without it, we will break it again, as no one will remember that we have to do this workaround.
> > > Do we have build with icx on Windows in our CI? I see only a linux one. @lukaszstolarczuk can you check this?
> > 
> > 
> > Yeah, of course we don't have such job. I tried to add it, but failed with the task 😞 @Maetveis, could you take a look if I'm missing something...? CMake seems to be unable to compile a simple program with icx in my workflow...
> > Workflow: [lukaszstolarczuk@e3dd2fe](https://github.com/lukaszstolarczuk/unified-memory-framework/commit/e3dd2fe22c03db8227bc1f141f1e7a498687be5b)
> > Example CI run: https://github.com/lukaszstolarczuk/unified-memory-framework/actions/runs/11181888055/job/31086892308#step:9:37
> 
> @lukaszstolarczuk I suppose that the environment variables should be set (the `.\setvars-vcvarsall.bat` and `.\setvars.bat` scripts should be called) at the beginning of each of the following steps: `Configure build`, `Build UMF` and `Run tests` - like we do on Linux: https://github.com/oneapi-src/unified-memory-framework/blob/main/.github/workflows/basic.yml#L153 https://github.com/oneapi-src/unified-memory-framework/blob/main/.github/workflows/basic.yml#L173 https://github.com/oneapi-src/unified-memory-framework/blob/main/.github/workflows/basic.yml#L179
> 
> because the environment variables set in one step are NOT passed to the next step.

haha, of course, thanks, I'll check it!

lukaszstolarczuk:
@ldorau, thanks, that was the issue, now the workflow is working properly!

@Maetveis, thanks for the patch. I verified it with my workflow and it fixes the issue!
// workflow will be delivered in a sep. PR.

lukaszstolarczuk (approved):

Issue #769, opened 2024-10-01
Can some memory providers not support the free() operation or not?


ldorau:
The File Memory Provider and the DevDax Memory Provider currently do not support the `free()` operation.
They should be used together with the Coarse Memory Provider (#715) or with the jemalloc pool manager that can handle such situation.

It may seem weird that some memory providers do not support the `free()` operation. This issue is just for the discussion of this topic.

Ref: #759

See #759 for the beginning of this discussion: https://github.com/oneapi-src/unified-memory-framework/pull/759#issuecomment-2382742685 :

> > > Another question regarding correctness: when memory tracker is cleared? When the memory provider allocates a memory region it is added to the memory tracker, but if the pool implementation does not release memory back and relies on the fact that the memory provider implementation will release virtual memory mapping who will cleanup the memory tracker? If it is not done then it is an error, because after the memory provider is destroyed the virtual memory address range might be re-used by other memory providers and the same virtual memory region might be added to the tracker and it will fail because the same virtual memory region already presented in the tracker.
> > 
> > 
> > Right. So I replaced it with clearing the tracker for the current pool - please re-review
> 
> I still think that not supporting free operation in the memory provider is weird:
> 
> * First, it means that not every pool manager can be used with such memory provider or user needs to use coarse provider on top of upstream provider that does not support free operation.
> * Second, in this PR you need to iterate via the tracker (that contains records from all pools) and filter only records for the corresponding pool. It might be a performance impact in real life.
> 
> Another consideration, I am working on the cache implementation for opened IPC handles, the cache is global and I need to remove entries from the IPC cache when the corresponding pool is destroyed. It looks like a similar problem. BUt I still think that in your case it is better to require free implementation in all providers.

@vinser52 @bratpiorka @pbalcer

pbalcer:
For fsdax we probably could free pages on destroy (memkind does: https://github.com/memkind/memkind/blob/master/src/memkind_pmem.c#L159). The issue there is performance - deallocating such page is a very expensive operation (since, for fsdax, you are punching a hole in the filesystem) and it may not make sense if we can handle a situation where a provider does not support page deallocation.
So by forcing the provider to provide a suboptimal free implementation, we are leaving performance on the table.

[...]

PR #773, opened 2024-10-03
Move free() to optional (ext) provider ops

ldorau:
### Description

Move `free()` to the optional (ext) provider ops.

Add `umfIsFreeOpDefault(hProvider)` to check
if `op.ext.free` is set to the default function pointer.

Ref: #769

### Checklist

- [x] All tests pass locally


bratpiorka (approved):

test/common/provider_trace.c
```diff
@@ -195,7 +195,7 @@ umf_memory_provider_ops_t UMF_TRACE_PROVIDER_OPS = {
     .initialize = traceInitialize,
     .finalize = traceFinalize,
     .alloc = traceAlloc,
-    .free = traceFree,
+    .ext.free = traceFree,
```
lplewa, line 198: please sort this list here and in other providers. (ext functions should be together)

lplewa (approved):

test/common/provider_trace.c
ldorau: Why? Maybe `.ext.free` could be still after `.alloc` ?

Issue #777, opened 2024-10-07
proxy lib with shm + IPC scenario doesn't work


bratpiorka:
## Please provide a reproduction of the bug:

Currently, 2 process scenario where the first process (client) uses UMF through Proxy Lib with UMF_PROXY="page.disposition=shared-shm" and sends IPC handles to the server, and second process (server) got data from the client and opens IPC handle doesn't work.
This is because the server needs to know the name of the shm, as it is needed as an argument to the OS provider create.

## How often bug is revealed:

always

## Actual behavior:

Server process call to OpenIPCHandle() returns error.

## Additional information about Priority and Help Requested:

Are you willing to submit a pull request with a proposed change? Yes

Requested priority: High

vinser52:
I thought the shared memory name is part of the IPC handle. Or I did not get the exact root cause of the problem. On the server side the `os_open_ipc_handle` function should use the name from the IPC handle.
@ldorau could you please clarify on this?

bratpiorka:
Currently, we open shm that was set during the provider initialization, see https://github.com/oneapi-src/unified-memory-framework/blob/main/src/provider/provider_os_memory.c#L1255

When the user calls OpenIPCHandle() function, it needs a pool handle as an argument, so the provider must already be initialized. Since the shm name is an argument in the provider create function, there is no way to set or change it after initialization. So the one way to support the scenario from this issue is to first get the name of shm in client process, then send it to the server and finally let the server create the provider with shm name as an arg.

I think we can't simply change the code and open the shm from the ipc handle - please consider the scenario where the provider was created with one shm and IPC handle points to the second shm - this is fine to return error in this case.

bratpiorka:
@vinser52 what about devdax and file provider? Can two processes use the same devdax or file provider?

vinser52:
> @vinser52 what about devdax and file provider? Can two processes use the same devdax or file provider?

The same considerations applied to all memory providers that works with memory that is visible/accessible by multiple processes. If two providers allocates from the same physical memory locations then you need a special pool manager that can coordinates between multiple pools (UMF does not have such pool manager, and AFAIK PMDK also had no such one). For cross-process communication we have IPC functionality when memory allocated by one process can be accessed by another process via IPC handle, but allocations from the same physical memory is not supported by multiple pools.

[...]

PR #779, opened 2024-10-07
open_ipc_handle should use shm from handle

bratpiorka:
open_ipc_handle should use shm from handle

see https://github.com/oneapi-src/unified-memory-framework/issues/777 for details

### DIFF
--- a/src/provider/provider_os_memory.c
+++ b/src/provider/provider_os_memory.c
@@ -1146,7 +1146,10 @@ typedef struct os_ipc_data_t {
     int fd;
     size_t fd_offset;
     size_t size;
-    char shm_name[NAME_MAX]; // optional - can be used or not (see below)
+    // shm_name is a Flexible Array Member because it is optional and its size
+    // varies on the Shared Memory object name
+    size_t shm_name_len;
+    char shm_name[];
 } os_ipc_data_t;
 
 static umf_result_t os_get_ipc_handle_size(void *provider, size_t *size) {
@@ -1160,13 +1163,8 @@ static umf_result_t os_get_ipc_handle_size(void *provider, size_t *size) {
         return UMF_RESULT_ERROR_INVALID_ARGUMENT;
     }
 
-    if (os_provider->shm_name[0]) {
-        // os_ipc_data_t->shm_name will be used
-        *size = sizeof(os_ipc_data_t);
-    } else {
-        // os_ipc_data_t->shm_name will NOT be used
-        *size = sizeof(os_ipc_data_t) - NAME_MAX;
-    }
+    // NOTE: +1 for '\0' at the end of the string
+    *size = sizeof(os_ipc_data_t) + strlen(os_provider->shm_name) + 1;
 
     return UMF_RESULT_SUCCESS;
 }
@@ -1195,9 +1193,10 @@ static umf_result_t os_get_ipc_handle(void *provider, const void *ptr,
     os_ipc_data->pid = utils_getpid();
     os_ipc_data->fd_offset = (size_t)value - 1;
     os_ipc_data->size = size;
-    if (os_provider->shm_name[0]) {
-        strncpy(os_ipc_data->shm_name, os_provider->shm_name, NAME_MAX - 1);
-        os_ipc_data->shm_name[NAME_MAX - 1] = '\0';
+    os_ipc_data->shm_name_len = strlen(os_provider->shm_name);
+    if (os_ipc_data->shm_name_len > 0) {
+        strncpy(os_ipc_data->shm_name, os_provider->shm_name,
+                os_ipc_data->shm_name_len);
     } else {
         os_ipc_data->fd = os_provider->fd;
     }
@@ -1222,8 +1221,12 @@ static umf_result_t os_put_ipc_handle(void *provider, void *providerIpcData) {
         return UMF_RESULT_ERROR_INVALID_ARGUMENT;
     }
 
-    if (os_provider->shm_name[0]) {
-        if (strncmp(os_ipc_data->shm_name, os_provider->shm_name, NAME_MAX)) {
+    size_t shm_name_len = strlen(os_provider->shm_name);
+    if (shm_name_len > 0) {
+        if (os_ipc_data->shm_name_len != shm_name_len) {
+            return UMF_RESULT_ERROR_INVALID_ARGUMENT;
+        } else if (strncmp(os_ipc_data->shm_name, os_provider->shm_name,
+                           shm_name_len)) {
             return UMF_RESULT_ERROR_INVALID_ARGUMENT;
         }
     } else {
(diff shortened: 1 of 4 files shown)


test/ipc_os_prov_consumer.c
```diff
@@ -14,8 +14,8 @@
 #include "ipc_os_prov_common.h"
 
 int main(int argc, char *argv[]) {
-    if (argc < 2) {
-        fprintf(stderr, "usage: %s <port> [shm_name]\n", argv[0]);
```
ldorau, line 18: IMHO it should stay as it was. The consumer can use also another SHM file or none.

test/ipc_os_prov_consumer.c
```diff
@@ -25,9 +25,6 @@ int main(int argc, char *argv[]) {
 
     os_params = umfOsMemoryProviderParamsDefault();
     os_params.visibility = UMF_MEM_MAP_SHARED;
-    if (argc >= 3) {
-        os_params.shm_name = argv[2];
-    }
```
ldorau, line 30: IMHO it should stay as it was. The consumer can use also another SHM file or none.

ldorau (changes requested):

src/provider/provider_os_memory.c
```diff
@@ -1160,13 +1160,9 @@ static umf_result_t os_get_ipc_handle_size(void *provider, size_t *size) {
...
     }
 
-    if (os_provider->shm_name[0]) {
-        // os_ipc_data_t->shm_name will be used
-        *size = sizeof(os_ipc_data_t);
-    } else {
-        // os_ipc_data_t->shm_name will NOT be used
-        *size = sizeof(os_ipc_data_t) - NAME_MAX;
-    }
+    // for simplicity, always return size of the full structure even if we
```
vinser52, line 1163: IPC handle is send/copied between processes by the client, so with this change the `NAME_MAX` bytes will be copied every time even they are not used.

bratpiorka: As I wrote - this is for simplicity. Do you think we should focus on optimizing this now?

bratpiorka: Yes, the current test allocates the buffer for IPC handle based on umfMemoryProviderGetIPCHandleSize(provider, ..) where the "provider" is the consumer provider, not the one used in the producer.

bratpiorka: I will change the test so that the producer sends the IPC handle size first, so that the consumer can allocate a buffer of the correct size instead of calling umfMemoryProviderGetIPCHandleSize()

test/ipc_os_prov_shm.sh
```diff
@@ -20,7 +20,7 @@ UMF_LOG_VAL="level:debug;flush:debug;output:stderr;pid:yes"
 rm -f /dev/shm/${SHM_NAME}
 
 echo "Starting ipc_os_prov_shm CONSUMER on port $PORT ..."
-UMF_LOG=$UMF_LOG_VAL ./umf_test-ipc_os_prov_consumer $PORT $SHM_NAME &
```
lukaszstolarczuk, line 23: should you also update `ipc_ipcapi` example?

src/provider/provider_os_memory.c
vinser52: > 

@vinser52 It makes possible receiving IPC handle of greater size (with SHM name) when the current process does not use SHM (and the IPCGetHandleSize() had returned smaller size (without SHM name) previously).

good point.

src/provider/provider_os_memory.c
```diff
@@ -1251,14 +1247,14 @@ static umf_result_t os_open_ipc_handle(void *provider, void *providerIpcData,
     umf_result_t ret = UMF_RESULT_SUCCESS;
     int fd;
 
-    if (os_provider->shm_name[0]) {
-        fd = utils_shm_open(os_provider->shm_name);
+    if (os_ipc_data->shm_name[0]) {
```
vinser52, line 1250: Is the memory zeroed on the producer side? This check make sense only if the `shm_name` string is zeroed on the producer side.

bratpiorka: yes, see L#245

vinser52: Another question, I see that you returned size optimization when shm_name is not used. In that case, we cannot check `os_ipc_data->shm_name[0]` because it is out of bound of the `os_ipc_data` structure.

bratpiorka: But the in case shm is not used we still can check the first byte - if it is set to 0 then we don't read the rest

test/ipc_os_prov_shm.sh
bratpiorka: The IPC example was written better :) but I changed the bash file that runs it

src/provider/provider_os_memory.c
ldorau: There should be added `+1`:
```c
        // os_ipc_data_t->shm_name will NOT be used
        *size = sizeof(os_ipc_data_t) - NAME_MAX + 1;
```
so that the first byte is accessible.
And later in the `os_get_ipc_handle()`:
```c
    if (os_provider->shm_name[0]) {
        strncpy(os_ipc_data->shm_name, os_provider->shm_name, NAME_MAX - 1);
        os_ipc_data->shm_name[NAME_MAX - 1] = '\0';
    } else {
        os_ipc_data->fd = os_provider->fd;
        os_ipc_data->shm_name[0] = '\0'; // <<<<<<<<<<<<<<<< THIS LINE SHOULD BE ADDED
    }

```

bratpiorka: I implemented this as Łukasz D suggested - this is potentially less error-prone in the future

ldorau (approved):

vinser52 (approved):

lplewa (changes requested):

src/provider/provider_os_memory.c
```diff
@@ -1165,7 +1165,7 @@ static umf_result_t os_get_ipc_handle_size(void *provider, size_t *size) {
         *size = sizeof(os_ipc_data_t);
     } else {
         // os_ipc_data_t->shm_name will NOT be used
-        *size = sizeof(os_ipc_data_t) - NAME_MAX;
+        *size = sizeof(os_ipc_data_t) - NAME_MAX + 1;
```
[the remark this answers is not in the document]
bratpiorka: Is it a blocking comment? what do you suggest? Please provider e.g. code snippet or other kind of constructive feedback :)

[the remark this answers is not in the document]
bratpiorka: Also I would suggest to fix this in separate PR as this is used in multiple places - please fill an issue if needed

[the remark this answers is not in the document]
lplewa: > Also I would suggest to fix this in separate PR as this is used in multiple places - please fill an issue if needed

I did fast grep over code. We use flexible array members in few places (what is correct), but i cannot find another case of this solution. Can use share where we are using it?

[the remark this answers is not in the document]
bratpiorka: @lplewa ok I didn't understand you at first - I made appropriate changes. Is it ok now?

src/provider/provider_os_memory.c
```diff
@@ -1146,7 +1146,10 @@ typedef struct os_ipc_data_t {
     int fd;
     size_t fd_offset;
     size_t size;
-    char shm_name[NAME_MAX]; // optional - can be used or not (see below)
+    // shm_name is a Flexible Array Member because it is optional and its size
+    // varies on the Shared Memory object name
+    size_t shm_name_len;
```
lukaszstolarczuk, line 1151: soo... should we get rid of the `NAME_MAX` define now...?

bratpiorka: NAME_MAX is used in other places and it is defined in limits.h

lukaszstolarczuk: I guess my question is - should we still use it in other places...?

vinser52: I think yes. User should know the IPC handle size on the producer and on the consumer sides.

bratpiorka: @vinser52 this means that we should have an api "umfGetIPCHandleSize(ipc_handle)" ?

lplewa (approved):

Issue #796, opened 2024-10-14
Multiple allocations from file provider doesn't work


bratpiorka:
Multiple allocations from the file provider don't work - in case of multiple allocations from the file provider, all get the same pointer.

Important: when fixing this issue, make sure multiple-allocation tests are added to all providers.

## Environment Information

- UMF version (hash commit or a tag):  latest
- OS(es) version(s): Linux

## Please provide a reproduction of the bug:
```

--- a/test/provider_file_memory.cpp
+++ b/test/provider_file_memory.cpp
@@ -80,13 +80,19 @@ struct FileProviderParamsShared : FileProviderParamsDefault {};
 static void test_alloc_free_success(umf_memory_provider_handle_t provider,
                                     size_t size, size_t alignment,
                                     purge_t purge) {
-    void *ptr = nullptr;
+    void *ptr = nullptr, *ptr2 = nullptr;
 
     umf_result_t umf_result =
         umfMemoryProviderAlloc(provider, size, alignment, &ptr);
     ASSERT_EQ(umf_result, UMF_RESULT_SUCCESS);
     ASSERT_NE(ptr, nullptr);
 
+    umf_result = umfMemoryProviderAlloc(provider, size, alignment, &ptr2);
+    ASSERT_EQ(umf_result, UMF_RESULT_SUCCESS);
+    ASSERT_NE(ptr2, nullptr);
+
+    ASSERT_NE(ptr, ptr2);
+
     memset(ptr, 0xFF, size);
 
     if (purge == PURGE_LAZY) {
```

## How often bug is revealed:

always

ldorau:
@bratpiorka Fix submitted: https://github.com/oneapi-src/unified-memory-framework/pull/800

PR #800, opened 2024-10-14
Fix `file_alloc()` of the file memory provider

ldorau:
### Description

Fix `file_alloc()` of the file memory provider.

Fixes: #796

### Checklist
- [x] All tests pass locally

### DIFF
--- a/src/provider/provider_file_memory.c
+++ b/src/provider/provider_file_memory.c
@@ -263,22 +263,38 @@ static umf_result_t file_mmap_aligned(file_memory_provider_t *file_provider,
         return UMF_RESULT_ERROR_INVALID_ARGUMENT; // arithmetic overflow
     }
 
-    if (offset_fd + extended_size > size_fd) {
-        if (utils_fallocate(fd, offset_fd, extended_size)) {
+    // offset_fd has to be also page-aligned since it is the offset of mmap()
+    size_t aligned_offset_fd = offset_fd;
+    rest = aligned_offset_fd & (page_size - 1);
+    if (rest) {
+        aligned_offset_fd += page_size - rest;
+    }
+    if (aligned_offset_fd < offset_fd) {
+        LOG_ERR("arithmetic overflow of file offset");
+        return UMF_RESULT_ERROR_INVALID_ARGUMENT; // arithmetic overflow
+    }
+
+    if (aligned_offset_fd + extended_size > size_fd) {
+        size_t new_size_fd = aligned_offset_fd + extended_size;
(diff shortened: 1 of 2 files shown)


bratpiorka (approved):

lukaszstolarczuk (commented):
👍

test/common/test_helpers.cpp
```diff
@@ -0,0 +1,57 @@
+// Copyright (C) 2023 Intel Corporation
```
lplewa, line 1: 2024

test/common/test_helpers.cpp
```diff
@@ -0,0 +1,57 @@
...
+// Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
```
lplewa, line 2: Also we have test_helpers.c and now you are adding test_helpers.cpp with other "c" functions. 

First of all, we are writing tests in c++ so we should use c++ features. We should have a base class for all "provider tests" which has this method. Then extend this class with provider specific test classes for given provider.

lplewa (changes requested):

test/common/test_helpers.cpp
ldorau: This change: "_We should have a base class for all "provider tests" which has this method. Then extend this class with provider specific test classes for given provider._" requires separate pull request IMHO.

The most important part of this PR is the fix for the file provider, because it is unusable at all now and this fix should be merged ASAP IMHO. Refactoring of tests can wait and should be done in a separate PR.

ldorau: If you insist, I can remove this and add only one simple test instead of this, like here: 
https://github.com/ldorau/unified-memory-framework/commit/d707a8b8bdece406aafbc5c62f78f719bb78317b

test/common/test_helpers.cpp
ldorau: Done

test/common/test_helpers.cpp
lplewa: An idea to have common test shared between providers was good. But it was implemented in wrong way.  If you do not want to do it clean way, we can do your simple test.  anyway we should have even better test for each provider which does multiple allocations and fire in (semi) random way, but it might be a topic for another day.

ldorau: First I have to provide the very important fix for the file provider - this is the most important thing now.
I added a simple test.

@lplewa Done

ldorau:
Code coverage of `provider_file_memory.c` increased from 76.8 % to 77.9 % (by 1.1 %).

lplewa (approved):

PR #806, opened 2024-10-17
Add parameter to destroy upstream_memory_provider in finalize()

ldorau:
### Description

Add parameter to destroy `upstream_memory_provider` in `finalize()`.

### Checklist
- [x] All tests pass locally

### DIFF
--- a/src/provider/provider_coarse.c
+++ b/src/provider/provider_coarse.c
@@ -34,6 +34,9 @@ typedef struct coarse_memory_provider_t {
     // memory allocation strategy
     coarse_memory_provider_strategy_t allocation_strategy;
 
+    // destroy upstream_memory_provider in finalize()
+    bool destroy_upstream_memory_provider;
+
     void *init_buffer;
 
     size_t used_size;
@@ -898,6 +901,13 @@ static umf_result_t coarse_memory_provider_initialize(void *params,
         return UMF_RESULT_ERROR_INVALID_ARGUMENT;
     }
 
+    if (coarse_params->destroy_upstream_memory_provider &&
+        !coarse_params->upstream_memory_provider) {
+        LOG_ERR("destroy_upstream_memory_provider is true, but an upstream "
+                "provider is not provided");
+        return UMF_RESULT_ERROR_INVALID_ARGUMENT;
+    }
+
     coarse_memory_provider_t *coarse_provider =
         umf_ba_global_alloc(sizeof(*coarse_provider));
     if (!coarse_provider) {
@@ -909,6 +919,8 @@ static umf_result_t coarse_memory_provider_initialize(void *params,
 
     coarse_provider->upstream_memory_provider =
         coarse_params->upstream_memory_provider;
+    coarse_provider->destroy_upstream_memory_provider =
+        coarse_params->destroy_upstream_memory_provider;
     coarse_provider->allocation_strategy = coarse_params->allocation_strategy;
     coarse_provider->init_buffer = coarse_params->init_buffer;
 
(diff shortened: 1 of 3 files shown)


bratpiorka:
why do we need this parameter? If the user creates an upstream provider, why can't he destroy it?

ldorau:
> why do we need this parameter? If the user creates an upstream provider, why can't he destroy it?

In order to prevent a memory leak in our tests using the `umfPoolTest` fixture (see: #808) the upstream provider has to be destroyed automatically.

igchor:
> > why do we need this parameter? If the user creates an upstream provider, why can't he destroy it?
> 
> In order to prevent a memory leak in our tests using the `umfPoolTest` fixture (see: #808) the upstream provider has to be destroyed automatically.

If it's only for tests then you can prevent the leak by just destroying the provider in pool_unique_handle_t. You just need to modify this line:
https://github.com/oneapi-src/unified-memory-framework/blob/e802af834525825be60dd6581b8b472338536801/test/poolFixtures.hpp#L53 

To be something like this:

```
auto dtor = [provider](umf_memory_pool_handle_t pool) {
    umfPoolDestory(pool):
    umfMemoryProviderDestory(provider);
};
return umf::pool_unique_handle_t(hPool, std;:move(dtor));
```

ldorau:
> > > why do we need this parameter? If the user creates an upstream provider, why can't he destroy it?
> > 
> > 
> > In order to prevent a memory leak in our tests using the `umfPoolTest` fixture (see: #808) the upstream provider has to be destroyed automatically.
> 
> If it's only for tests then you can prevent the leak by just destroying the provider in pool_unique_handle_t. You just need to modify this line:
> 
> https://github.com/oneapi-src/unified-memory-framework/blob/e802af834525825be60dd6581b8b472338536801/test/poolFixtures.hpp#L53
> 
> To be something like this:
> 
> ```
> auto dtor = [coarse_provider ](umf_memory_pool_handle_t pool) {
>     umfPoolDestory(pool):
>     umfMemoryProviderDestory(coarse_provider );
> };
> return umf::pool_unique_handle_t(hPool, std;:move(dtor));
> ```

Thanks @igchor ! I applied that in #808.

ldorau:
> > why do we need this parameter? If the user creates an upstream provider, why can't he destroy it?
> 
> @igchor @bratpiorka @ldorau
> 
> But what if the user creates a memory pool with the `UMF_POOL_CREATE_FLAG_OWN_PROVIDER` flag? In that case, the user expects that the pool destructor will destroy the memory provider used by the pool. And in case the coarse provider is used, it is destroyed automatically, but the upstream memory provider is not.
> 
> Did I miss something?

@vinser52 You are right. It seems this parameter is needed when `UMF_POOL_CREATE_FLAG_OWN_PROVIDER` is used.
@bratpiorka @lplewa Do you agree?

bratpiorka:
@ldorau agree + please add test for this (for both cases - when UMF_POOL_CREATE_FLAG_OWN_PROVIDER is set to true and false)

ldorau:
> @ldorau agree + please add test for this (for both cases - when UMF_POOL_CREATE_FLAG_OWN_PROVIDER is set to true and false)

@bratpiorka  Done

bratpiorka (approved):

lukaszstolarczuk (commented):
👍

src/provider/provider_coarse.c
```diff
@@ -898,6 +901,13 @@ static umf_result_t coarse_memory_provider_initialize(void *params,
         return UMF_RESULT_ERROR_INVALID_ARGUMENT;
     }
 
+    if (coarse_params->destroy_upstream_memory_provider &&
+        !coarse_params->upstream_memory_provider) {
```
vinser52, line 905: Should the coarse provider always require the upstream provider to be specified?

What is the reason for not specifying the upstream provider? @bratpiorka what do you think?

P.S. I understand that it is out of scope for this particular PR.

ldorau: The coarse provider contains the fixed provider yet - in the case of the fixed provider the upstream provider is not required. The fixed provider will be moved to a separate provider and the upstream provider will be always required then.

PR #808, opened 2024-10-17
Add optional coarse provider to `umfPoolTest` fixture

ldorau:
### Description

Add optional coarse provider to `umfPoolTest` fixture.
Add 5th argument (`coarse_memory_provider_params_t *`)
to the `poolCreateExtParams` tuple (can be `nullptr`).
If it is `non-nullptr`, coarse provider is created
with the given provider as the upstream provider.

It will be used to test the following providers:
- file provider
- devdax provider

as the upstream providers with the coarse provider
and different pool managers.

Requires:
- [x] #806 

### Checklist

- [x] All tests pass locally

### DIFF
--- a/test/memoryPoolAPI.cpp
+++ b/test/memoryPoolAPI.cpp
@@ -181,14 +181,16 @@ TEST_F(test, BasicPoolByPtrTest) {
 INSTANTIATE_TEST_SUITE_P(
     mallocPoolTest, umfPoolTest,
     ::testing::Values(poolCreateExtParams{&MALLOC_POOL_OPS, nullptr,
-                                          &UMF_NULL_PROVIDER_OPS, nullptr},
+                                          &UMF_NULL_PROVIDER_OPS, nullptr,
+                                          nullptr},
                       poolCreateExtParams{umfProxyPoolOps(), nullptr,
-                                          &MALLOC_PROVIDER_OPS, nullptr}));
+                                          &MALLOC_PROVIDER_OPS, nullptr,
+                                          nullptr}));
 
 INSTANTIATE_TEST_SUITE_P(mallocMultiPoolTest, umfMultiPoolTest,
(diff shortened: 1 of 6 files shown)


test/disjointCoarseMallocPool.cpp
```diff
@@ -282,8 +286,10 @@ TEST_P(CoarseWithMemoryStrategyTest, disjointCoarseMallocPool_basic) {
     ASSERT_EQ(GetStats(prov).alloc_size, init_buffer_size);
 
     umfPoolDestroy(pool);
-    umfMemoryProviderDestroy(coarse_memory_provider);
-    umfMemoryProviderDestroy(malloc_memory_provider);
+    // Both coarse_memory_provider and malloc_memory_provider
```
PatKamin, line 289: You can add a check in the test below that umfPoolDestroy doesn't destroy the upstream memory provider (`OWN_PROVIDER` flag not set and the new parameter set to false by default).

ldorau: @PatKamin There is already such test: `disjointCoarseMallocPool_simple1` for example

PatKamin: Yes, what I meant is to add a check after `umfDestroyPool`, like `GTEST_NE(provider, nullptr)` in the test you've mentioned.

ldorau: Such check gives us nothing, because destroying a provider or not does not change the provider's pointer saved in a variable: after calling `umfMemoryProviderDestroy(coarse_memory_provider);` the `coarse_memory_provider` is still `non-NULL`.

PatKamin: Ok

PatKamin (approved):

lukaszstolarczuk (approved):

PR #813, opened 2024-10-21
fix: substitute UT_ASSERTs with asserts from GTEST part last

EuphoricThinking:
Amen.

ref. #569

### Checklist

- [ ] All tests pass locally

### DIFF
--- a/test/memspaces/memspace_host_all.cpp
+++ b/test/memspaces/memspace_host_all.cpp
@@ -126,7 +126,7 @@ TEST_F(memspaceHostAllProviderTest, hostAllDefaults) {
     EXPECT_BIND_MASK_EQ(ptr1, ptr2);
 
     auto ret2 = munmap(ptr2, size);
-    UT_ASSERTeq(ret2, 0);
+    ASSERT_EQ(ret2, 0);
 
     ret = umfMemoryProviderFree(hProvider, ptr1, size);
     ASSERT_EQ(ret, UMF_RESULT_SUCCESS);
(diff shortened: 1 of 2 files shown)


pbalcer (approved):
... and let's never speak of this again.

Might be useful to remove [all these asserts](https://github.com/oneapi-src/unified-memory-framework/blob/7ed878716bfaa76067d542d947a03b1de9a6ceac/test/common/test_helpers.h#L25-L76) entirely.

EuphoricThinking:
> Might be useful to remove [all these asserts](https://github.com/oneapi-src/unified-memory-framework/blob/7ed878716bfaa76067d542d947a03b1de9a6ceac/test/common/test_helpers.h#L25-L76) entirely.

These asserts are used in .c files in test/c_api, since GTEST is C++ framework

lukaszstolarczuk (approved):

pbalcer:
> > Might be useful to remove [all these asserts](https://github.com/oneapi-src/unified-memory-framework/blob/7ed878716bfaa76067d542d947a03b1de9a6ceac/test/common/test_helpers.h#L25-L76) entirely.
> 
> These asserts are used in .c files in test/c_api, since GTEST is C++ framework. Do you mean replacing using another framework?

I'm a bit worried that this same issue will creep up again. No, using another framework just for the C api isn't worth it.

One option would be to rewrite those tests to use gtest. At first I thought this was done without gtestso that it's possible to compile and test this with a C compiler. But, looking at CMake's, it looks as if all the tests are compiled the same way.

Alternatively, we could move those test_helpers into the `c_api` directory to make sure it's not possible to use those C-like asserts in the C++ part of the codebase.

vinser52 (approved):

lplewa:
> > > Might be useful to remove [all these asserts](https://github.com/oneapi-src/unified-memory-framework/blob/7ed878716bfaa76067d542d947a03b1de9a6ceac/test/common/test_helpers.h#L25-L76) entirely.
> > 
> > 
> > These asserts are used in .c files in test/c_api, since GTEST is C++ framework. Do you mean replacing using another framework?
> 
> I'm a bit worried that this same issue will creep up again. No, using another framework just for the C api isn't worth it.
> 
> One option would be to rewrite those tests to use gtest. At first I thought this was done without gtest so that it's possible to compile and test this with a C compiler. But, looking at CMake's, it looks as if all the tests are compiled the same way.
> 
> Alternatively, we could move those test_helpers into the `c_api` directory to make sure it's not possible to use those C-like asserts in the C++ part of the codebase.

There is few IPC tests which are implemented without gtest as they require multiple processes. So we cannod rid of those UTASSERTS

lplewa (approved):

PR #823, opened 2024-10-22
use pools not only providers in IPC tests

bratpiorka:
* change IPC tests to use Pool (Scalable - requires TBB, Disjoint for GPU) instead of Provider API
* add 60 sec timeout for IPC tests


vinser52 (approved):
Yeah, I also considered changing this test to use pools instead of providers.

ldorau (approved):

PR #824, opened 2024-10-22
Add additional tests for CUDA and L0 providers

vinser52:
### Description
To improve code coverage the new tests for CUDA and L0 providers are added. In next PR this tests can be refactored to avoid code duplications.

### Checklist
- [x] All tests pass locally

### DIFF
--- a/test/providers/provider_level_zero.cpp
+++ b/test/providers/provider_level_zero.cpp
@@ -141,14 +141,9 @@ struct umfLevelZeroProviderTest
 
         auto [l0_params, accessor] = this->GetParam();
         params = l0_params;
-        hDevice = (ze_device_handle_t)params.level_zero_device_handle;
+        memAccessor = accessor;
         hContext = (ze_context_handle_t)params.level_zero_context_handle;
 
-        if (params.memory_type == UMF_MEMORY_TYPE_HOST) {
-            ASSERT_EQ(hDevice, nullptr);
-        } else {
-            ASSERT_NE(hDevice, nullptr);
-        }
         ASSERT_NE(hContext, nullptr);
 
         switch (params.memory_type) {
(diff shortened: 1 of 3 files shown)


test/providers/provider_level_zero.cpp
```diff
@@ -222,6 +210,97 @@ TEST_P(umfLevelZeroProviderTest, basic) {
...
+
+    umf_result = umfMemoryProviderGetMinPageSize(provider, nullptr, nullptr);
+    ASSERT_EQ(umf_result, UMF_RESULT_ERROR_INVALID_ARGUMENT);
+
+    umf_result = umfMemoryProviderGetRecommendedPageSize(provider, 0, nullptr);
+    ASSERT_EQ(umf_result, UMF_RESULT_ERROR_INVALID_ARGUMENT);
+
+    umfMemoryProviderDestroy(provider);
+}
+
```
bratpiorka, line 303: please remove TODO from L305 :)

test/providers/provider_level_zero.cpp
```diff
@@ -167,21 +162,14 @@ struct umfLevelZeroProviderTest
         }
 
         ASSERT_NE(zeMemoryTypeExpected, ZE_MEMORY_TYPE_UNKNOWN);
-
-        memAccessor = accessor;
     }
 
-    void TearDown() override {
-        int ret = destroy_context(hContext);
```
bratpiorka, line 175: where do you call destroy context?

ldorau (changes requested):
Could you add tests for `cu_memory_provider_get_last_native_error` either?
`cu_memory_provider_get_last_native_error` and `cu_store_last_native_error` are not used at all.

ldorau:
> Could you add tests for `cu_memory_provider_get_last_native_error` either? `cu_memory_provider_get_last_native_error` and `cu_store_last_native_error` are not used at all.

It can be in a separate (next) PR if you prefer.

ldorau (approved):
> Could you add tests for `cu_memory_provider_get_last_native_error` either? `cu_memory_provider_get_last_native_error` and `cu_store_last_native_error` are not used at all.

It can be in a separate (next) PR if you prefer.

test/providers/provider_level_zero.cpp
vinser52: Done, removed.

Right now only the following `TODO` comments remains:
```c
// TODO add tests that mixes Level Zero Memory Provider and Disjoint Pool
```

test/providers/provider_level_zero.cpp
vinser52: `l0Params_device_memory`, `l0Params_shared_memory`, `l0Params_host_memory` are global variables, see L306 - L311. So it is an error to destroy the context in the `TearDown` method because it is called after each test.

The same is true for the CUDA tests

vinser52:
> > Could you add tests for `cu_memory_provider_get_last_native_error` either? `cu_memory_provider_get_last_native_error` and `cu_store_last_native_error` are not used at all.
> 
> It can be in a separate (next) PR if you prefer.

Yeah, better to add them in a separate PR. Need to think about the scenario when CUDA provider returns provider-specific error.

lukaszstolarczuk (approved):

PR #825, opened 2024-10-23
Reverse UMF_BUILD_LEVEL_ZERO_PROVIDER and UMF_BUILD_CUDA_PROVIDER

ldorau:
### Description

`UMF_BUILD_LEVEL_ZERO_PROVIDER` and `UMF_BUILD_CUDA_PROVIDER` are `OFF` by default from now on.

### Checklist
- [x] All tests pass locally


bratpiorka:
please update run_build.sh also

vinser52:
Why we are disabling CUDA and L0 by default?

ldorau:
> please update run_build.sh also

It is OK now.

bratpiorka:
@vinser52 they will still be a part of the build for oneAPI. 
Here we disable these options by default so that we don't break CI systems that don't have access to github and can't download CUDA headers, which is a cmake step when you choose to compile a CUDA provider. I think building the CUDA provider could be further improved anyway, but in the separate PR (issue https://github.com/oneapi-src/unified-memory-framework/issues/827)

vinser52:
> @vinser52 they will still be a part of the build for oneAPI. Here we disable these options by default so that we don't break CI systems that don't have access to github and can't download CUDA headers, which is a cmake step when you choose to compile a CUDA provider. I think building the CUDA provider could be further improved anyway, but in the separate PR (issue #827)

But it is a question about defaults whether we want developer to disable CUDA and L0 on the system where download is not possible or do we want to enable it in all other cases. I prefer to have them ON by default and disable only if build system cannot download headers.

bratpiorka:
> But it is a question about defaults whether we want developer to disable CUDA and L0 on the system where download is not possible or do we want to enable it in all other cases. I prefer to have them ON by default and disable only if build system cannot download headers.

But how to check this? The problem occurs on intel/llvm (see https://github.com/intel/llvm/pull/15686). Please note that these CMAKE options are set by UR in this case, not directly in llvm.

vinser52:
> > But it is a question about defaults whether we want developer to disable CUDA and L0 on the system where download is not possible or do we want to enable it in all other cases. I prefer to have them ON by default and disable only if build system cannot download headers.
> 
> But how to check this? The problem occurs on intel/llvm (see [intel/llvm#15686](https://github.com/intel/llvm/pull/15686)). Please note that these CMAKE options are set by UR in this case, not directly in llvm.

I see… Is it OK to have UMF build without CUDA and L0 providers for intel/llvm? Today UR uses its own implementation of L0 and CUDA providers, but in a long term UR should use ones from UMF. I think the right solution in that case is to have L0/CUDA headers to be part of UMF repo.

pbalcer (approved):
We should do what intel/llvm and UR does - make extra features opt-in. UR doesn't even build the level-zero adapter by default.
We can't be breaking other people's builds with extraneous dependencies that they don't want or need.

Once the CUDA adapter is ready, there should be a cmake variable that specifies where are the headers. UR will supply the one it uses when the CUDA adapter is built (which is also an opt-in), and this will avoid the download.

vinser52:
> > I see… Is it OK to have UMF build without CUDA and L0 providers for intel/llvm? Today UR uses its own implementation of L0 and CUDA providers, but in a long term UR should use ones from UMF. I think the right solution in that case is to have L0/CUDA headers to be part of UMF repo.
> 
> I'm not sure. I think there are scenarios where they already have these headers on their system. For L0 we got a CMAKE var to handle this: UMF_LEVEL_ZERO_INCLUDE_DIR - it is currently used in UR. We should add a similar var for CUDA

What is the reason to use user provided headers? As I can understand, there are 2 possible cases:

- User provided headers are the same like we have/downloaded in our repo.
- User provided headers are different (e.g. L0 API that we are using is changed) but in that case we need to fix our implementation anyway. It is not only a question to update headers.

pbalcer:
> What is the reason to use user provided headers? As I can understand, there are 2 possible cases:

So that we avoid a download from a third-party.

vinser52:
> > What is the reason to use user provided headers? As I can understand, there are 2 possible cases:
> 
> So that we avoid a download from a third-party.

Sorry, I was not clear. My question was about option when we just commit CUDA/L0 headers to our repo to avoid download step.

pbalcer:
> > > What is the reason to use user provided headers? As I can understand, there are 2 possible cases:
> > 
> > 
> > So that we avoid a download from a third-party.
> 
> Sorry, I was not clear. My question was about option when we just commit CUDA/L0 headers to our repo to avoid download step.

Why would we do that? At that point, you are responsible for redistributing the cuda headers, which complicates the release process.

Adding a variable to specify the headers is exactly what other components do (pretty much all the adapters in UR have this option).

vinser52:
I think what @pbalcer suggested is doable but from my point of view it is overcomplication (vs just commiting the headers to the repo) and for me it is not clear for what reasons (I mean what is the benefits to use user provided headers).

vinser52:
> > > > What is the reason to use user provided headers? As I can understand, there are 2 possible cases:
> > > 
> > > 
> > > So that we avoid a download from a third-party.
> > 
> > 
> > Sorry, I was not clear. My question was about option when we just commit CUDA/L0 headers to our repo to avoid download step.
> 
> Why would we do that? At that point, you are responsible for redistributing the cuda headers, which complicates the release process.
> 
> Adding a variable to specify the headers is exactly what other components do (pretty much all the adapters in UR have this option).


I am worried that it might cause the situation when we will have two versions of `libumf.so`. One from `intel/llvm` and another from oneAPI kit. But if there is a questions with CUDA license… But we are not going to redistribute CUDA headers, we are going to use them during the build.

pbalcer:
> I think what @pbalcer suggested is doable but from my point of view it is overcomplication (vs just commiting the headers to the repo) and for me it is not clear for what reasons (I mean what is the benefits to use user provided headers).

What exactly is the problem though? We would align the behavior of UMFs build system to that of the rest of the stack. Checking-in a third-party component (which is not that small) with a proprietary license, adding the extra burden of maintaining that, for I'm not exactly sure what purpose?

Especially since, in the compiler build from where cuda is enabled, those headers already exist. We don't need that extra copy.

vinser52:
And to be clear, I am not strictly against this changes. I just raised questions/concerns.

pbalcer:
> Why should we change defaults and disable L0/CUDA by default? If UR always knows if it build L0/CUDA adapter or not than UR always can pass the ON/OFF options for L0/CUDA providers.

On that I don't have a strong opinion. But we need to either change the defaults here or always make sure we disable all we don't need in UR when we do FetchContent. My slight preference would be to disabling everything UMF (just to be consistent with UR).

vinser52:
> > Why should we change defaults and disable L0/CUDA by default? If UR always knows if it build L0/CUDA adapter or not than UR always can pass the ON/OFF options for L0/CUDA providers.
> 
> On that I don't have a strong opinion. But we need to either change the defaults here or always make sure we disable all we don't need in UR when we do FetchContent. My slight preference would be to disabling everything UMF (just to be consistent with UR).

I think you are biased because you are working on UR :)
From UMF perspective, if we are changing defaults of UMF than developers who use UMF should always enable CUDA/L0 adapters. And the only case when it should be disabled is when network is not accessible and headers cannot be downloaded from the github.

vinser52:
> we download L0 headers only when tests and examples are built (which is disabled by default). So L0 headers wouldn't be fetched by default while L0 provider would be built.

I did not know that. I thought that we download L0 headers to build the L0 provider as well. In that case it is another point not to disable (at least L0 provider) by default.

Regarding CUDA we can investigate how it would be to get rid of `cuda.h` to build the CUDA provider.

PR #834, opened 2024-10-24
Fix devdax_open_ipc_handle() and devdax_close_ipc_handle()

ldorau:
### Description

Fix devdax_open_ipc_handle() and devdax_close_ipc_handle()

devdax_open_ipc_handle() has to use the path of the remote
/dev/dax got from the IPC handle, not the local one.

devdax_open_ipc_handle() has to use also the memory protection
got from the IPC handle, so let's add the memory protection
to the IPC handle.

devdax_open_ipc_handle() should mmap only the required range of memory,
not the whole /dev/dax device, so let's add the length of the allocation
to the IPC handle.

devdax_close_ipc_handle() should unmap only the required range of memory,
not the whole /dev/dax device.

### Checklist

- [x] All tests pass locally

### DIFF
--- a/src/provider/provider_devdax_memory.c
+++ b/src/provider/provider_devdax_memory.c
@@ -369,9 +369,10 @@ static umf_result_t devdax_allocation_merge(void *provider, void *lowPtr,
 }
 
 typedef struct devdax_ipc_data_t {
-    char dd_path[PATH_MAX]; // path to the /dev/dax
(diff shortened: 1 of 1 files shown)


bratpiorka (approved):

ldorau:
It requires more changes, closing for now, will be reopened

ldorau:
Reopened in https://github.com/oneapi-src/unified-memory-framework/pull/838

PR #839, opened 2024-10-24
[WIP] Add IPC tests (`umfIpcTest`) to the devdax provider

ldorau:
### Description

Add IPC tests (`umfIpcTest`) to the devdax provider.

### Checklist
- [x] All tests pass locally


src/provider/provider_devdax_memory.c
```diff
@@ -36,12 +37,13 @@ umf_memory_provider_ops_t *umfDevDaxMemoryProviderOps(void) {
...
-    size_t offset;       // offset in the file used for memory mapping
-    utils_mutex_t lock;  // lock of ptr and offset
-    unsigned protection; // combination of OS-specific protection flags
+    char path[PATH_MAX];         // a path to the device DAX
+    size_t size;                 // size of the file used for memory mapping
+    void *base;                  // base address of memory mapping
+    size_t offset;               // offset in the file used for memory mapping
+    utils_mutex_t lock;          // lock of ptr and offset
+    unsigned protection;         // combination of OS-specific protection flags
+    bool ipc_consumer_only_mode; // when path==NULL and size==0
```
vinser52, line 46: I am not sure that we do need explicit `ipc_consumer_only_mode` flag. Alternatively we can allow path to be empty in that case allocations would not be possible.

ldorau: But `path` is not passed to all API ops.

ldorau: > But `path` is not passed to all API ops.

Wait, it is.

ldorau: So, do you suggest to always check `if (path == NULL)` ? I think checking `ipc_consumer_only_mode` is more clear.

ldorau: BTW. please review https://github.com/oneapi-src/unified-memory-framework/pull/831 instead of this one

ldorau:
Replaced with https://github.com/oneapi-src/unified-memory-framework/pull/845

PR #761, opened 2024-09-26
enable tracker in ProxyLib by default

bratpiorka:
### Description

Enable Tracking Provider in the Proxy Lib by default.

### Checklist

- [x] Code compiles without errors locally
- [x] All tests pass locally
- [x] CI workflows execute properly
- [x] New tests added, especially if they will fail without my changes
- [x] Added/extended example(s) to cover this functionality
- [x] Extended the README/documentation
- [x] Logger (with debug/info/... messages) is used

PR #845, opened 2024-10-28
Add IPC tests (`umfIpcTest`) to the devdax provider

ldorau:
### Description

Add IPC tests (`umfIpcTest`) to the devdax provider.

### Checklist

- [x] All tests pass locally

### DIFF
--- a/test/ipcFixtures.hpp
+++ b/test/ipcFixtures.hpp
@@ -352,9 +352,11 @@ TEST_P(umfIpcTest, AllocFreeAllocTest) {
               get_umf_result_of_free(freeNotSupported, UMF_RESULT_SUCCESS));
 
     pool.reset(nullptr);
-    EXPECT_EQ(stat.allocCount, stat.getCount);
+    // TODO fix it - it does not work in case of IPC cache hit
+    // EXPECT_EQ(stat.allocCount, stat.getCount);
     EXPECT_EQ(stat.getCount, stat.putCount);
-    EXPECT_EQ(stat.openCount, stat.getCount);
+    // TODO fix it - it does not work in case of IPC cache hit
+    // EXPECT_EQ(stat.openCount, stat.getCount);
     EXPECT_EQ(stat.openCount, stat.closeCount);
 }
 
--- a/test/ipcFixtures.hpp
+++ b/test/ipcFixtures.hpp
@@ -46,22 +46,26 @@ class HostMemoryAccessor : public MemoryAccessor {
     }
 };
 
+// ipcTestParams:
+// pool_ops, pool_params, provider_ops, provider_params, memoryAccessor, free_not_supp
+// free_not_supp (bool) - provider does not support the free() op
 using ipcTestParams =
     std::tuple<umf_memory_pool_ops_t *, void *, umf_memory_provider_ops_t *,
-               void *, MemoryAccessor *>;
+               void *, MemoryAccessor *, bool>;
 
 struct umfIpcTest : umf_test::test,
                     ::testing::WithParamInterface<ipcTestParams> {
     umfIpcTest() {}
     void SetUp() override {
         test::SetUp();
-        auto [pool_ops, pool_params, provider_ops, provider_params, accessor] =
-            this->GetParam();
+        auto [pool_ops, pool_params, provider_ops, provider_params, accessor,
+              free_not_supp] = this->GetParam();
         poolOps = pool_ops;
         poolParams = pool_params;
         providerOps = provider_ops;
         providerParams = provider_params;
         memAccessor = accessor;
+        freeNotSupported = free_not_supp;
     }
 
     void TearDown() override { test::TearDown(); }
@@ -120,7 +124,17 @@ struct umfIpcTest : umf_test::test,
     void *poolParams = nullptr;
     umf_memory_provider_ops_t *providerOps = nullptr;
     void *providerParams = nullptr;
+    bool freeNotSupported = false;
 };
+
+static inline umf_result_t
+get_umf_result_of_free(bool freeNotSupported, umf_result_t expected_result) {
+    if (freeNotSupported) {
+        return UMF_RESULT_ERROR_NOT_SUPPORTED;
+    }
+
+    return expected_result;
+}
 
 TEST_P(umfIpcTest, GetIPCHandleSize) {
     size_t size = 0;
@@ -163,7 +177,8 @@ TEST_P(umfIpcTest, GetIPCHandleInvalidArgs) {
     EXPECT_EQ(ret, UMF_RESULT_ERROR_INVALID_ARGUMENT);
 
     ret = umfFree(ptr);
-    EXPECT_EQ(ret, UMF_RESULT_SUCCESS);
+    EXPECT_EQ(ret,
+              get_umf_result_of_free(freeNotSupported, UMF_RESULT_SUCCESS));
 }
 
 TEST_P(umfIpcTest, BasicFlow) {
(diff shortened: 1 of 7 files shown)


ldorau:
See the issue: https://github.com/oneapi-src/unified-memory-framework/issues/846

test/provider_devdax_memory.cpp
```diff
@@ -14,6 +14,10 @@
 #include "cpp_helpers.hpp"
 #include "test_helpers.h"
 
+#define UMF_TEST_PROVIDER_FREE_NOT_SUPPORTED 1
+#include "ipcFixtures.hpp"
+#undef UMF_TEST_PROVIDER_FREE_NOT_SUPPORTED
```
vinser52, line 19: Instead of a macro, I suggest extending the `ipcTestParams` to contain a flag that tells if the provider supports free operation.

test/ipcFixtures.hpp
```diff
@@ -57,6 +64,21 @@ struct umfIpcTest : umf_test::test,
         test::SetUp();
         auto [pool_ops, pool_params, provider_ops, provider_params, accessor] =
             this->GetParam();
+
+        if (provider_ops == umfDevDaxMemoryProviderOps()) {
```
vinser52, line 68: Why do we need these checks here? In the `provider_devdax_memory.cpp` you can decide whether you would like to instantiate the test or not.

vinser52 (changes requested):

test/ipcFixtures.hpp
ldorau: Because if any of the following **environment variables** is not set: `UMF_TESTS_DEVDAX_PATH` or `UMF_TESTS_DEVDAX_SIZE` then the tests should be skipped.

test/provider_devdax_memory.cpp
ldorau: Done

test/provider_devdax_memory.cpp
```diff
@@ -349,3 +351,15 @@ TEST_F(test, create_wrong_size_0) {
     EXPECT_EQ(ret, UMF_RESULT_ERROR_INVALID_ARGUMENT);
     EXPECT_EQ(hProvider, nullptr);
 }
+
+HostMemoryAccessor hostAccessor;
+
+static std::vector<ipcTestParams> ipcProxyPoolTestParamsList = {
+    {umfProxyPoolOps(), nullptr, umfDevDaxMemoryProviderOps(),
```
vinser52, line 358: Should we also test with other pool implementations, e.g. scalable_pool, jemalloc_pool.

test/provider_os_memory.cpp
```diff
@@ -387,7 +387,7 @@ umf_disjoint_pool_params_t disjointParams = disjointPoolParams();
 static std::vector<ipcTestParams> ipcTestParamsList = {
 #if (defined UMF_POOL_DISJOINT_ENABLED)
     {umfDisjointPoolOps(), &disjointParams, umfOsMemoryProviderOps(),
-     &os_params, &hostAccessor},
+     &os_params, &hostAccessor, false},
```
vinser52, line 390: You should also update tests for all providers (L0, etc.) who use it.

test/ipcFixtures.hpp
vinser52: I understand your intention to check if the `UMF_TESTS_DEVDAX_PATH` and `UMF_TESTS_DEVDAX_SIZE` environments are set. But I do not think we should check for a particular memory provider here.

My suggestion is the following: instead of always instantiating this test in the `provider_devdax_memory.cpp` on the line 364 and do check here we should not instantiate the `umfIpctests` for devdax provider if the required environment variable were not set. Or you can keep the `ipcProxyPoolTestParamsList` list empty if required environment is not set.

vinser52 (changes requested):

test/ipcFixtures.hpp
ldorau: But it would cause multiplication of the code with checking those variables - the same code will have to be placed in each and every provider that uses `ipcFixtures.hpp` - is it really better?

test/provider_os_memory.cpp
ldorau: Done

test/provider_devdax_memory.cpp
ldorau: Done

test/ipcFixtures.hpp
vinser52: > the same code will have to be placed in each and every provider that uses ipcFixtures.hpp

Hmm, I did not get your point. This check is specific for DEVDAX provider.

test/pools/ipc_jemalloc_devdax.cpp
```diff
@@ -0,0 +1,25 @@
...
+                                         : "0"));
+
+static std::vector<ipcTestParams> ipcJemallocPoolTestParamsList = {
+    {umfJemallocPoolOps(), nullptr, umfDevDaxMemoryProviderOps(),
+     &defaultDevDaxParams, &hostAccessor, false},
+};
+
+GTEST_ALLOW_UNINSTANTIATED_PARAMETERIZED_TEST(umfIpcTest);
+
+INSTANTIATE_TEST_SUITE_P(DevDaxProviderProxyPoolTest, umfIpcTest,
```
vinser52, line 24: I think this instantiation of the devdax + jemalloc should be done in the `provider_devdax_memory.cpp`. Something like this:

```c++
static std::vector<ipcTestParams> ipcProxyPoolTestParamsList = {
    {umfProxyPoolOps(), nullptr, umfDevDaxMemoryProviderOps(),
    &defaultDevDaxParams, &hostAccessor, false},
#ifdef UMF_JEMALLOC_ENABLED // I do not remember exact macros
    {umfJemallocPoolOps(), nullptr, umfDevDaxMemoryProviderOps(),
     &defaultDevDaxParams, &hostAccessor, false},
#endif
#ifdef UMF_SCALABLE_POOL_ENABLED // I do not remember exact macros
    {umfScalablePoolOps(), nullptr, umfDevDaxMemoryProviderOps(),
     &defaultDevDaxParams, &hostAccessor, false},
#endif
```

I think it is natural to instantiate tests for a particular provider with all available pools in a single palce

vinser52: you can refer to the #860. How I enabled IPC tests for OS provider with scalable pool

test/provider_devdax_memory.cpp
vinser52: We should enable them here as I mentioned in this comment: https://github.com/oneapi-src/unified-memory-framework/pull/845/files#r1827620658

test/ipcFixtures.hpp
ldorau: Yes, I meant for each and every pool (not provider), but if I use https://github.com/oneapi-src/unified-memory-framework/pull/845#discussion_r1827643900 it will be used only once in one file, so it will be OK.

ldorau: Done, is it OK now?

test/provider_devdax_memory.cpp
ldorau: Done, is it OK now?

vinser52 (approved):

bratpiorka (approved):

bratpiorka:
@ldorau switched to draft - what do you think - should we merge https://github.com/oneapi-src/unified-memory-framework/pull/761 first? Let's discuss this offline

vinser52:
> @ldorau switched to draft - what do you think - should we merge #761 first? Let's discuss this offline

I have two questions/comments:

1. Why we should postpone merging additional tests?
2. Before merging #761 I suggest to agree on design first.

bratpiorka:
ok we can merge this if CI passes

PR #849, opened 2024-10-29
fix for the apply of the HWLOC security patch

bratpiorka:
fix for the apply of the HWLOC security patch

This is a fix for a scenario where the user runs the cmake configuration with the UMF_LINK_HWLOC_STATICALLY setting multiple times for the same build directory. On the first run, cmake downloads the HWLOC sources and successfully applies the required security patch, but the patching fails the second time (the patch has already been applied). This PR fixes this error.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -131,12 +131,6 @@ else()
         set(HWLOC_ENABLE_TESTING OFF)
         set(HWLOC_SKIP_LSTOPO ON)
         set(HWLOC_SKIP_TOOLS ON)
-        set(HWLOC_PATCH
-            git
-            apply
-            ${PROJECT_SOURCE_DIR}/cmake/fix_coverity_issues.patch
-            ||
-            (exit 0))
 
         message(
             STATUS
@@ -147,8 +141,7 @@ else()
             hwloc_targ
             GIT_REPOSITORY ${UMF_HWLOC_REPO}
             GIT_TAG ${UMF_HWLOC_TAG}
-            PATCH_COMMAND ${HWLOC_PATCH} SOURCE_SUBDIR contrib/windows-cmake/
-                          FIND_PACKAGE_ARGS)
+            SOURCE_SUBDIR contrib/windows-cmake/ FIND_PACKAGE_ARGS)
         FetchContent_MakeAvailable(hwloc_targ)
 
         set(LIBHWLOC_INCLUDE_DIRS
@@ -157,13 +150,6 @@ else()
             ${hwloc_targ_BINARY_DIR}/Release;${hwloc_targ_BINARY_DIR}/Debug)
     else()
         include(FetchContent)
-        set(HWLOC_PATCH
-            git
-            apply
-            ${PROJECT_SOURCE_DIR}/cmake/fix_coverity_issues.patch
-            ||
-            (exit 0))
-
         message(
             STATUS
                 "Will fetch hwloc from ${UMF_HWLOC_REPO} (tag: ${UMF_HWLOC_TAG})"
@@ -172,8 +158,7 @@ else()
         FetchContent_Declare(
             hwloc_targ
             GIT_REPOSITORY ${UMF_HWLOC_REPO}
-            GIT_TAG ${UMF_HWLOC_TAG}
-            PATCH_COMMAND ${HWLOC_PATCH})
+            GIT_TAG ${UMF_HWLOC_TAG})
         FetchContent_MakeAvailable(hwloc_targ)
 
         add_custom_command(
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -217,6 +202,22 @@ else()
     message(STATUS "    LIBHWLOC_LIBRARY_DIRS = ${LIBHWLOC_LIBRARY_DIRS}")
 endif()
 
+if(hwloc_targ_SOURCE_DIR)
+    # apply security patch for HWLOC
+    execute_process(
```
KFilipek, line 207: Look at the options named `ECHO_OUTPUT_VARIABLE, ECHO_ERROR_VARIABLE`.
Generally it's not blocking, but probably you can remove all if blocks below.
Source: https://cmake.org/cmake/help/latest/command/execute_process.html

KFilipek (approved):
LGTM

CMakeLists.txt
bratpiorka: ok but this is enabled in cmake >= 3.18 and I do not want to bump min required ver right now

CMakeLists.txt
```diff
@@ -217,6 +202,22 @@ else()
...
+        COMMAND git apply ${PROJECT_SOURCE_DIR}/cmake/fix_coverity_issues.patch
+        WORKING_DIRECTORY ${hwloc_targ_SOURCE_DIR}
+        OUTPUT_VARIABLE UMF_HWLOC_PATCH_OUTPUT
+        ERROR_VARIABLE UMF_HWLOC_PATCH_ERROR)
+
+    if(UMF_HWLOC_PATCH_OUTPUT)
+        message(STATUS "HWLOC patch:\n${UMF_HWLOC_PATCH_OUTPUT}")
+    endif()
+    if(UMF_HWLOC_PATCH_ERROR)
+        message(WARNING "HWLOC patch:\n${UMF_HWLOC_PATCH_ERROR}")
```
PatKamin, line 217: Shouldn't this stop the build?

bratpiorka: no - in case we run cmake config twice, the second time we got this message: 

```
CMake Warning at CMakeLists.txt:217 (message):
  HWLOC patch command output:

  error: patch failed: hwloc/topology-x86.c:1375

  error: hwloc/topology-x86.c: patch does not apply
```

it means that the command failed because the patch is already applied which is OK

CMakeLists.txt
```diff
@@ -217,6 +202,22 @@ else()
...
+if(hwloc_targ_SOURCE_DIR)
+    # apply security patch for HWLOC
+    execute_process(
+        COMMAND git apply ${PROJECT_SOURCE_DIR}/cmake/fix_coverity_issues.patch
+        WORKING_DIRECTORY ${hwloc_targ_SOURCE_DIR}
+        OUTPUT_VARIABLE UMF_HWLOC_PATCH_OUTPUT
+        ERROR_VARIABLE UMF_HWLOC_PATCH_ERROR)
+
+    if(UMF_HWLOC_PATCH_OUTPUT)
+        message(STATUS "HWLOC patch:\n${UMF_HWLOC_PATCH_OUTPUT}")
```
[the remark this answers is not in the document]
bratpiorka: done

CMakeLists.txt
```diff
@@ -217,6 +202,22 @@ else()
...
+        COMMAND git apply ${PROJECT_SOURCE_DIR}/cmake/fix_coverity_issues.patch
+        WORKING_DIRECTORY ${hwloc_targ_SOURCE_DIR}
+        OUTPUT_VARIABLE UMF_HWLOC_PATCH_OUTPUT
+        ERROR_VARIABLE UMF_HWLOC_PATCH_ERROR)
+
+    if(UMF_HWLOC_PATCH_OUTPUT)
+        message(STATUS "HWLOC patch command output:\n${UMF_HWLOC_PATCH_OUTPUT}")
+    endif()
+    if(UMF_HWLOC_PATCH_ERROR)
+        message(WARNING "HWLOC patch command output:\n${UMF_HWLOC_PATCH_ERROR}")
```
PatKamin, line 217: This output looks like something went wrong, although the patch does not apply because it has been already applied:
```
CMake Warning at CMakeLists.txt:217 (message):
  HWLOC patch:

  error: patch failed: hwloc/topology-x86.c:1375

  error: hwloc/topology-x86.c: patch does not apply
```
The build continues, as expected, though.

PatKamin: Perhaps add `--allow-empty` to the `git apply` command?

PatKamin (approved):

PR #854, opened 2024-10-31
fix for the apply of the HWLOC security patch

bratpiorka:
fix for the apply of the HWLOC security patch

This is a fix for a scenario where the user runs the cmake configuration with the UMF_LINK_HWLOC_STATICALLY setting multiple times for the same build directory. On the first run, cmake downloads the HWLOC sources and successfully applies the required security patch, but the patching fails the second time (the patch has already been applied). This PR fixes this error.

NOTE: this is a v0.9x ver - previously reviewed here: https://github.com/oneapi-src/unified-memory-framework/pull/849


lukaszstolarczuk:
to be rebased after https://github.com/oneapi-src/unified-memory-framework/pull/857

lukaszstolarczuk (approved):

PR #858, opened 2024-10-31
Add tests for IPC not supported case

staniewzki:
### Checklist

- [x] All tests pass locally

### DIFF
--- a/test/ipc_negative.cpp
+++ b/test/ipc_negative.cpp
@@ -1,0 +1,53 @@
+// Copyright (C) 2024 Intel Corporation
+// Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "base.hpp"
+#include "pool_null.h"
+#include "provider_null.h"
+
+#include <umf/ipc.h>
+#include <umf/memory_pool.h>
+#include <umf/memory_provider.h>
+
+#include <array>
+
+struct IpcNotSupported : umf_test::test {
+  protected:
+    void SetUp() override {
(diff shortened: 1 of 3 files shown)


test/ipc_negative.cpp
```diff
@@ -0,0 +1,49 @@
...
+    umf_ipc_handle_t ipc_handle = {};
+    void *ptr;
+    auto ret = umfOpenIPCHandle(pool, ipc_handle, &ptr);
+    EXPECT_EQ(ret, UMF_RESULT_ERROR_NOT_SUPPORTED);
+}
+
+TEST_F(IpcNotSupported, CloseIPCHandleInvalid) {
+    int local_var;
+    auto ret = umfCloseIPCHandle(&local_var);
+    EXPECT_EQ(ret, UMF_RESULT_ERROR_INVALID_ARGUMENT);
```
vinser52, line 48: In fact, if you pass some pointer that was not result of the `umfOpenIPCHandle` function then you get `UMF_RESULT_ERROR_INVALID_ARGUMENT` error. So with this test you are not checking if UMF tries to call `provider_ops.ipc.close_ipc_handle` callback, the UMf fails earlier because the argument is invalid.

staniewzki: Yes, that was the idea - that case was missing for coverage. I see how it wasn't clear there so I have renamed test and added a comment.

vinser52: in `ipcFixture.cpp` we already have `TEST_P(umfIpcTest, GetIPCHandleSizeInvalidArgs)` and `TEST_P(umfIpcTest, GetIPCHandleInvalidArgs)` to test invalid args with other IPC APIs. Maybe we should move this particular test there?

vinser52 (approved):

bratpiorka (approved):

vinser52 (approved):

Issue #848, opened 2024-10-29
IPC API of the file provider does not work


ldorau:
IPC API of the file provider does not work.

Ref: #847

## Environment Information

- UMF version (hash commit or a tag): b3bbdd498a8a3e2b55b1a669e0a8fc090bbbd0f1
- OS(es) version(s): Linux

## Please provide a reproduction of the bug:

See: #847 
See: https://github.com/oneapi-src/unified-memory-framework/actions/runs/11572419084/job/32212568727?pr=847

## How often bug is revealed:

always

ldorau:
@vinser52

ldorau:
@bratpiorka

ldorau:
@vinser52 @bratpiorka @lplewa

PR #838, opened 2024-10-24
Fix devdax_open_ipc_handle() and devdax_close_ipc_handle()

ldorau:
### Description

Fix devdax_open_ipc_handle() and devdax_close_ipc_handle()

devdax_open_ipc_handle() has to use the path of the remote
/dev/dax got from the IPC handle, not the local one.

devdax_open_ipc_handle() has to use also the memory protection
got from the IPC handle, so let's add the memory protection
to the IPC handle.

devdax_open_ipc_handle() should mmap only the required range of memory,
not the whole /dev/dax device, so let's add the length of the allocation
to the IPC handle.

devdax_close_ipc_handle() should unmap only the required range of memory,
not the whole /dev/dax device.

Fixes: #846

### Checklist

- [x] Code compiles without errors locally
- [x] All tests pass locally
- [x] CI workflows execute properly

PR #865, opened 2024-11-05
Fix `file_alloc()` and `file_open/close_ipc_handle()`

ldorau:
### Description

Alignment passed to `file_alloc()` must be a power of two and a multiple or a divider of the page size.
Align up the alignment in `file_alloc()` to the page size.

`file_open_ipc_handle()` has to use the memory protection and visibility got from the IPC handle, so let's add the memory protection and visibility to the IPC handle.

Ref: #838

Fixes: #848

### Checklist
- [x] All tests pass locally

### DIFF
--- a/src/provider/provider_file_memory.c
+++ b/src/provider/provider_file_memory.c
@@ -397,16 +397,21 @@ static umf_result_t file_alloc(void *provider, size_t size, size_t alignment,
         return UMF_RESULT_ERROR_INVALID_ARGUMENT;
     }
 
-    // alignment must be a power of two and a multiple of sizeof(void *)
-    if (alignment &&
-        ((alignment & (alignment - 1)) || (alignment % sizeof(void *)))) {
-        LOG_ERR("wrong alignment: %zu (not a power of 2 or a multiple of "
-                "sizeof(void *))",
-                alignment);
+    file_memory_provider_t *file_provider = (file_memory_provider_t *)provider;
+
+    // alignment must be a power of two and a multiple or a divider of the page size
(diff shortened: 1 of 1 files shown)


kswiecicki (approved):

src/provider/provider_file_memory.c
```diff
@@ -397,16 +397,21 @@ static umf_result_t file_alloc(void *provider, size_t size, size_t alignment,
...
-        ((alignment & (alignment - 1)) || (alignment % sizeof(void *)))) {
-        LOG_ERR("wrong alignment: %zu (not a power of 2 or a multiple of "
-                "sizeof(void *))",
-                alignment);
+    file_memory_provider_t *file_provider = (file_memory_provider_t *)provider;
+
+    // alignment must be a power of two and a multiple or a divider of the page size
+    if (alignment && ((alignment & (alignment - 1)) ||
+                      ((alignment % file_provider->page_size) &&
+                       (file_provider->page_size % alignment)))) {
```
vinser52, line 405: Just to make sure that I understand it right, with file provider we do not have the issue like with the devdax provider because the page size is 4KB in case of file provider. Right?

ldorau: Do you mean issue that size is not a multiplier of the page size?

ldorau: Yes, I updated the code.

vinser52: > Do you mean issue that size is not a multiplier of the page size?

yes

vinser52 (approved):

PR #866, opened 2024-11-05
[main] Fix disabling of pci support in hwloc

PatKamin:
The flag that hwloc recognize is --disable-pci, not --disable-pciaccess.


bratpiorka:
could you write more why we disable this? Maybe we should review the full list of flags and disable more?

PatKamin:
> could you write more why we disable this? Maybe we should review the full list of flags and disable more?

Minimized the changes to current flag fix, with explanation.

bratpiorka (approved):

Issue #864, opened 2024-11-05
IPC API does not work with the proxy library


ldorau:
IPC API does not work with the proxy library.
IPC tests: `umf-provider_devdax_memory_ipc` and `umf-provider_file_memory_ipc` fail with the proxy library.

The reason: issues with the tracker (ref: #761).

See: https://github.com/oneapi-src/unified-memory-framework/actions/runs/11680149618/job/32522557687
See: https://github.com/ldorau/unified-memory-framework/actions/runs/11681287895/job/32526033443

Ref: #863 (commented out IPC test `umf-provider_devdax_memory_ipc`)
Ref: #847 (commented out IPC test `umf-provider_file_memory_ipc`)

Ref: #868 
See: https://github.com/oneapi-src/unified-memory-framework/actions/runs/11683497460

## Environment Information

- UMF version (hash commit or a tag): cdc031001999644771107e6a7f80f0860559636c
- OS(es) version(s): Linux

## Please provide a reproduction of the bug:

Run the `umf-provider_devdax_memory_ipc` with the proxy library:
See: https://github.com/oneapi-src/unified-memory-framework/actions/runs/11680149618/job/32522557687

or

run the `umf-provider_file_memory_ipc` with the proxy library:
See: https://github.com/ldorau/unified-memory-framework/actions/runs/11681287895/job/32526033443

## How often bug is revealed:

always

ldorau:
@vinser52 @bratpiorka @lplewa

PR #868, opened 2024-11-05
Enable IPC tests with the proxy library

ldorau:
### Description

Enable IPC tests with the proxy library which were disabled, because of the issue: https://github.com/oneapi-src/unified-memory-framework/issues/864

Fixes: #864

Requires:
- [x] #931 
- [x] #932 

### Checklist

- [x] All tests pass locally

### DIFF
--- a/.github/workflows/reusable_proxy_lib.yml
+++ b/.github/workflows/reusable_proxy_lib.yml
@@ -59,11 +59,9 @@ jobs:
       - name: Build UMF
         run: cmake --build ${{env.BUILD_DIR}} -j $(nproc)
 
-      # TODO enable the provider_file_memory_ipc test when the IPC tests with the proxy library are fixed
-      # see the issue: https://github.com/oneapi-src/unified-memory-framework/issues/864
       - name: Run "ctest --output-on-failure" with proxy library
         working-directory: ${{env.BUILD_DIR}}
-        run: LD_PRELOAD=./lib/libumf_proxy.so ctest --output-on-failure -E provider_file_memory_ipc
+        run: LD_PRELOAD=./lib/libumf_proxy.so ctest --output-on-failure
 
       - name: Run "./test/umf_test-memoryPool" with proxy library
         working-directory: ${{env.BUILD_DIR}}
@@ -77,14 +75,12 @@ jobs:
         working-directory: ${{env.BUILD_DIR}}
         run: UMF_PROXY="page.disposition=shared-shm" LD_PRELOAD=./lib/libumf_proxy.so /usr/bin/date
 
-      # TODO enable the provider_file_memory_ipc test when the IPC tests with the proxy library are fixed
-      # see the issue: https://github.com/oneapi-src/unified-memory-framework/issues/864
       - name: Run "ctest --output-on-failure" with proxy library and size.threshold=128
         working-directory: ${{env.BUILD_DIR}}
         run: >
           UMF_PROXY="page.disposition=shared-shm;size.threshold=128"
           LD_PRELOAD=./lib/libumf_proxy.so
-          ctest --output-on-failure -E provider_file_memory_ipc
+          ctest --output-on-failure
 
       - name: Check coverage
         if:  ${{ matrix.build_type == 'Debug' }}
--- a/.github/workflows/reusable_dax.yml
+++ b/.github/workflows/reusable_dax.yml
@@ -106,8 +106,6 @@ jobs:
           UMF_TESTS_FSDAX_PATH_2=${{env.UMF_TESTS_FSDAX_PATH_2}}
           ctest -C ${{matrix.build_type}} -V -R "file|fsdax"
 
-      # TODO: enable the provider_devdax_memory_ipc test when the IPC tests with the proxy library are fixed
-      # see the issue: https://github.com/oneapi-src/unified-memory-framework/issues/864
       - name: Run the DEVDAX tests with the proxy library
         # proxy library is built only if libumf is a shared library
         if: ${{ matrix.shared_library == 'ON' }}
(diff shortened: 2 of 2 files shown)


ldorau:
@vinser52 please review

ldorau:
It fixes: https://github.com/oneapi-src/unified-memory-framework/issues/864

src/provider/provider_tracking.c
```diff
@@ -382,6 +382,12 @@ static umf_result_t trackingFree(void *hProvider, void *ptr, size_t size) {
     ret = umfMemoryProviderFree(p->hUpstream, ptr, size);
     if (ret != UMF_RESULT_SUCCESS) {
         LOG_ERR("upstream provider failed to free the memory");
+        // Do not add memory back to the tracker,
+        // if the provider does not support the free() op.
+        if (ret == UMF_RESULT_ERROR_NOT_SUPPORTED) {
```
vinser52, line 387: I agree that we should not add memory back to the tracker in case of memory provider does not support `free`.

But why is pool implementation tries to call `umfMemoryProviderFree()` at all?

ldorau: Do you mean `umfMemoryProviderFree()` above ? - inside `trackingFree()` in line no. 382 ?

vinser52: With such fix we get a memory leak.

Imagine the pool calls `umfMemoryProviderFree()` for the address range `[a. b)`. After that pool calls `umfMemoryProviderAlloc()` my understanding is that memory provider will never return the range `[a. b)` as a result of malloc, but it will return the next chunk `[b,c)`. If the pool will call `umfMemoryProviderFree()` for the `[b,c)` range it also will be lost and never be used again.

ldorau: Yes, the file provider and the devdax provider work exactly as as you wrote - they were designed to work in this way.

src/provider/provider_tracking.c
```diff
@@ -382,6 +382,12 @@ static umf_result_t trackingFree(void *hProvider, void *ptr, size_t size) {
     ret = umfMemoryProviderFree(p->hUpstream, ptr, size);
     if (ret != UMF_RESULT_SUCCESS) {
         LOG_ERR("upstream provider failed to free the memory");
```
bratpiorka, line 384: in case the free() is not supported we should call LOG_ERR. Please do not nest ifs here and use switch or if/else if

src/provider/provider_tracking.c
bratpiorka: ... but we should not just return "success" here because the memory is actually not released. Maybe this fix is wrong... We should fix only the case when the pool that uses devdax/file providers is destroyed. Only then we should remove the entries from the tracker as we also unmap the memory. 

There is also an additional way fix this bug - free() in devdax/file could do the unmap but we could keep the offset unchanged - this way the free() would be supported and entries would be correctly removed from the tracker while we still keep file/devdax provider logic simple (so the memory would be still unavailable after free).

vinser52: > There is also an additional way fix this bug - free() in devdax/file could do the unmap but we could keep the offset unchanged - this way the free() would be supported and entries would be correctly removed from the tracker while we still keep file/devdax provider logic simple (so the memory would be still unavailable after free).

I agree that at least we are not wasting VA space (but it is not a problem), but since we never decrement the offset in devdax and file providers by design we still have a kind of memory leak. Because if the `umfMemoryProviderAlloc` and `umfMemoryProviderFree` are called multiple times, after some time `umfMemoryProviderAlloc` will return OOM because offset reached the end of file/devdax.


We discussed this with @ldorau offline. My main comment that currently this PR eliminates the issue:
 - today we have a memory leak if `umfMemoryProviderFree` is called. If we do not remove the range from the tracker our diagnostic in DEBUG mode will detect it and report.
 - with this PR the memory leak still there but our diagnostic is not able to detect it and report.

vinser52: > I agree with @ldorau - this is how these providers work. And because of this, in the real-life scenarios they should be used either with a pool that never calls free() or with CoarseProvider

Just to clarify, I am not challenging how devdax or file providers work. we agreed on that. My concern is about this particular flow when some pool calls `umfmemoryProviderFree` for devdax or file providers. This situation should not happens or should be handled correctly. Today looks like there is a memory leak in that case and this PR does not fix the leak.

ldorau: Removed (code change). It will be fixed in another way.
Done.

src/provider/provider_tracking.c
ldorau: Removed (code change). It will be fixed in another way.
Done.

src/provider/provider_tracking.c
lplewa: I was saying that providers without free should have a null ptr to function, and we should have an API that pool can check if free function is supported if not - if it's not supported it should never call providerFree(). 

Or as another option - pool informs umf if it can support provider without free - if pool cannot handle it, we always wrap it with coarse provider.

ldorau: Code changed.
Done.

src/provider/provider_tracking.c
ldorau: Code changed.
Done.

ldorau:
Requires:
- [ ] #931 
- [ ] #932

ldorau:
@bratpiorka @vinser52 please review

ldorau:
This PR changes only 2 workflow files (+3/-9 lines) in order to enable IPC tests that have not been passing so far.

ldorau:
@lukaszstolarczuk please review

.github/workflows/reusable_proxy_lib.yml
```diff
@@ -59,11 +59,9 @@ jobs:
       - name: Build UMF
         run: cmake --build ${{env.BUILD_DIR}} -j $(nproc)
 
-      # TODO enable the provider_file_memory_ipc test when the IPC tests with the proxy library are fixed
-      # see the issue: https://github.com/oneapi-src/unified-memory-framework/issues/864
       - name: Run "ctest --output-on-failure" with proxy library
         working-directory: ${{env.BUILD_DIR}}
-        run: LD_PRELOAD=./lib/libumf_proxy.so ctest --output-on-failure -E provider_file_memory_ipc
+        run: LD_PRELOAD=./lib/libumf_proxy.so ctest --output-on-failure
```
lukaszstolarczuk, line 65: there's another one down below, pointing to the same issue

lukaszstolarczuk (changes requested):

.github/workflows/reusable_proxy_lib.yml
ldorau: @lukaszstolarczuk Fixed.
Done.

lukaszstolarczuk (approved):

bratpiorka (approved):

PR #871, opened 2024-11-05
Workflows clean up (part 2)

lukaszstolarczuk:
Each commit should be checked separately.

QEMU build is split, to only run selected jobs per PR and the whole suite (with extra OS) on nightly. QEMU in short run executes only 3 configs and valgrind is executed only on `default`.
// nightly qemu (long) run was executed on my fork: https://github.com/lukaszstolarczuk/unified-memory-framework/actions/runs/11722236706/job/32651339247

CUDA and L0 - similarly - only a few jobs on PR, the rest during the nigth.

Multi Numa workflow is extended, because it's executed very quickly.

### Checklist

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -175,3 +175,12 @@ jobs:
         call "C:\Program Files (x86)\Intel\oneAPI\setvars.bat"
         call "C:\Program Files (x86)\Intel\oneAPI\setvars-vcvarsall.bat"
         ctest -C ${{matrix.build_type}} --output-on-failure --test-dir test
+
+  L0:
+    uses: ./.github/workflows/reusable_gpu.yml
+    with:
+      name: "LEVEL_ZERO"
+  CUDA:
+    uses: ./.github/workflows/reusable_gpu.yml
+    with:
+      name: "CUDA"
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -184,3 +184,10 @@ jobs:
     uses: ./.github/workflows/reusable_gpu.yml
     with:
       name: "CUDA"
+
+  # Full exeuction of QEMU tests
+  QEMU:
+    uses: ./.github/workflows/reusable_qemu.yml
+    with:
+      short_run: false
+      os: "['ubuntu-23.04', 'ubuntu-24.04']"
--- a/.github/workflows/reusable_gpu.yml
+++ b/.github/workflows/reusable_gpu.yml
@@ -1,10 +1,26 @@
-# This workflow builds and tests providers using GPU memory. It requires 
-# appropriately labelled self-hosted runners installed on systems with the 
-# correct GPU and drivers
-
+# This workflow builds and tests providers using GPU memory. It requires properly
+# labelled self-hosted runners on systems with the correct GPU and drivers.
 name: GPU
 
-on: [workflow_call]
+on:
+  workflow_call:
+    inputs:
+      name:
+        description: Provider name
+        type: string
+        required: true
+      os:
+        description: A list of OSes
+        type: string
+        default: "['Ubuntu', 'Windows']"
+      build_type:
+        description: A list of build types
+        type: string
+        default: "['Debug', 'Release']"
+      shared_lib:
+        description: A list of options for building shared library
+        type: string
+        default: "['ON', 'OFF']"
 
 permissions:
   contents: read
@@ -15,18 +31,20 @@ env:
   COVERAGE_DIR : "${{github.workspace}}/coverage"
 
 jobs:
-  gpu-Level-Zero:
-    name: Level-Zero
+  gpu:
+    name: "${{matrix.os}}, ${{matrix.build_type}}, shared=${{matrix.shared_library}}"
     env:
-      VCPKG_PATH: "${{github.workspace}}/../../../../vcpkg/packages/hwloc_x64-windows;${{github.workspace}}/../../../../vcpkg/packages/tbb_x64-windows;${{github.workspace}}/../../../../vcpkg/packages/jemalloc_x64-windows"
-      COVERAGE_NAME : "exports-coverage-gpu-L0"
+      VCPKG_PATH: "${{github.workspace}}/build/vcpkg/packages/hwloc_x64-windows;${{github.workspace}}/build/vcpkg/packages/tbb_x64-windows;${{github.workspace}}/build/vcpkg/packages/jemalloc_x64-windows;"
+      CUDA_PATH: "C:/cuda"
+      COVERAGE_NAME : "exports-coverage-${{inputs.name}}"
     # run only on upstream; forks will not have the HW
     if: github.repository == 'oneapi-src/unified-memory-framework'
     strategy:
+      fail-fast: false
       matrix:
-        shared_library: ['ON', 'OFF']
-        os: ['Ubuntu', 'Windows']
-        build_type: ['Debug', 'Release']
+        shared_library: ${{ fromJSON(inputs.shared_lib)}}
+        os: ${{ fromJSON(inputs.os)}}
+        build_type: ${{ fromJSON(inputs.build_type)}}
         include:
           - os: 'Ubuntu'
             compiler: {c: gcc, cxx: g++}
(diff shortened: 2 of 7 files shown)


scripts/qemu/run-tests.sh
```diff
@@ -26,19 +29,22 @@ echo password | sudo bash -c "echo 0 > /proc/sys/kernel/yama/ptrace_scope"
 numactl -H
 
 cd build
+echo "## Running all tests ..."
 ctest --verbose
 
-# run tests bound to a numa node
+echo "## Running tests bound to a numa node 0 & 1 ..."
```
PatKamin, line 35: & -> and

.github/workflows/reusable_qemu.yml
```diff
@@ -1,14 +1,29 @@
...
+  workflow_call:
+    inputs:
+      short_run:
+        description: Should the workflow run only basic tests?
+        type: boolean
+        default: false
+      os:
+        description: List of OSes
+        type: string
+        default: '["ubuntu-23.04"]'
```
PatKamin, line 14: Why are we testing non-LTS OS here?

.github/workflows/reusable_gpu.yml
```diff
@@ -4,7 +4,25 @@
...
+    inputs:
+      name:
+        description: Provider name
+        type: string
+        required: true
+      os:
+        description: A list of OSes
+        type: string
+        required: true
+      build_type:
```
PatKamin, line 18: I find this input unused by other workflows. Perhaps I've omitted some usage?

.github/workflows/reusable_gpu.yml
```diff
@@ -15,30 +33,30 @@ env:
...
             compiler: {c: gcc, cxx: g++}
             number_of_processors: '$(nproc)'
-          - os: 'Windows'
-            compiler: {c: cl, cxx: cl}
-            number_of_processors: '$Env:NUMBER_OF_PROCESSORS'
-        exclude:
-          - os: 'Windows'
-            build_type: 'Debug'
+          # FIXME:
+          # Temporarily disable including Windows options until we have CUDA enabled.
```
PatKamin, line 54: I don't know why these options matter to CUDA. You set them as hardcoded anyway...

.github/workflows/reusable_gpu.yml
```diff
@@ -64,39 +82,41 @@ jobs:
...
           -DUMF_BUILD_GPU_TESTS=ON
           -DUMF_BUILD_GPU_EXAMPLES=ON
-          -DUMF_FORMAT_CODE_STYLE=OFF
           -DUMF_DEVELOPER_MODE=ON
           -DUMF_BUILD_LIBUMF_POOL_DISJOINT=ON
           -DUMF_BUILD_LIBUMF_POOL_JEMALLOC=ON
-          -DUMF_BUILD_LEVEL_ZERO_PROVIDER=ON
           -DUMF_BUILD_CUDA_PROVIDER=OFF
+          -DUMF_BUILD_LEVEL_ZERO_PROVIDER=OFF
+          -DUMF_BUILD_${{inputs.name}}_PROVIDER=ON
```
PatKamin, line 90: Does it work properly 100% of time? Setting the same option to OFF and to ON in the same config command?

.github/workflows/reusable_gpu.yml
```diff
@@ -64,39 +82,41 @@ jobs:
...
           -DUMF_TESTS_FAIL_ON_SKIP=ON
-          ${{ matrix.build_type == 'Debug' && '-DUMF_USE_COVERAGE=ON' || '' }}
 
       - name: Build UMF
-        run: cmake --build ${{env.BUILD_DIR}} --config ${{matrix.build_type}} -j ${{matrix.number_of_processors}}
+        run: cmake --build ${{env.BUILD_DIR}} --config ${{matrix.build_type}} -j8
```
PatKamin, line 119: Why using a magic number of cpus instead of getting available cpus count from os?

.github/workflows/reusable_qemu.yml
```diff
@@ -79,11 +94,26 @@ jobs:
...
+      run: |
+        echo "SHORT_RUN=true" >> $GITHUB_ENV
+        declare -a short_configs=("default.xml" "sock_2_var3.xml" "sock_4_var1_hmat.xml")
+        echo "CONFIG_OPTIONS=${short_configs[@]}" >> $GITHUB_ENV
+
+    - name: Set vars if long run
+      if: ${{ inputs.short_run == false }}
+      run: |
+        echo "SHORT_RUN=false" >> $GITHUB_ENV
+        echo "CONFIG_OPTIONS=umf/scripts/qemu/configs/*.xml" >> $GITHUB_ENV
```
PatKamin, line 108: Why this path differs from the paths passed in the `SHORT_RUN=true` case?

.github/workflows/reusable_gpu.yml
lukaszstolarczuk: you're right - at this moment it's used only its default value; I was using this to skip some troublesome builds, like exclude Debug from L0, which I found useful and left here for others to use, if needed

.github/workflows/reusable_qemu.yml
lukaszstolarczuk: good question, I didn't change that, we used this before and I think as long as it works I would keep it; we can also extend/change the list of OSes if needed (in another PR)

scripts/qemu/run-tests.sh
lukaszstolarczuk: done

.github/workflows/reusable_gpu.yml
ldorau: But why the CUDA Win build is required to use `${{matrix.number_of_processors}}` ? I think we can use it without the Win build too.

.github/workflows/reusable_qemu.yml
```diff
@@ -98,24 +128,26 @@ jobs:
...
             echo "Waiting for QEMU to shut down..."
             sleep 5
           done
 
-          echo "\n ### Testing ${config_name} ###"
+          echo "### Testing ${config_name} ###"
           umf/scripts/qemu/start_qemu.sh ${config_name}
 
-          ssh testuser@127.0.0.1 -p 2222 -t "bash /home/testuser/scripts/qemu/run-tests.sh COVERAGE ${config_name}"
+          ssh testuser@127.0.0.1 -p 2222 -t "export SHORT_RUN=${SHORT_RUN} OS_FULL_NAME=${{matrix.os}} && /home/testuser/scripts/qemu/run-tests.sh COVERAGE ${config_name}"
```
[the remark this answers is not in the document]
lukaszstolarczuk: it should, but it's executed via ssh and it caused me an error like `not recognized command SHORT_RUN`

.github/workflows/reusable_gpu.yml
lukaszstolarczuk: Win CUDA build is already on PR, should be merged pretty soon, so my temp. workaround will be removed soon.

We could use it and not use my `include` way, but it would require some copy-pasting of the commands for Windows and Linux, for now -j8 should be fine.

.github/workflows/reusable_qemu.yml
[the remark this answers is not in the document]
ldorau: OK

.github/workflows/nightly.yml
```diff
@@ -175,3 +175,21 @@ jobs:
...
+  L0:
+    uses: ./.github/workflows/reusable_gpu.yml
+    with:
+      name: "LEVEL_ZERO"
+      os: "['Ubuntu', 'Windows']"
+  CUDA:
+    uses: ./.github/workflows/reusable_gpu.yml
+    with:
+      name: "CUDA"
+      os: "['Ubuntu']"
```
bratpiorka, line 188: missing windows?

bratpiorka (approved):

.github/workflows/nightly.yml
lukaszstolarczuk: missed that, thx!

.github/workflows/reusable_qemu.yml
PatKamin: That should be dropped in some other PR, I think.

.github/workflows/reusable_qemu.yml
PatKamin: Hadn't noticed this `basename` before, thanks.

PR #872, opened 2024-11-05
Fix Level Zero related CMake configs

vinser52:
### Description
There are two L0-related fixes in this PR:

1. When `UMF_LEVEL_ZERO_INCLUDE_DIR` is set the test cannot be built because of `ze_api.h` not found. It happens because we set `LEVEL_ZERO_INCLUDE_DIRS` in the `src/CMakeLists.txt`. The fix moves corresponding CMake code to the `./CmakeLists.txt`. @lukaszstolarczuk, @bratpiorka, looks like we never checked such flow in CI, can we do that?
2. Some versions of the L0 driver do not return an error when a user tries to allocate 0 bytes. So I removed the corresponding test.

### Checklist

- [x] All tests pass locally

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -218,6 +218,64 @@ if(hwloc_targ_SOURCE_DIR)
     endif()
 endif()
 
+# Fetch L0 loader only if needed i.e.: if building L0 provider is ON and L0
+# headers are not provided by the user (via setting UMF_LEVEL_ZERO_INCLUDE_DIR).
+if(UMF_BUILD_LEVEL_ZERO_PROVIDER AND (NOT UMF_LEVEL_ZERO_INCLUDE_DIR))
+    include(FetchContent)
+
+    set(LEVEL_ZERO_LOADER_REPO "https://github.com/oneapi-src/level-zero.git")
+    set(LEVEL_ZERO_LOADER_TAG v1.17.39)
+
+    message(
+        STATUS
+            "Fetching L0 loader (${LEVEL_ZERO_LOADER_TAG}) from ${LEVEL_ZERO_LOADER_REPO} ..."
+    )
+
+    FetchContent_Declare(
+        level-zero-loader
+        GIT_REPOSITORY ${LEVEL_ZERO_LOADER_REPO}
+        GIT_TAG ${LEVEL_ZERO_LOADER_TAG}
+        EXCLUDE_FROM_ALL)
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -218,6 +218,64 @@ if(hwloc_targ_SOURCE_DIR)
...
+        EXCLUDE_FROM_ALL)
+    FetchContent_MakeAvailable(level-zero-loader)
+
+    set(LEVEL_ZERO_INCLUDE_DIRS
+        ${level-zero-loader_SOURCE_DIR}/include
+        CACHE PATH "Path to Level Zero Headers")
+    message(STATUS "Level Zero include directory: ${LEVEL_ZERO_INCLUDE_DIRS}")
+elseif(UMF_BUILD_LEVEL_ZERO_PROVIDER)
+    # Only header is needed to build UMF
+    set(LEVEL_ZERO_INCLUDE_DIRS ${UMF_LEVEL_ZERO_INCLUDE_DIR})
```
lukaszstolarczuk, line 247: as for point 1) in your description:
- in this `elseif` the `LEVEL_ZERO_INCLUDE_DIRS` variable is not set as `CACHE`, so it was not propagated properly in the entire project

your solution should be ok as well.

We tested (and still test) the `UMF_LEVEL_ZERO_INCLUDE_DIR` setup in UR workflow (but we don't run there our tests).

lukaszstolarczuk (approved):

CMakeLists.txt
vinser52: > as for point 1) in your description:
> 
> - in this elseif the LEVEL_ZERO_INCLUDE_DIRS variable is not set as CACHE, so it was not propagated properly in the entire project

But the issue happens when I set the `UMF_LEVEL_ZERO_INCLUDE_DIR` so it is `if` block. Why the `elseif` block matters?

> We tested (and still test) the UMF_LEVEL_ZERO_INCLUDE_DIR setup in UR workflow (but we don't run there our tests).

Yeah, I know that. But UMF is a standalone project if we have such an option it should be tested on our side.

lukaszstolarczuk: och, right, I misread that... then... I'm not sure what happened 😄 
<br />

> Yeah, I know that. But UMF is a standalone project if we have such an option it should be tested on our side.

I guess, you're right, we could test that with an example's CMake... I'll add an issue for that.

PR #882, opened 2024-11-08
Add a RUNPATH to installed libumf.so

PatKamin:


CMakeLists.txt
```diff
@@ -312,6 +312,8 @@ endif()
 # dependencies.
 add_library(umf_headers INTERFACE)
 
+set(CMAKE_INSTALL_RPATH "\$ORIGIN")
```
lukaszstolarczuk, line 315: I think it requires at least a small-ish comment here 😉

bratpiorka, line 315: please move this to "install(FILES .. " with other installation-related code

PatKamin: Added a comment

PatKamin: Can't move it there as this variable has to be set before creating a umf target. Moved it to the CMakeLists file right before defining of a umf target.

bratpiorka (approved):

lukaszstolarczuk (approved):

PR #887, opened 2024-11-12
Deprecate the `UMF_MEM_MAP_SYNC` flag

ldorau:
### Description

Deprecate the `UMF_MEM_MAP_SYNC` flag

The `UMF_MEM_MAP_SYNC` flag is DEPRECATED,
(`UMF_MEM_MAP_SHARED` should be used instead).
`UMF_MEM_MAP_SYNC` will be removed in the next release.

Verify if /dev/dax was mapped with `MAP_SYNC`.

### Checklist

- [ ] All tests pass locally


include/umf/memory_provider.h
```diff
@@ -21,7 +21,8 @@ extern "C" {
 typedef enum umf_memory_visibility_t {
     UMF_MEM_MAP_PRIVATE = 1, ///< private memory mapping
     UMF_MEM_MAP_SHARED,      ///< shared memory mapping (Linux only)
-    UMF_MEM_MAP_SYNC, ///< direct mapping of persistent memory (supported only for files supporting DAX, Linux only)
+    UMF_MEM_MAP_SYNC =
```
lplewa, line 24: i think __attribute__((deprecated)) should work here

bratpiorka: please just redefine new flag (remove "UMF_MEM_MAP_SYNC =")

vinser52: On the UMF tech meeting, we agreed that `UMF_MEM_MAP_SYNC` can be simply removed.

It is an API/ABI breaking change but in this particular case, it is OK because we are mostly sure that nobody uses it.

ldorau: Removed.
Done

vinser52 (approved):

lplewa (approved):

PR #890, opened 2024-11-12
Do not run tests under proxy lib if libumf is not a shared lib

ldorau:
### Description

Do not run tests under the proxy library if libumf is not a shared library,
because the proxy library is built only if libumf is a shared library.

ref. #761

### Checklist

- [x] All tests pass locally


bratpiorka:
this is a fix for existing fails in CI?

bratpiorka (approved):

lukaszstolarczuk (approved):

PR #902, opened 2024-11-17
Enable jemalloc pool test with BA_GLOBAL_PROVIDER

vinser52:
### Description

Enable Jemalloc pool test with `BA_GLOBAL_PROVIDER`

### Checklist

- [ ] All tests pass locally


lplewa (approved):

vinser52:
@bratpiorka @lplewa I created this PR as a reproducer for the issue with jemalloc (See #903 ). It works only from a certain version of jemalloc, but looks like our CI needs to be fixed to use the "proper" version of jemalloc.

ldorau:
@vinser52 Please test this PR on top of https://github.com/oneapi-src/unified-memory-framework/pull/941

ldorau (approved):
> @vinser52 Please test this PR on top of #941

@vinser52 I've checked - this PR works on top of #941

vinser52:
> > @vinser52 Please test this PR on top of #941
> 
> @vinser52 I've checked - this PR works on top of #941

thank you, @ldorau, Sorry I was busy yesterday and wanted to check it today.

ldorau:
@vinser52 https://github.com/oneapi-src/unified-memory-framework/pull/941 has been merged, so you can rebase this PR now

bratpiorka:
I guess this would require to add the wrapper for hwloc_topology_init() with __attribute__((disable_sanitizer_instrumentation)) or similar

ldorau:
@vinser52 I've tried to fix that, but compiler attributes disabling AddressSanitizer do not work:
https://github.com/ldorau/unified-memory-framework/commit/6bf72751c6ddb14bd226fecde93e2bb6468b0881
https://github.com/ldorau/unified-memory-framework/actions/runs/12947284287

ldorau:
BTW: on Ubuntu 24.04 it works well

vinser52:
> @vinser52 @bratpiorka Should this PR be still opened?

So I just have no enough time to resolve the CI issues. if someone can take over I would be happy. But in general I think it makes sense to have such tests in CI

bratpiorka:
> @vinser52 @bratpiorka This test will not work because of an issue in jemalloc - jemalloc does not free the allocated memory extents on which the `arena_extent_split()` failed:
> 
> Jemalloc splits allocated memory extents - it calls `arena_extent_split()` that calls `umfMemoryProviderAllocationSplit()` but `BA_GLOBAL_PROVIDER` does not support `split()` op and `arena_extent_split()` always fails. Then jemalloc allocates a new smaller memory extent, but it never frees the first extent on which the `split()` failed and - as a result - the assert in the tracker is always triggered, because the tracker is not empty

Thanks for looking at this. I think that there would be no problem with adding support for split and merge for BA but the question is if we want to do this because of these tests

ldorau:
> I think that there would be no problem with adding support for split and merge for BA but the question is if we want to do this because of these tests

@vinser52 @bratpiorka 
In my opinion adding support for split() and merge() to `BA_GLOBAL_PROVIDER` would require too much effort now and it is not worth doing that just for enabling only one test.

In my opinion this PR can be closed for now.

vinser52:
> > I think that there would be no problem with adding support for split and merge for BA but the question is if we want to do this because of these tests
> 
> @vinser52 @bratpiorka In my opinion adding support for split() and merge() to `BA_GLOBAL_PROVIDER` would require too much effort now and it is not worth doing that just for enabling only one test.
> 
> In my opinion this PR can be closed for now.

I think we should not waste time on adding split/merge support to BA allocator.

did I get correctly that there is a bug in jemalloc? It should support scenario when split/merge fails but in fact, there is a memory leak, right? Should we create an issue in their GitHub?

ldorau:
> > > I think that there would be no problem with adding support for split and merge for BA but the question is if we want to do this because of these tests
> > 
> > 
> > @vinser52 @bratpiorka In my opinion adding support for split() and merge() to `BA_GLOBAL_PROVIDER` would require too much effort now and it is not worth doing that just for enabling only one test.
> > In my opinion this PR can be closed for now.
> 
> I think we should not waste time on adding split/merge support to BA allocator.
> 
> did I get correctly that there is a bug in jemalloc? It should support scenario when split/merge fails but in fact, there is a memory leak, right? Should we create an issue in their GitHub?

Right. I have submitted the issue in jemalloc: https://github.com/jemalloc/jemalloc/issues/2815

@vinser52 Can we close this PR?

ldorau:
@vinser52 I have submitted https://github.com/oneapi-src/unified-memory-framework/pull/1193 instead of this PR

Issue #903, opened 2024-11-17
Crash when Jemalloc pool is used with BA_PROVIDER


vinser52:
## Environment Information

- UMF version (hash commit or a tag):                         main
- OS(es) version(s):                                          Ubuntu 22.04.5 LTS

## Please provide a reproduction of the bug:

I have created PR #902 as a reproducer for this bug.

## How often bug is revealed:

The bug is always reproduced on Ubuntu 22.04.4, but works fine on Ubuntu 24.04.1. Probably it is caused by different versions of `libjemalloc` library.

## Actual behavior:

Tests are crashed with SEGFAULT. The reason is that the jemalloc pool calls memory provider with an invalid pointer and an internal check of the base allocator caught this.

Here is an output of the test enabled in the PR #902 with UMF logs.

### Run on the Ubuntu22.04.4:

```
svinogra@gkdse-dnp-01:~/unified-memory-framework/build$ UMF_LOG="level:debug;flush:debug;output:stdout" LD_LIBRARY_PATH=./lib:$LD_LIBRARY_PATH ./test/umf_test-jemalloc_pool --gtest_filter=jemallocPoolTest/umfPoolTest.allocFree/1
[INFO  UMF] utils_log_init: Logger enabled (UMF version: 0.10.0-dev.git311.ga3050d4, level: DEBUG, flush: DEBUG, pid: no, timestamp: no)
[DEBUG UMF] umfMemoryTrackerCreate: tracker created, handle=0x7ff9754ad068, segment map=0x7ff9754a6008
Running main() from /user/svinogra/unified-memory-framework/build/_deps/googletest-src/googletest/src/gtest_main.cc
Note: Google Test filter = jemallocPoolTest/umfPoolTest.allocFree/1
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from jemallocPoolTest/umfPoolTest
[ RUN      ] jemallocPoolTest/umfPoolTest.allocFree/1
[DEBUG UMF] umfTrackingMemoryProviderCreate: upstream=0x7ff9741f7068, tracker=0x7ff9754ad068, pool=0x7ff9754a80e8, ipcCache=0x7ff974c5e008, hIpcMappedCache=0x7ff9754b0068
[INFO  UMF] umfPoolCreateInternal: Memory pool created: 0x7ff9754a80e8
[...]

PR #906, opened 2024-11-18
Fix umf_ba_free() - do not assert on unknown pointer

ldorau:
### Description

Fix `umf_ba_free()` - do not assert on unknown pointer.

Ref: #903

### Checklist

- [x] All tests pass locally


src/base_alloc/base_alloc.c
```diff
@@ -278,7 +278,11 @@ void umf_ba_free(umf_ba_pool_t *pool, void *ptr) {
     umf_ba_chunk_t *chunk = (umf_ba_chunk_t *)ptr;
 
     utils_mutex_lock(&pool->metadata.free_lock);
-    assert(pool_contains_pointer(pool, ptr));
+    if (!pool_contains_pointer(pool, ptr)) {
```
vinser52, line 281: why should we remove this `assert`. It was able to catch the issue #903. With this PR we just disable it and issue become silent.

Issue #900, opened 2024-11-15
IPC API is not implemented in the Coarse provider


ldorau:
IPC API is not implemented in the Coarse provider.

PR #907, opened 2024-11-18
[DRAFT] Implement IPC API of the Coarse provider

ldorau:
### Description

Implement IPC API of the Coarse provider.

Split and merge in Coarse provider to enable IPC API.
Coarse provider should always call:
- `umfMemoryProviderAllocationSplit(upstream_provider)` or
- `umfMemoryProviderAllocationMerge(upstream_provider)`

when splitting or merging memory blocks, because the IPC API would not work on split/merged memory blocks otherwise.

Fixes: #900

### Checklist

- [x] All tests pass locally

### DIFF
--- a/src/provider/provider_coarse.c
+++ b/src/provider/provider_coarse.c
@@ -23,11 +23,6 @@
 
 #define COARSE_BASE_NAME "coarse"
 
-#define IS_ORIGIN_OF_BLOCK(origin, block)                                      \
-    (((uintptr_t)(block)->data >= (uintptr_t)(origin)->data) &&                \
-     ((uintptr_t)(block)->data + (block)->size <=                              \
-      (uintptr_t)(origin)->data + (origin)->size))
-
 typedef struct coarse_memory_provider_t {
     umf_memory_provider_handle_t upstream_memory_provider;
 
@@ -41,9 +36,6 @@ typedef struct coarse_memory_provider_t {
 
     size_t used_size;
     size_t alloc_size;
-
-    // upstream_blocks - tree of all blocks allocated from the upstream provider
-    struct ravl *upstream_blocks;
 
     // all_blocks - tree of all blocks - sorted by an address of data
     struct ravl *all_blocks;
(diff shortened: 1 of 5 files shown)


bratpiorka:
please add the following test:
* use Proxy Pool, Coarse Provider and L0 Provider
* alloc 4KB, free it - this way L0 should allocate single 4KB buffer that should be kept inside the Coarse Provider
* alloc 2KB twice - Coarse Prov should call split
* open IPC handle on both allocations

lplewa (changes requested):
This PR will not work with L0/Cuda providers as they do not (and probably will not) have split/merge functionality. 

Not sure how we should solve this issue. Tracking provider all ready has this functionality on the pool level . We could copy it to coarse provider - but is this HORRIBLE idea (code duplication and performance implications).

Performance is something which should also consider. We have too much trackers, list, and trees. I'm currently working on benchmark those, but still in my opinion, we should try to limit those to absolute minimum. When at each level we have minimum one extra data structure with non constant time complexity of inserts or delete. 

For now i would recommend to make this PR a draft, and have good architectural discussion about this issue.

vinser52:
> This PR will not work with L0/Cuda providers as they do not (and probably will not) have split/merge functionality.
> 
> Not sure how we should solve this issue. Tracking provider all ready has this functionality on the pool level . We could copy it to coarse provider - but is this HORRIBLE idea (code duplication and performance implications).
> 
> Performance is something which should also consider. We have too much trackers, list, and trees. I'm currently working on benchmark those, but still in my opinion, we should try to limit those to absolute minimum. When at each level we have minimum one extra data structure with non constant time complexity of inserts or delete.
> 
> For now i would recommend to make this PR a draft, and have good architectural discussion about this issue.

I totally agree with @lplewa. In fact, the coarse provider implements memory pooling logic, but it is implemented as a provider and used inside UMF stack as a provider. As a result, it resides below our memory tracker, because tracking provider is always on the top (between the pool manager and memory providers). I agree that we need an architecture discussion current implementation experience shows that current approach has issues.

ldorau:
> please add the following test:
> 
> * use Proxy Pool, Coarse Provider and L0 Provider
> * alloc 4KB, free it - this way L0 should allocate single 4KB buffer that should be kept inside the Coarse Provider
> * alloc 2KB twice - Coarse Prov should call split
> * open IPC handle on both allocations

```c
static umf_result_t ze_memory_provider_allocation_split(void *provider,
                                                        void *ptr,
                                                        size_t totalSize,
                                                        size_t firstSize) {
    (void)provider;
    (void)ptr;
    (void)totalSize;
    (void)firstSize;

    // TODO not supported yet
}
```

ldorau:
Replaced by #931

Issue #904, opened 2024-11-17
The `disable_provider_free` option of jemalloc cause memory leak


vinser52:
## Environment Information

- UMF version (hash commit or a tag):                         main branch

## Please provide a reproduction of the bug:

When the `disable_provider_free` option for the jemalloc pool is set it causes memory leaks.

The `arena_extent_alloc` function:
```c
static void *arena_extent_alloc(extent_hooks_t *extent_hooks, void *new_addr,
                                size_t size, size_t alignment, bool *zero,
                                bool *commit, unsigned arena_ind) {
    (void)extent_hooks; // unused
    (void)arena_ind;    // unused
    umf_result_t ret;

    jemalloc_memory_pool_t *pool = get_pool_by_arena_index(arena_ind);

    void *ptr = new_addr;
    ret = umfMemoryProviderAlloc(pool->provider, size, alignment, &ptr);
    if (ret != UMF_RESULT_SUCCESS) {
        return NULL;
    }

    if (new_addr != NULL && ptr != new_addr) {
        if (!pool->disable_provider_free) {
            umfMemoryProviderFree(pool->provider, ptr, size);
        }
        return NULL;
    }

    ...

    return ptr;
}
```

So if `new_addr != NULL && ptr != new_addr` and `pool->disable_provider_free` are both `true` the function returns NULL. So the memory returned by the `umfMemoryProviderAlloc` is never used by the pool and is not returned back to the provider.

## How often bug is revealed:

always

## Actual behavior:

Memory leaks.

## Expected behavior:

No memory leaks

## Details

I still want to discuss whether it makes sense to make the `free` function of the memory provider ops structure optional.

I understand that DEVDAX and file providers do not support `free` today. But there is no existing pool manager that can work with it. And as mentioned in this bug report we even introduced a memory leak in jemalloc when we tried to make the jemalloc pool work with DEVDAX and file providers.

So I want to raise two related questions:
1. Should any pool manager assume that the memory provider could not support the `free` operation? If yes, how a pool manager should handle it? As we can see above it is an error-prone approach.
2. Can the DEVDAX and FILE providers be used directly by the pool, or should they always be wrapped with the coarse provider? We already have the coarse provider that solves why not just always wrap DEVDAX and file providers with the coarse provider? We should do it implicitly for the user. And all pool managers would not require special support for the case when memory provider does not support `free` operation.

## Additional information about Priority and Help Requested:

[...]

PR #911, opened 2024-11-19
Remove the disable provider free parameter of jemalloc pool

ldorau:
### Description

Remove the disable_provider_free parameter of jemalloc pool.
Replace it with the Coarse provider in the example.

Fixes: #904

Requires:
- [x] #931 
- [x] #932 

### Checklist

- [x] All tests pass locally


include/umf/pools/pool_jemalloc.h
```diff
@@ -19,7 +19,8 @@ extern "C" {
 
 /// @brief Configuration of Jemalloc Pool
 typedef struct umf_jemalloc_pool_params_t {
-    /// Set to true if umfMemoryProviderFree() should never be called.
+    /// DEPRECATED and unused, will be removed in the next release,
```
vinser52, line 22: Let's not change interfaces without discussion. I am worried that we are changing interfaces carelessly.

Why are you removing this option? Is it a proper fix? Let's discuss first. It is a bad practice that we are changing interfaces without deep analysis. Let's discuss it at the UMF technical meeting.

ldorau: > Let's not change interfaces without discussion. I am worried that we are changing interfaces carelessly.
> 
OK

> Why are you removing this option? 
>

Enabling this option will always cause a memory leak.

include/umf/pools/pool_jemalloc.h
```diff
@@ -34,15 +34,6 @@ umfJemallocPoolParamsCreate(umf_jemalloc_pool_params_handle_t *hParams);
 umf_result_t
 umfJemallocPoolParamsDestroy(umf_jemalloc_pool_params_handle_t hParams);
 
-/// @brief  Set if \p umfMemoryProviderFree() should never be called.
```
vinser52, line 37: Hmm, now we have no params for the jemalloc pool.
Do we need `umfJemallocPoolParamsCreate` and  `umfJemallocPoolParamsDestroy` functions at all?

ldorau: We do not need them now. I have left them for now in case they will be needed in the future. 
I am not sure if they should be removed or not, but I can do it now, of course.
@vinser52 @bratpiorka what do you think?

vinser52: Yeah. It might confuse users that such API exists but is useless.
I vote to remove it. We can easily add it in the future.

bratpiorka: agree to remove it

ldorau: Removed.
Done.

ldorau:
@bratpiorka @vinser52 please re-review.

bratpiorka:
I like the line add/remove statistics in this PR: +4 −287 :D

bratpiorka (approved):

vinser52 (approved):

vinser52:
Since it is an API change, let's include it in `0.10`. I do not want this API to appear in 0.10 and then be removed.

ldorau:
@lukaszstolarczuk @bratpiorka 0.10 ?

vinser52:
Ok, after the discussion with @bratpiorka we agreed not to merge it in 0.10. The reason is that this PR depends on coarse provider changes and would require a lot of related changes to be merged into the 0.10. And we would like to avoid that. jemalloc is not used by oneAPI anyway.

Issue #844, opened 2024-10-25
Backward compatibility for configs of pools/providers


vinser52:
## Rationale

UMF should provide backward compatible interfaces.

### Background

MPI team experienced the issue after the PR #692 extends `level_zero_memory_provider_params_t` config structure. The root cause of the issue relates to how MPI instantiate/initialize the L0 provider config. They did the following:
```c
level_zero_memory_provider_params_t l0_config = {
    .level_zero_context_handle = hContext;
    .level_zero_device_handle = hDevice;
    .memory_type = UMF_MEMORY_TYPE_DEVICE;
};

umf_memory_provider_handle_t hProvider = NULL;
umf_memory_provider_ops_t *l0_ops = umfLevelZeroMemoryProviderOps();
umfMemoryProviderCreate(l0_ops, &l0_config, &hProvider);
```

After PR #692 the code above start crashing because two new fields were added to the `level_zero_memory_provider_params_t` data structure and the example above does not initialize these new fields:
```c
typedef struct level_zero_memory_provider_params_t {
    ze_context_handle_t level_zero_context_handle;
    ze_device_handle_t level_zero_device_handle;
    umf_usm_memory_type_t memory_type;

    // New fields
    ze_device_handle_t *resident_device_handles;
    uint32_t resident_device_count;
} level_zero_memory_provider_params_t;
```

A quick fix for the issue was to init the config data structure with `0` and than assign the required fields:
```c
level_zero_memory_provider_params_t l0_config = { 0 }; // zero-initialize
l0_config.level_zero_context_handle = hContext;
l0_config.level_zero_device_handle = hDevice;
l0_config.memory_type = UMF_MEMORY_TYPE_DEVICE;
```

### Open Questions

The quick fix above works only because it is OK to init fields of the `level_zero_memory_provider_params_t` structure with zeroes. But there are 2 related major question we should address:

#### 1. How to initialize the configs with default values?

In general case, not every field of the config data structure could/should be initialized with zeroes.

#### 2. How to support backward compatibility?

Config data structures are defined in the headers. If application was built with old version of UMF but on the system there is a newer version than even the application initialize all config's fields properly it initializes only fields that exists in the old version of UMF.
Consider an example. There is a `provider_foo` and corresponding `foo_config_t` structure that contains an int field in the 1st version:

```c
// provider_foo.h:
typedef struct foo_params_t {
    int field 1;
} foo_config_t;

// provider_foo.c:
[...]

PR #920, opened 2024-11-20
Update Level Zero provider config API

vinser52:
### Description
Change the L0 provider config as we agreed in #844

### Checklist

- [ ] All tests pass locally


examples/level_zero_shared_memory/level_zero_shared_memory.c
```diff
@@ -49,18 +49,47 @@ int main(void) {
...
     umf_memory_provider_handle_t ze_memory_provider;
-    res = umfMemoryProviderCreate(umfLevelZeroMemoryProviderOps(),
-                                  &ze_memory_provider_params,
-                                  &ze_memory_provider);
+    res =
+        umfMemoryProviderCreate(umfLevelZeroMemoryProviderOps(),
+                                ze_memory_provider_params, &ze_memory_provider);
     if (res != UMF_RESULT_SUCCESS) {
         fprintf(stderr, "Failed to create a memory provider!\n");
         ret = -1;
```
PatKamin, line 95: Destroy params -> `goto provider_params_destroy`

vinser52: Done

bratpiorka (approved):

include/umf/providers/provider_level_zero.h
```diff
@@ -17,20 +17,56 @@ extern "C" {
...
-    uint32_t
-        resident_device_count; ///< Number of devices for which the memory should be made resident
-} level_zero_memory_provider_params_t;
+struct umf_level_zero_memory_provider_params_t;
+
+/// @brief handle to the parameters of the level zero memory provider.
+typedef struct umf_level_zero_memory_provider_params_t
+    *umf_level_zero_memory_provider_params_handle_t;
+
+/// @brief  Create a struct to store parameters of the level zero memory provider.
```
bratpiorka, line 26: please use capital letters in comments referring to UMF Level Zero Memory Provider, as it is a proper noun (we should probably cleanup this in other places in our code / documentation in separate PRs)

vinser52: Done

test/providers/level_zero_helpers.cpp
```diff
@@ -63,9 +63,68 @@ struct libze_ops {
...
+template <typename... Args> ze_result_t noop_stub(Args &&...) {
+    return ZE_RESULT_SUCCESS; // Always return ZE_RESULT_SUCCESS
+}
+
 struct DlHandleCloser {
     void operator()(void *dlHandle) {
         if (dlHandle) {
+            // Reset all function pointers to no-op stubs in case the library
+            // but some other global object still try to call Level Zero functions.
+            libze_ops.zeInit = [](auto... args) { return noop_stub(args...); };
```
igchor, line 76: As I understand, the issue here is that the library is unloaded (when we use dlopen) before we call destroyContext.

Should we also add something similar to the L0 provider itself? For example, I can imagine someone (SYCL) trying to free memory in the global object destructor. @vinser52 could you extend the test to see if calling umfMemoryProviderFree() in the global object dtor works? (assuming the provider itself is also handled by that global object)

@pbalcer any thoughts on that? do you know if there were any changes in how SYCL destroys global objects on windows?

src/provider/provider_level_zero.c
```diff
@@ -134,14 +194,105 @@ static void init_ze_global_state(void) {
...
+
 static umf_result_t ze_memory_provider_initialize(void *params,
                                                   void **provider) {
     if (params == NULL) {
         return UMF_RESULT_ERROR_INVALID_ARGUMENT;
     }
 
-    level_zero_memory_provider_params_t *ze_params =
-        (level_zero_memory_provider_params_t *)params;
+    umf_level_zero_memory_provider_params_handle_t ze_params =
```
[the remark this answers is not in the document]
vinser52: Done

src/provider/provider_level_zero.c
```diff
@@ -134,14 +194,105 @@ static void init_ze_global_state(void) {
...
+        (umf_level_zero_memory_provider_params_handle_t)params;
 
     if (!ze_params->level_zero_context_handle) {
```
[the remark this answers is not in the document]
vinser52: Done.
Not moved but copied. I added the similar check to the `umfLevelZeroMemoryProviderParamsSetContext`. BUt we still need to check here because the `umfLevelZeroMemoryProviderParamsCreate` function creates `params` with `NULL` context by default.

include/umf/providers/provider_level_zero.h
```diff
@@ -17,20 +17,56 @@ extern "C" {
...
+
+/// @brief  Destroy parameters struct.
+/// @param  hParams handle to the parameters of the Level Zero Memory Provider.
+/// @return UMF_RESULT_SUCCESS on success or appropriate error code on failure.
+umf_result_t umfLevelZeroMemoryProviderParamsDestroy(
+    umf_level_zero_memory_provider_params_handle_t hParams);
+
+/// @brief  Set the Level Zero context handle in the parameters struct.
+/// @param  hParams handle to the parameters of the Level Zero Memory Provider.
+/// @param  hContext handle to the Level Zero context.
```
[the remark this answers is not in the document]
vinser52: Done

include/umf/providers/provider_level_zero.h
```diff
@@ -17,20 +17,56 @@ extern "C" {
...
+/// @return UMF_RESULT_SUCCESS on success or appropriate error code on failure.
+umf_result_t umfLevelZeroMemoryProviderParamsSetContext(
+    umf_level_zero_memory_provider_params_handle_t hParams,
+    ze_context_handle_t hContext);
+
+/// @brief  Set the Level Zero device handle in the parameters struct.
+/// @param  hParams handle to the parameters of the Level Zero Memory Provider.
+/// @param  hDevice handle to the Level Zero device.
```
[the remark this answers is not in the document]
vinser52: Done

test/providers/level_zero_helpers.cpp
vinser52: > As I understand, the issue here is that the library is unloaded (when we use dlopen) before we call destroyContext.

Correct. Our test creates the L0 context and stores it in a global object. The test also opens L0 library via dlopen and stores the handle to it in a global `std::unique_ptr'. When the test ends, the global objects start destructing. The `unique_ptr' with `dlHandle' destructs first and calls `dlclose' for the L0 library. Then the global object with the L0 context is destroyed and calls the `libze_ops.zeContextDestroy` which is already invalid. Such order of destruction happens in our CI only on Windows, but I believe it is also possible on Linux.

> Should we also add something similar to the L0 provider itself? For example, I can imagine someone (SYCL) trying to free memory in the global object destructor.

Yeah, I also thought about such scenario. But L0 provider does not control load/unload of L0 library, the L0 provider just calls `dlsym`. So in general it is a client responsibility to make sure that L0 library is unloaded after L0 provider is destroyed.
When a user uses L0 Provider there is the following sequence of actions to create L0 provider: 
1. L0 library should be loaded.
2. Create L0 context.
3. Create L0 provider.
4. allocate memory from L0 provider

The destruction should happens in the reverse order:
1. Deallocate all memory to L0 provider.
2. Destroy L0 provider.
3. Destroy L0 context.
4. L0 library is unloaded.

If the destruction order is broken it is an issue on the client side. On the UMF level we cannot control that. The L0 provider does not know when L0 library is unloaded, as a result, we do not know when to substitute real callbacks with stubs.

>  @vinser52 could you extend the test to see if calling umfMemoryProviderFree() in the global object dtor works? (assuming the provider itself is also handled by that global object)

We can add such test, but as I described below there is nothing we can do on the UMF side. The test will just test itself (not the UMF implementation) how well it is handle destruction order. Anyway I think it is should be a separate PR.

PR #928, opened 2024-11-22
Update DevDax provider config API

vinser52:
### Description

Change the L0 provider config as we agreed in #844 

### Checklist

- [x] All tests pass locally

### DIFF
--- a/include/umf/providers/provider_devdax_memory.h
+++ b/include/umf/providers/provider_devdax_memory.h
@@ -18,15 +18,43 @@ extern "C" {
 #define UMF_DEVDAX_RESULTS_START_FROM 2000
 /// @endcond
 
-/// @brief Memory provider settings struct
-typedef struct umf_devdax_memory_provider_params_t {
-    /// path of the device DAX
-    char *path;
-    /// size of the device DAX in bytes
-    size_t size;
-    /// combination of 'umf_mem_protection_flags_t' flags
-    unsigned protection;
-} umf_devdax_memory_provider_params_t;
+struct umf_devdax_memory_provider_params_t;
+
+typedef struct umf_devdax_memory_provider_params_t
+    *umf_devdax_memory_provider_params_handle_t;
+
+/// @brief  Create a struct to store parameters of the Devdax Memory Provider.
+/// @param  hParams [out] handle to the newly created parameters struct.
+/// @param  path [in] path of the device DAX.
+/// @param  size [in] size of the device DAX in bytes.
+/// @return UMF_RESULT_SUCCESS on success or appropriate error code on failure.
+umf_result_t umfDevDaxMemoryProviderParamsCreate(
+    umf_devdax_memory_provider_params_handle_t *hParams, const char *path,
+    size_t size);
+
+/// @brief  Destroy parameters struct.
+/// @param  hParams [in] handle to the parameters of the Devdax Memory Provider.
+/// @return UMF_RESULT_SUCCESS on success or appropriate error code on failure.
+umf_result_t umfDevDaxMemoryProviderParamsDestroy(
+    umf_devdax_memory_provider_params_handle_t hParams);
+
+/// @brief  Set a device DAX in the parameters struct. Overwrites the previous value.
+///         It provides an ability to use the same instance of params to create multiple
+///         instances of the provider for different DAX devices.
+/// @param  hParams [in] handle to the parameters of the Devdax Memory Provider.
+/// @param  path [in] path of the device DAX.
+/// @param  size [in] size of the device DAX in bytes.
+/// @return UMF_RESULT_SUCCESS on success or appropriate error code on failure.
+umf_result_t umfDevDaxMemoryProviderParamsSetDeviceDax(
+    umf_devdax_memory_provider_params_handle_t hParams, const char *path,
+    size_t size);
+
+/// @brief  Set the protection flags in the parameters struct.
+/// @param  hParams [in] handle to the parameters of the Devdax Memory Provider.
+/// @param  protection [in] combination of 'umf_mem_protection_flags_t' flags.
+/// @return UMF_RESULT_SUCCESS on success or appropriate error code on failure.
+umf_result_t umfDevDaxMemoryProviderParamsSetProtection(
+    umf_devdax_memory_provider_params_handle_t hParams, unsigned protection);
 
 /// @brief Devdax Memory Provider operation results
 typedef enum umf_devdax_memory_provider_native_error {
@@ -38,18 +66,6 @@ typedef enum umf_devdax_memory_provider_native_error {
 } umf_devdax_memory_provider_native_error_t;
 
 umf_memory_provider_ops_t *umfDevDaxMemoryProviderOps(void);
-
-/// @brief Create default params for the devdax memory provider
-static inline umf_devdax_memory_provider_params_t
-umfDevDaxMemoryProviderParamsDefault(char *path, size_t size) {
-    umf_devdax_memory_provider_params_t params = {
-        path, /* path of the device DAX */
-        size, /* size of the device DAX in bytes */
-        UMF_PROTECTION_READ | UMF_PROTECTION_WRITE, /* protection */
-    };
-
-    return params;
-}
 
 #ifdef __cplusplus
 }
--- a/src/provider/provider_devdax_memory.c
+++ b/src/provider/provider_devdax_memory.c
@@ -24,9 +24,41 @@ umf_memory_provider_ops_t *umfDevDaxMemoryProviderOps(void) {
     return NULL;
 }
 
+umf_result_t umfDevDaxMemoryProviderParamsCreate(
+    umf_devdax_memory_provider_params_handle_t *hParams, const char *path,
+    size_t size) {
+    (void)hParams;
+    (void)path;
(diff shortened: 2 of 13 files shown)


ldorau (changes requested):
The code coverage of the `src/provider/provider_devdax_memory.c` file on the main branch is 80.2% while on this PR only 75.6%

vinser52:
> The code coverage of the `src/provider/provider_devdax_memory.c` file on the main branch is 80.2% while on this PR only 75.6%

Can we somehow see which lines are not covered?

lukaszstolarczuk:
> > The code coverage of the `src/provider/provider_devdax_memory.c` file on the main branch is 80.2% while on this PR only 75.6%
> 
> Can we somehow see which lines are not covered?

when you enter summary of the actions ([here for your PR](https://github.com/oneapi-src/unified-memory-framework/actions/runs/12011348948) you can scroll down for artifacts - there's coverage_html_report included (to browse locally on your computer)

ldorau (approved):
The code coverage of the `src/provider/provider_devdax_memory.c` file on the main branch is 80.2% while on this PR 78.7% - still can be better ;-)

ldorau (approved):
The code coverage of the `src/provider/provider_devdax_memory.c` file is **82.5%** now - great! :-) Thanks!

bratpiorka (approved):

test/provider_devdax_memory_not_impl.cpp
```diff
@@ -0,0 +1,30 @@
...
+    umf_result =
+        umfDevDaxMemoryProviderParamsSetDeviceDax(nullptr, "path", 4096);
+    EXPECT_EQ(umf_result, UMF_RESULT_ERROR_NOT_SUPPORTED);
+
+    umf_result = umfDevDaxMemoryProviderParamsSetProtection(nullptr, 0);
+    EXPECT_EQ(umf_result, UMF_RESULT_ERROR_NOT_SUPPORTED);
+
+    umf_memory_provider_ops_t *ops = umfDevDaxMemoryProviderOps();
+    EXPECT_EQ(ops, nullptr);
+}
```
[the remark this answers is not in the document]
vinser52: Done

include/umf/providers/provider_devdax_memory.h
```diff
@@ -18,15 +18,41 @@ extern "C" {
...
+/// @return UMF_RESULT_SUCCESS on success or appropriate error code on failure.
+umf_result_t umfDevDaxMemoryProviderParamsDestroy(
+    umf_devdax_memory_provider_params_handle_t hParams);
+
+/// @brief  Set device DAX in the parameters struct.
+/// @param  hParams [in] handle to the parameters of the Devdax Memory Provider.
+/// @param  path [in] path of the device DAX.
+/// @param  size [in] size of the device DAX in bytes.
+/// @return UMF_RESULT_SUCCESS on success or appropriate error code on failure.
+umf_result_t umfDevDaxMemoryProviderParamsSetDeviceDax(
```
lplewa, line 46: Do we need this function, if this information is passed on params create? This might be confusing for the user

src/provider/provider_devdax_memory.c
```diff
@@ -511,4 +550,100 @@ umf_memory_provider_ops_t *umfDevDaxMemoryProviderOps(void) {
...
+        return UMF_RESULT_ERROR_INVALID_ARGUMENT;
+    }
+
+    size_t path_len = strlen(path);
+    if (path_len == 0) {
+        LOG_ERR("DevDax path is empty");
+        return UMF_RESULT_ERROR_INVALID_ARGUMENT;
+    }
+
+    path_len += 1; // for the null terminator
```
lplewa, line 620: nit: I really prefer something like this, but this is also fine.
```suggestion
    path_len += sizeof('\0');`
```

src/provider/provider_devdax_memory.c
```diff
@@ -511,4 +550,100 @@ umf_memory_provider_ops_t *umfDevDaxMemoryProviderOps(void) {
...
+    char *new_path = umf_ba_global_alloc(path_len);
+    if (new_path == NULL) {
+        LOG_ERR("Allocating memory for the DevDax path failed");
+        return UMF_RESULT_ERROR_OUT_OF_HOST_MEMORY;
+    }
+
+    strncpy(new_path, path, path_len);
+
+    umf_ba_global_free(hParams->path);
```
lplewa, line 629:
```suggestion
    if (hParmams->path)
        umf_ba_global_free(hParams->path);
```

src/provider/provider_devdax_memory.c
```diff
@@ -511,4 +550,100 @@ umf_memory_provider_ops_t *umfDevDaxMemoryProviderOps(void) {
...
+}
+
+umf_result_t umfDevDaxMemoryProviderParamsSetProtection(
+    umf_devdax_memory_provider_params_handle_t hParams, unsigned protection) {
+    if (hParams == NULL) {
+        LOG_ERR("DevDax Memory Provider params handle is NULL");
+        return UMF_RESULT_ERROR_INVALID_ARGUMENT;
+    }
+
+    hParams->protection = protection;
```
lplewa, line 644: We should check protcection contains only UMF_PROTECTION_* bit set.

src/provider/provider_devdax_memory.c
vinser52: We do not need to check for `NULL`

src/provider/provider_devdax_memory.c
vinser52: Done. Corresponding tests are added.

src/provider/provider_devdax_memory.c
```diff
@@ -511,4 +550,110 @@ umf_memory_provider_ops_t *umfDevDaxMemoryProviderOps(void) {
...
+umf_result_t umfDevDaxMemoryProviderParamsSetProtection(
+    umf_devdax_memory_provider_params_handle_t hParams, unsigned protection) {
+    if (hParams == NULL) {
+        LOG_ERR("DevDax Memory Provider params handle is NULL");
+        return UMF_RESULT_ERROR_INVALID_ARGUMENT;
+    }
+
+    // verify that protection contains only valid bits set
+    // (UMF_PROTECTION_MAX-1) - highest possible bit
+    // (UMF_PROTECTION_MAX-1) << 1 - next highest bit
```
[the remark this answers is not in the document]
vinser52: hmmm, I do not think `second` is a good word here.
Maybe `next after highest possible bit`? What do you think?

include/umf/providers/provider_devdax_memory.h
vinser52: I tried to extend the comment. Please review.

src/provider/provider_devdax_memory.c
lplewa: It's only to mark it that is for null terminator, instead of comment - as i said it do not matter, and you can left it as it is.

lplewa (approved):

test/pools/jemalloc_coarse_devdax.cpp
```diff
@@ -7,14 +7,39 @@
...
 #include "pool_coarse.hpp"
 
+using devdax_params_unique_handle_t =
+    std::unique_ptr<umf_devdax_memory_provider_params_t,
+                    decltype(&umfDevDaxMemoryProviderParamsDestroy)>;
+
+devdax_params_unique_handle_t create_devdax_params() {
+    char *path = getenv("UMF_TESTS_DEVDAX_PATH");
+    char *size = getenv("UMF_TESTS_DEVDAX_SIZE");
+    if (path == nullptr || path[0] == 0 || size == nullptr || size[0] == 0) {
```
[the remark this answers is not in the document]
lukaszstolarczuk: hmm... not sure if I got your comment right, but in case `devdaxParams` handle is `NULL` we return empty vector (`std::vector<poolCreateExtParams>{};`) and the test is silently passing with zero actual test cases (tested this locally). I guess we could SKIP if empty vector is passed to the `umfPoolTest` (but I'm not sure if this will work with other test suites using `umfPoolTest`)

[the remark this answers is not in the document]
vinser52: > in case devdaxParams handle is NULL we return empty vector (`std::vector<poolCreateExtParams>{};`) 

Ah, you are right, I worgot that in case of `NULL` we are not instantiating the test at all.

> and the test is silently passing with zero actual test cases (tested this locally)

If params is `NULL` the test is not instantiated at all with this PR.

> I guess we could SKIP if empty vector is passed to the umfPoolTest

As I understand, the test test is not instantiated if the vector is empty.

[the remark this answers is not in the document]
vinser52: I understand that you want somehow check if the required environment variables were not set, but I do not know what is the right way.

I think it is not a good idea: passing `NULL` params and check inside generic test if it is an instantiation of the test for devdax provider and params is `NULL` and SKIP.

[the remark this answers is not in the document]
lukaszstolarczuk: to conclude I think we have this silent "SKIP" issue in multiple places in our tests, so we don't have to fix it in this PR; let's leave it as is

Issue #843, opened 2024-10-25
Re-enable ICX test in CI


lukaszstolarczuk:
ref. #842

lukaszstolarczuk:
if any fix is required it should also be applied to v0.9.x stable branch, ref. https://github.com/oneapi-src/unified-memory-framework/pull/857

lukaszstolarczuk:
ups, we still have to fix this in nightly (ICX on Windows)
https://github.com/oneapi-src/unified-memory-framework/blob/main/.github/workflows/nightly.yml#L93

PR #936, opened 2024-11-26
disable ICX tests in nightly builds

bratpiorka:
disable ICX tests in nightly builds

issue: https://github.com/oneapi-src/unified-memory-framework/issues/843

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -89,92 +89,93 @@ jobs:
     - name: Run tests under valgrind
       run: ${{github.workspace}}/test/test_valgrind.sh ${{github.workspace}} ${{github.workspace}}/build ${{matrix.tool}}
 
-  icx:
-    name: ICX
-    env:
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly.yml
```diff
@@ -89,92 +89,93 @@ jobs:
...
-    - name: Run tests
-      shell: cmd
-      working-directory: ${{env.BUILD_DIR}}
-      run: |
-        call "C:\Program Files (x86)\Intel\oneAPI\setvars.bat"
-        call "C:\Program Files (x86)\Intel\oneAPI\setvars-vcvarsall.bat"
-        ctest -C ${{matrix.build_type}} --output-on-failure --test-dir test
+  # TODO fix #843
+  #icx:
+  #  name: ICX
```
lukaszstolarczuk, line 94: FYI, it would suffice to just add `if: false` line here; advantage is that it would still show up on the jobs' list (and remind us to fix it 😉), but would be just skipped

lukaszstolarczuk (approved):

PR #939, opened 2024-11-27
skip ba_global_free when BA is already destroyed

bratpiorka:
Skip ba_global_free when BA is already destroyed.

This could happen when the globally allocated object with a destructor that calls any UMF free/destroy function is called after UMF tear down and base alloc destroy. Then, any dereferencing of previously allocated (and mmaped) by base alloc pointer would cause a crash (as the pointer is already unmapped).

### DIFF
--- a/src/base_alloc/base_alloc_global.h
+++ b/src/base_alloc/base_alloc_global.h
@@ -8,6 +8,8 @@
 #ifndef UMF_BASE_ALLOC_GLOBAL_H
 #define UMF_BASE_ALLOC_GLOBAL_H 1
 
+#include <stdbool.h>
+
 #include "base_alloc.h"
 
 #ifdef __cplusplus
@@ -17,6 +19,7 @@ extern "C" {
 void *umf_ba_global_alloc(size_t size);
 void umf_ba_global_free(void *ptr);
 void umf_ba_destroy_global(void);
+bool umf_ba_global_is_destroyed(void);
 size_t umf_ba_global_malloc_usable_size(void *ptr);
 void *umf_ba_global_aligned_alloc(size_t size, size_t alignment);
 
--- a/src/base_alloc/base_alloc_global.c
+++ b/src/base_alloc/base_alloc_global.c
@@ -23,6 +23,7 @@
 
 // global base allocator used by all providers and pools
 static UTIL_ONCE_FLAG ba_is_initialized = UTIL_ONCE_FLAG_INIT;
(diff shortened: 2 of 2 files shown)


vinser52:
Do we have reproducer for such scenario? Should we add it as a test?

src/base_alloc/base_alloc_global.h
```diff
@@ -17,6 +19,7 @@ extern "C" {
 void *umf_ba_global_alloc(size_t size);
 void umf_ba_global_free(void *ptr);
 void umf_ba_destroy_global(void);
+bool umf_ba_global_is_destroyed();
```
vinser52, line 22: Where it is used?

bratpiorka: right now nowhere but it would be needed when Disjoint Pool would use base alloc (in C version) in umfDisjointPoolParamsDestroy

bratpiorka:
> Do we have reproducer for such scenario? Should we add it as a test?

this bug also occurs here: https://github.com/oneapi-src/unified-memory-framework/actions/runs/12041220655/job/33590212027

In the CUDA provider test (after params refactoring) we are calling umfCUDAMemoryProviderParamsDestroy in the destructor of global objects: cuParams_device_memory, cuParams_shared_memory and cuParams_host_memory. umfCUDAMemoryProviderParamsDestroy calls BA free which dereferences already unmapped pointer

vinser52:
> > Do we have reproducer for such scenario? Should we add it as a test?
> 
> this bug also occurs here: https://github.com/oneapi-src/unified-memory-framework/actions/runs/12041220655/job/33590212027
> 
> In the CUDA provider test (after params refactoring) we are calling umfCUDAMemoryProviderParamsDestroy in the destructor of global objects: cuParams_device_memory, cuParams_shared_memory and cuParams_host_memory. umfCUDAMemoryProviderParamsDestroy calls BA free which dereferences already unmapped pointer

Oh, I see. I saw the similar issue with L0 provider params when working on it and fixed it. But the corresponding CUDA PR was already merged and I forgot to apply the same fix for CUDA params. Here is the PR #940 that should fix this failure.

The issue is that `libumfInit()` is not called, because of that destructors order is incorrect.

src/base_alloc/base_alloc_global.h
vinser52: So should we close this PR?

PR #943, opened 2024-11-27
Add test for not implemented OS provider

vinser52:
### Description

Add test for not implemented OS provider

### Checklist

- [x] All tests pass locally

### DIFF
--- a/test/provider_os_memory_not_impl.cpp
+++ b/test/provider_os_memory_not_impl.cpp
@@ -1,0 +1,46 @@
+// Copyright (C) 2024 Intel Corporation
+// Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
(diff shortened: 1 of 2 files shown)


test/provider_os_memory_not_impl.cpp
```diff
@@ -0,0 +1,43 @@
...
+// Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#include "base.hpp"
+
+#include <umf/providers/provider_os_memory.h>
+
+using umf_test::test;
+
+TEST_F(test, os_provider_not_implemented) {
```
lukaszstolarczuk, line 11: you could also call `umfOsMemoryProviderOps` and check for NULL

vinser52: Done. Good catch.

bratpiorka (approved):

PR #949, opened 2024-11-28
Add missing DisjointPool symbols in .map/.def

lukaszstolarczuk:


vinser52:
Today, disjoint pool is a separate static library. So UMF has nothing to export related to disjoint pool.

PR #932, opened 2024-11-25
Use libcoarse in the file and the devdax providers

ldorau:
### Description

Use libcoarse in the file and the devdax providers.

This PR is required by 3 PRs that fix the following issues:
- https://github.com/oneapi-src/unified-memory-framework/issues/864
- https://github.com/oneapi-src/unified-memory-framework/issues/900
- https://github.com/oneapi-src/unified-memory-framework/issues/904

### Checklist

- [x] Code compiles without errors locally
- [x] All tests pass locally
- [x] CI workflows execute properly

PR #953, opened 2024-12-02
Revert moving `free()` to optional (ext) provider ops

ldorau:
### Description

Revert moving `free()` to optional (ext) provider ops.

This reverts commit b0bfbb7e92b76cc56c3bcd91c7ac063467136376.

Remove `umfDefaultFree()` and `umfIsFreeOpDefault()`.

Remove the `disable_upstream_provider_free` parameter
of the Coarse provider.

Remove the `upstreamDoesNotFree` argument of
the `umfTrackingMemoryProviderCreate()` function.

Requires:
- [x] #932 

### Checklist

- [x] All tests pass locally


ldorau:
@bratpiorka @vinser52 please review

KFilipek (approved):

vinser52 (approved):

vinser52:
@bratpiorka  I think it should be part of the `0.10`

ldorau:
> @bratpiorka I think it should be part of the `0.10`

@vinser52 Are you sure? This change requires `libcoarse` which is not present in `0.10` ...

vinser52:
> > @bratpiorka I think it should be part of the `0.10`
> 
> @vinser52 Are you sure? This change requires `libcoarse` which is not present in `0.10` ...

We are changing interfaces, the earlier we do that -> less impact for the future users. 

Regarding `libcoarse`, does it impact only file and devdax providers? Can we merge the `libcoarse` to the 0.10?

ldorau:
> Regarding `libcoarse`, does it impact only file and devdax providers?

Yes, only file and devdax providers.

bratpiorka:
@vinser52 @ldorau devdax and file providers would not be used in 0.10x and I prefer to minimize the number of changes there at least right now. If we know that there are no major problems with 0.10x, then we could think about merging changes related to the coarse lib (this is a lot of code and multiple PRs).

PR #960, opened 2024-12-03
Update umfOpenIPCHandle API to use IPC handler instead of pool

vinser52:
### Description

As we agreed on the UMF technical meeting, this PR introduces `umf_ipc_handler_handle_t` that is used by the `umfOpenIPCHandle` function instead of pool handle to open the IPC handle. In current implementation the `umf_ipc_handler_handle_t` is just a memory provider handle to the tracking provider instance.

### Checklist

- [x] All tests pass locally

### DIFF
--- a/src/ipc.c
+++ b/src/ipc.c
@@ -119,12 +119,18 @@ umf_result_t umfPutIPCHandle(umf_ipc_handle_t umfIPCHandle) {
     return ret;
 }
 
-umf_result_t umfOpenIPCHandle(umf_memory_pool_handle_t hPool,
+umf_result_t umfOpenIPCHandle(umf_ipc_handler_handle_t hIPCHandler,
                               umf_ipc_handle_t umfIPCHandle, void **ptr) {
 
-    // We cannot use umfPoolGetMemoryProvider function because it returns
-    // upstream provider but we need tracking one
-    umf_memory_provider_handle_t hProvider = hPool->provider;
+    // IPC handler is an instance of tracking memory provider
+    if (*(uint32_t *)hIPCHandler != UMF_VERSION_CURRENT) {
+        // It is a temporary hack to verify that user passes correct IPC handler,
+        // not a pool handle, as it was required in previous version.
+        LOG_ERR("Invalid IPC handler.");
+        return UMF_RESULT_ERROR_INVALID_ARGUMENT;
+    }
+
+    umf_memory_provider_handle_t hProvider = hIPCHandler;
     void *base = NULL;
 
     umf_result_t ret = umfMemoryProviderOpenIPCHandle(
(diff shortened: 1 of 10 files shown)


bratpiorka:
should this be ready for review?

vinser52:
> should this be ready for review?

The code part is ready, but I also want to update the `examples.rst` file in this PR.

vinser52:
The `umfPoolByPtr` implementation will be updated in a separate PR.

src/ipc.c
```diff
@@ -119,12 +119,11 @@ umf_result_t umfPutIPCHandle(umf_ipc_handle_t umfIPCHandle) {
...
 }
 
-umf_result_t umfOpenIPCHandle(umf_memory_pool_handle_t hPool,
+umf_result_t umfOpenIPCHandle(umf_ipc_handler_handle_t hIPCHandler,
                               umf_ipc_handle_t umfIPCHandle, void **ptr) {
 
-    // We cannot use umfPoolGetMemoryProvider function because it returns
-    // upstream provider but we need tracking one
-    umf_memory_provider_handle_t hProvider = hPool->provider;
+    // IPC handler is an instance of tracking memory provider
```
lplewa, line 125: Can we add check  `(uint32_t *)hIPCHandler == UMF_VERSION` here?

provider structure has version on first field. Pool struct has pool priv. So in this way we can detect that user provided IPChandler, not a pool ptr. This might be useful, as this function, in fact does not change from ABI - it still ()(void *, void *, void**), so in that way we can detect if user did something wrong, and give better error message. 

We could later on remove this check, but for first version with different api it might be useful

vinser52: I understand the trick you suggest to implement. You want to catch the case when a user still passes the pool handle as an argument. 

But it looks like a hack, should we use it in a situation when we have only 3 internal customers who are using this API so far? We can just negotiate with them the required change.

vinser52: Ok, Done.

src/ipc.c
```diff
@@ -119,12 +119,18 @@ umf_result_t umfPutIPCHandle(umf_ipc_handle_t umfIPCHandle) {
...
+    if (*(uint32_t *)hIPCHandler != UMF_VERSION_CURRENT) {
+        // it is a temporar hack to verify that user passes correct IPC handler,
```
ldorau, line 127: // it is a temporar hack 
- `it` -> `It`
- `temporar` -> `temporary`

ldorau (changes requested):

src/ipc.c
vinser52: Done

lplewa (approved):

vinser52:
> Do we need it on 0.10 branch?
> If yes please change target branch on this PR and then rebase to 0.10 branch(in this order)

Yes, we need. I will rebase.

How do the changes then go to the main branch? Will they be merged automatically or do I need separate PR?

lplewa:
we do PR from 0.10 to main.

vinser52:
> we do PR from 0.10 to main. Like this one:#959

Who will do that? Should I do that, or someone from your team do it from time to time?

lukaszstolarczuk:
> > we do PR from 0.10 to main. Like this one:#959
> 
> Who will do that? Should I do that, or someone from your team do it from time to time?

Yeah, I do it from time to time, or when need to include a specific patch to the main

PR #963, opened 2024-12-04
L0 provider: implement support for defer and blocking free

igchor:
### Description

This feature is necessary to properly support urUSMFree for v2 L0 adapter.

### Checklist

- [ ] All tests pass locally


bratpiorka (approved):

bratpiorka:
@igchor do we need this for 0.10x / 2025.1?

src/provider/provider_level_zero.c
```diff
@@ -61,6 +61,14 @@ umf_result_t umfLevelZeroMemoryProviderParamsSetResidentDevices(
     return UMF_RESULT_ERROR_NOT_SUPPORTED;
 }
 
+umf_result_t umfLevelZeroMemoryProviderParamsSetFreePolicy(
```
vinser52, line 64: Could you please update the test in the `provider_level_zero_not_impl.cpp` file?

vinser52 (changes requested):

igchor:
> @igchor do we need this for 0.10x / 2025.1?

No, for now I only want to use it for v2 L0 adapter which will not be released as part of 2025.1

include/umf/providers/provider_level_zero.h
```diff
@@ -68,6 +68,21 @@ umf_result_t umfLevelZeroMemoryProviderParamsSetResidentDevices(
     umf_level_zero_memory_provider_params_handle_t hParams,
     ze_device_handle_t *hDevices, uint32_t deviceCount);
 
+typedef enum umf_level_zero_memory_provider_free_policy_t {
+    UMF_LEVEL_ZERO_MEMORY_PROVIDER_FREE_POLICY_DEFAULT =
```
vinser52, line 72: Can we somehow test different policies?

src/provider/provider_level_zero.c
igchor: Done.

include/umf/providers/provider_level_zero.h
igchor: That's a bit problematic. We would need to have a test that launches a kernel, and that means an extra dependency on a compiler or committing a binary. As a first step, I wanted to test this in UR (that's why I kept this as draft). Once we have  test there, I will think if we can easily port it to this repo.

vinser52: Sounds good. I asked just in case there is a simple test. In future, we can consider running integration tests with UR/compiler.

vinser52 (approved):

vinser52:
> As a first step, I wanted to test this in UR (that's why I kept this as draft).

@igchor So we are not merging that till you confirm that UR works with the new UMF L0 config options and they do what UR needs?

igchor:
> > As a first step, I wanted to test this in UR (that's why I kept this as draft).
> 
> @igchor So we are not merging that till you confirm that UR works with the new UMF L0 config options and they do what UR needs?

Yes, I'll undraft the PR once I manage to test this in UR

include/umf/providers/provider_level_zero.h
```diff
@@ -68,6 +68,21 @@ umf_result_t umfLevelZeroMemoryProviderParamsSetResidentDevices(
...
+        0, ///< Free memory immediately. Default.
+    UMF_LEVEL_ZERO_MEMORY_PROVIDER_FREE_POLICY_BLOCKING_FREE, ///< Blocks until all commands using the memory are complete before freeing.
+    UMF_LEVEL_ZERO_MEMORY_PROVIDER_FREE_POLICY_DEFER_FREE, ///< Schedules the memory to be freed but does not free immediately.
+} umf_level_zero_memory_provider_free_policy_t;
+
+/// @brief  Set the memory free policy.
+/// @param  hParams handle to the parameters of the Level Zero Memory Provider.
+/// @param  policy memory free policy.
+/// @return UMF_RESULT_SUCCESS on success or appropriate error code on failure.
+umf_result_t umfLevelZeroMemoryProviderParamsSetFreePolicy(
```
vinser52, line 82: You also need to update the `src/libumf.def` and `src/libumf.map` files.

bratpiorka: .. and place this call in the proper set of exported functions (UMF_0.11?)

igchor: Ok, I added it under UMF_0.11 scope but how do we want to do that for windows? In .def file we have a single VERSION specified.

bratpiorka: on Windows I think we could just put the new calls at the end of the list with a comment "; added in UMF_0.11" 
This way, the map and def files will be comparable to some extent

lplewa:
any tests?

include/umf/providers/provider_level_zero.h
igchor: Ok, done.

igchor:
> any tests?

For now, this will only be tested in UR. We can consider adding some tests in the future, but this will require adding a dependency on igc or committing a binary (the test requires executing a kernel).

igchor:
@bratpiorka could you please merge this? I have a few more fixes pending that but I'd like this to be merged first

bratpiorka (approved):

PR #966, opened 2024-12-05
Fix and add error messages

ldorau:
### Description

Fix two error messages and add some new ones.

### Checklist

- [x] All tests pass locally

### DIFF
--- a/src/provider/provider_file_memory.c
+++ b/src/provider/provider_file_memory.c
@@ -485,6 +485,8 @@ static umf_result_t file_alloc_cb(void *provider, size_t size, size_t alignment,
 
     file_memory_provider_t *file_provider = (file_memory_provider_t *)provider;
 
+    *resultPtr = NULL;
+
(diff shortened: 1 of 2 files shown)


KFilipek (approved):

src/provider/provider_file_memory.c
```diff
@@ -519,6 +519,10 @@ static umf_result_t file_alloc_cb(void *provider, size_t size, size_t alignment,
                 addr, alloc_offset_fd);
     }
 
+    LOG_DEBUG("inserted a value to the file descriptor offset map (addr=%p, "
```
lukaszstolarczuk, line 522: in case of error we print both: error and debug msg

pls print perhaps before the `if (ret)` and/or check if we should not exit after LOG_ERR...?

ldorau: Fixed.
Done.

lukaszstolarczuk (approved):

lplewa (approved):

PR #969, opened 2024-12-05
Add getter functions for DisjointPoolsParams

lukaszstolarczuk:
ref. #844

// Used in UR: https://github.com/oneapi-src/unified-runtime/pull/2426

### DIFF
--- a/include/umf/pools/pool_disjoint.h
+++ b/include/umf/pools/pool_disjoint.h
@@ -52,6 +52,8 @@ umfDisjointPoolParamsDestroy(umf_disjoint_pool_params_handle_t hParams);
 umf_result_t
 umfDisjointPoolParamsSetSlabMinSize(umf_disjoint_pool_params_handle_t hParams,
                                     size_t slabMinSize);
+size_t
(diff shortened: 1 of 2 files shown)


lplewa:
> Please let me know if this is something we want to do, and if so - should this be done for all params functions?

can we workaround this in UR - i really dislike getter idea.

igchor:
@lukaszstolarczuk I think we could just pre-generate that entire output in UR in DisjointPoolAllConfigs ctor. You can probably use std::stringstream to keep the same formatting and not having to modify the code too much.

vinser52:
> > Please let me know if this is something we want to do, and if so - should this be done for all params functions?
> 
> can we workaround this in UR - i really dislike getter idea. As far i see it is only to std::cout configuration of the pool. This is not a great usecase to add new api

I agree with @lplewa I intentionally did not add the getter functions to the config API. I think it is a design issue in UR. UR should not use UMF config for their own needs. Probably UR can have their own structure to store configs values. Or use the idea that @igchor suggested.

lukaszstolarczuk:
Perfect, that's why I preped this simple draft. Much appreciate your feedback, I'll give it a try tomorrow.

PR #970, opened 2024-12-05
add test dax_provider_purge_force on unalligned address

lplewa:
I guess it will fail

### DIFF
--- a/test/provider_devdax_memory.cpp
+++ b/test/provider_devdax_memory.cpp
@@ -233,6 +233,13 @@ TEST_P(umfProviderTest, purge_force) {
     test_alloc_free_success(provider.get(), page_size, 0, PURGE_FORCE);
 }
 
+TEST_P(umfProviderTest, purge_force_unalligned_alloc) {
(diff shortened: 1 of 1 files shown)


bratpiorka:
PASSED!

test/provider_devdax_memory.cpp
```diff
@@ -233,6 +233,13 @@ TEST_P(umfProviderTest, purge_force) {
     test_alloc_free_success(provider.get(), page_size, 0, PURGE_FORCE);
 }
 
+TEST_P(umfProviderTest, purge_force_unalligned_alloc) {
```
PatKamin, line 236: unalligned -> unaligned

PatKamin, line 236: Isn't it that devdax malloc is always aligned?
https://github.com/oneapi-src/unified-memory-framework/blob/a08a1218b09573dfe8ad25ee8d56b97b1743f4cd/src/provider/provider_devdax_memory.c#L281-L284

PR #978, opened 2024-12-07
[POC] add free handler to the proxy lib

bratpiorka:
[POC] Add free handler to the proxy lib 

This PR:
* adds new API call umfSetProxyLibHandlerFreePre to the Proxy Lib which set the user's callback called before the Proxy Lib calls free()
* moves proxy_lib.h to the public API so it is now installed with other headers
* moves Proxy Lib tests to separate test folder
* adds simple test for umfSetProxyLibHandlerFreePre

NOTE: the baseline is v0.10.x as it contains IPC handler

TODO:
- [ ] Documentation
- [ ] Other handlers

### Checklist

- [ ] All tests pass locally


include/umf/proxy_lib_handlers.h
```diff
@@ -0,0 +1,27 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+// malloc API handlers
+typedef void (*umf_proxy_lib_handler_free_pre_t)(void *ptr,
+                                                 umf_memory_pool_handle_t pool);
+
+void umfSetProxyLibHandlerFreePre(umf_proxy_lib_handler_free_pre_t handler);
```
vinser52, line 21: I think we need to discuss the following question at the UMF technical meeting: Should this API be part of the proxy library or the core UMF library.

PR #982, opened 2024-12-09
set fixed iteration count for benchmarks

lplewa:
### Checklist

- [ ] All tests pass locally


bratpiorka:
Agreed as a temporary solution, as all these fixed numbers should be parameters.

lplewa:
> > Agreed as a temporary solution, as all these fixed numbers should be parameters.
> 
> then at least it would be nice to have a TODO in the code 😉

we have issue which list features for benchmark and this thing is there

pbalcer (approved):
lgtm, but please create a ticket to add CLI options to set the iterations.

PR #988, opened 2024-12-10
Run pr_push on schedule at night to get full code coverage from QEMU

ldorau:
### Description

Run pr_push on schedule at night to get full code coverage from QEMU.

### Checklist

- [x] All tests pass locally

### DIFF
--- a/.github/workflows/pr_push.yml
+++ b/.github/workflows/pr_push.yml
@@ -7,6 +7,8 @@ on:
     branches-ignore:
       - 'dependabot/**'
   pull_request:
+  schedule:
+    - cron:  '0 1 * * *'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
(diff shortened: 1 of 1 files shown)


.github/workflows/pr_push.yml
```diff
@@ -53,7 +55,7 @@ jobs:
     needs: [FastBuild]
     uses: ./.github/workflows/reusable_qemu.yml
     with:
-      short_run: true
+      short_run: ${{ github.event_name == 'schedule' && 'false' || 'true' }}
```
PatKamin, line 58: Remove `ubuntu-22.04` build from Nightly as this seems to be doubled now.

Also, shouldn't this workflow be triggered as part of a Nightly build? Now this job's results will not be as easy to find as if they were run as part of a Nightly as there are plenty of pr_push runs and only one Nightly a day.

ldorau: 1) I do not want to modify Nightly.
2) This workflow is not reusable, so AFAIK it cannot be run as a part of the Nightly build.

@lukaszstolarczuk what do you think?

ldorau: > I would ask a different question - did you observe a better code coverage with long QEMU run? TBH, I don't expect increase in number of covered lines.
> 

You are right :-) Closing ...

PR #990, opened 2024-12-11
add codespell workflow

luszczewskakasia1:
### Checklist

- [ ] All tests pass locally


.github/scripts/codespell.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+OUTPUT=$1
```
lukaszstolarczuk, line 3: is this `OUTPUT` used anywhere?

.github/workflows/reusable_codespell.yml
```diff
@@ -0,0 +1,38 @@
+name: Codespell
```
lukaszstolarczuk, line 1: I believe you can use `reusable_checks.yml` workflow and just add an extra step next to existing `Run a spell check`

.github/workflows/reusable_codespell.yml
```diff
@@ -0,0 +1,38 @@
...
+on:
+  pull_request:
+    branches: [ "main" ]
+  workflow_dispatch:
```
lukaszstolarczuk, line 5: workflow dispatch shouldn't be needed

.github/scripts/codespell.sh
```diff
@@ -0,0 +1,6 @@
...
+
+echo "Run codespell..."
+codespell --quiet-level=2 --skip "*.h,*.cmake,*.c,*.hpp,*.cpp,*.sh,*.py,test/supp/*.supp" -i 0 -L "ASSER,Tne,ba,BA" || true
```
ldorau, line 6: empty line missing at EOF

.github/workflows/reusable_codespell.yml
```diff
@@ -0,0 +1,38 @@
...
+        with:
+          python-version: '3.x'
+
+      - name: Install codespell
+        run: |
+          pip install codespell
+
+      - name: Run codespell
+        run: |
+          bash .github/scripts/codespell.sh
```
lukaszstolarczuk, line 38: if the script is just a few lines you can just put it directly here, no need for extra sh file

.github/scripts/codespell.sh
ldorau, line 6: `|| true` - means that it will always succeed, so it will never find any issues, so it is useless ;-)
It should fail, when it finds errors.

ldorau (changes requested):
@luszczewskakasia1 please squash your commits from this PR and do not commit in the merge commits:
image

.github/scripts/codespell.sh
lplewa: set -e would be nice too.

.github/workflows/reusable_codespell.yml
lplewa: Script might be useful in case in someone want to use it from their development setup. So instead looking to yml, and copying command, they can just call script which has all correct configuration. 

TBH the best solution would be to put this checker in to cmake so we have additional target, similar to cstyle, but script is still beter then yml

.github/workflows/reusable_codespell.yml
lplewa: Script might be useful in case in someone want to use it from their development setup. So instead looking to yml, and copying command, they can just call script which has all correct configuration.

TBH the best solution would be to put this checker in to cmake so we have additional target, similar to cstyle, but script is still beter then yml

luszczewskakasia1:
Created new PR in a correct way --> https://github.com/oneapi-src/unified-memory-framework/pull/994

lukaszstolarczuk:
replaced with #994

PR #995, opened 2024-12-16
Assert in umf_ba_destroy() in DEBUG and UMF_DEVELOPER_MODE

ldorau:
### Description

if (NDEBUG is not defined) and (UMF_DEVELOPER_MODE is defined)
assert in umf_ba_destroy() if there are any memory leaks
in the base allocator.

It works as a "**base-allocator-memory-leaks-checker**".

### Checklist

- [x] All tests pass locally


ldorau:
It works as a "base-allocator-memory-leaks-checker".

src/base_alloc/base_alloc.c
```diff
@@ -303,7 +303,18 @@ void umf_ba_destroy(umf_ba_pool_t *pool) {
...
+        LOG_ERR("number of base allocator memory leaks: %zu",
+                pool->metadata.n_allocs);
+
+#ifdef UMF_DEVELOPER_MODE
+        assert(pool->metadata.n_allocs == 0 &&
+               "memory leaks in base allocator occurred");
+#endif
+
+// UMF_DEVELOPER_MODE should be turned on in CI
+#if defined(UMF_CI_GITHUB_ACTION) && !defined(UMF_DEVELOPER_MODE)
```
bratpiorka, line 315: I'm not convinced that we should be doing any checks related to our CI in the UMF code.
I see that DEV MODE was probably not enabled by mistake, but imo these kinds of problems should be cached during the review, or entirely in CI. Please note that the real problem here was reported errors that no one looked at... and to solve this we should run our tests with logging enabled (nightly?) and catch any unexpected error messages.

bratpiorka (changes requested):

src/base_alloc/base_alloc.c
ldorau: OK. Removed.
Done.

bratpiorka (approved):

lukaszstolarczuk (approved):

PR #1016, opened 2024-12-31
enable building examples on win static hwloc CI

bratpiorka:


CMakeLists.txt
```diff
@@ -251,8 +251,20 @@ else()
 
         set(LIBHWLOC_INCLUDE_DIRS
             ${hwloc_targ_SOURCE_DIR}/include;${hwloc_targ_BINARY_DIR}/include)
-        set(LIBHWLOC_LIBRARY_DIRS
-            ${hwloc_targ_BINARY_DIR}/Release;${hwloc_targ_BINARY_DIR}/Debug)
+        if(CMAKE_BUILD_TYPE STREQUAL "Release")
```
lukaszstolarczuk, line 254: at some point we discovered that the value of `CMAKE_BUILD_TYPE` in Windows does not make any sense and we select build type during building (`cmake --build build --config Release`)... so we'll never set `CMAKE_BUILD_TYPE` and the default always be picked here

PatKamin: Try using generators instead:
```
set(LIBHWLOC_LIBRARY_DIRS ${hwloc_targ_BINARY_DIR}/$<CONFIG>)
set(LIBHWLOC_LIBRARIES ${hwloc_targ_BINARY_DIR}/$<CONFIG>/hwloc.lib)
```

bratpiorka: thanks, with $<CONFIG> both Debug and Release works fine!

lukaszstolarczuk (approved):

PatKamin (approved):

Issue #979, opened 2024-12-08
Use prctl when ptrace is restricted


vinser52:
## Rationale

Today, most UMF's memory providers (all except CUDA provider) on Linux use the `pidfd_getfd(2)` system call to obtain a duplicate of another process's file descriptor. The ptrace governs permission to duplicate another process's file descriptor.
Today UMF requires the `/proc/sys/kernel/yama/ptrace_scope` to be `0`. Otherwise, the `pidfd_getfd` fails. It is a significant limitation because on Ubuntu the `/proc/sys/kernel/yama/ptrace_scope` is `1` ("restricted ptrace") by default and some UMF users might have no permission to change it.

## Description

The `PR_SET_PTRACER` allows other processes to ptrace(2) the calling process (see [man page](https://man7.org/linux/man-pages/man2/PR_SET_PTRACER.2const.html) for details). The idea is that the process that creates the IPC handle (calls `umfGetIPCHandle`) can call the `int prctl(PR_SET_PTRACER, long pid);` syscall to allow other processes to `ptrace` the current process when the `/proc/sys/kernel/yama/ptrace_scope` is `1`.
The oneCCL already employs that for the L0 backend, Intel MPI will introduce it soon. We should use this mechanism as well.

## API Changes

No API changes. Only implementation should be updated

## Implementation details

1. The `int prctl(PR_SET_PTRACER, long pid);` syscall should called once per process. We can do it in the UMF init or we can do it lazily at first access to IPC functionality.
2. By default, oneCCL and Intel MPI use the parent process ID (retrieved using `getppid()`) to allow only processes with the same parent (MPI/oneCCL ranks have the same parent: `mpirun`) to ptrace the current process. UMF can use the same approach and provide an environment variable to tweak that and allow all processes to ptrace the current process. 

### How to test ###

Set the `/proc/sys/kernel/yama/ptrace_scope` to `1` and run IPC tests.

## Meta

[PR_SET_PTRACER](https://man7.org/linux/man-pages/man2/PR_SET_PTRACER.2const.html)

bratpiorka:
looks great

ldorau:
> Today, most UMF's memory providers (all except CUDA provider) on Linux use the `pidfd_getfd(2)` system call to obtain a duplicate of another process's file descriptor. 

AFAIK `pidfd_getfd(2)` is used only in the `utils_duplicate_fd()` function, which is used **only in 2 providers**: OS and L0.

[...]

PR #1018, opened 2025-01-07
Set ptracer permissions for IPC handle creation

PatKamin:
Fixes: #979 

### Checklist

- [ ] All tests pass locally

- [x] Logger (with debug/info/... messages) is used

### DIFF
--- a/examples/ipc_ipcapi/ipc_ipcapi_anon_fd.sh
+++ b/examples/ipc_ipcapi/ipc_ipcapi_anon_fd.sh
@@ -1,5 +1,5 @@
 #
-# Copyright (C) 2024 Intel Corporation
+# Copyright (C) 2024-2025 Intel Corporation
 #
 # Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
@@ -16,16 +16,8 @@ PORT=$(( 1024 + ( $$ % ( 65535 - 1024 ))))
 # to obtain a duplicate of another process's file descriptor.
 # Permission to duplicate another process's file descriptor
 # is governed by a ptrace access mode PTRACE_MODE_ATTACH_REALCREDS check (see ptrace(2))
-# that can be changed using the /proc/sys/kernel/yama/ptrace_scope interface.
-PTRACE_SCOPE_FILE="/proc/sys/kernel/yama/ptrace_scope"
-VAL=0
-if [ -f $PTRACE_SCOPE_FILE ]; then
-	PTRACE_SCOPE_VAL=$(cat $PTRACE_SCOPE_FILE)
-	if [ $PTRACE_SCOPE_VAL -ne $VAL ]; then
-		echo "SKIP: ptrace_scope is not set to 0 (classic ptrace permissions) - skipping the test"
-		exit 125 # skip code defined in CMakeLists.txt
-	fi
-fi
+# In the producer binary used in this example prctl(PR_SET_PTRACER, getppid()) is used
+# to allow consumer to duplicate file descriptor of producer.
 
 UMF_LOG_VAL="level:debug;flush:debug;output:stderr;pid:yes"
 
--- a/examples/ipc_ipcapi/ipc_ipcapi_producer.c
+++ b/examples/ipc_ipcapi/ipc_ipcapi_producer.c
@@ -6,9 +6,11 @@
  */
 
 #include <arpa/inet.h>
+#include <errno.h>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#include <sys/prctl.h>
 #include <sys/socket.h>
 #include <unistd.h>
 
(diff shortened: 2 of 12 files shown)


KFilipek (approved):

examples/ipc_ipcapi/ipc_ipcapi_anon_fd.sh
```diff
@@ -12,21 +12,6 @@ set -e
 # port should be a number from the range <1024, 65535>
 PORT=$(( 1024 + ( $$ % ( 65535 - 1024 ))))
```
lukaszstolarczuk, line 14: it's this time of the year - pls, bump license(s)

lukaszstolarczuk (changes requested):

lplewa:
1. Should the library call it? imho this should be something done by the user, not by the library. So we should only update examples, and documentation, so user knows how to use this API. Imho this is security variability, if we do it always, without user consent. 

2. Do it even work? We set permission to parent process. Our tests will work as producer and customer shares the same parent process. What if producer, and consumer are different parents? I.E are start from different bash processes.

PatKamin:
> 1. Should the library call it? imho this should be something done by the user, not by the library. So we should only update examples, and documentation, so user knows how to use this API. Imho this is security variability, if we do it always, without user consent.

It's for the ease-of-use of the UMF IPC API. Perhaps at least we should enable tracing with a new CMake option so the user enables tracing explicitly? And this would be disabled by default as UMF users not using IPC API would not be interested in setting `ptrace` scope wider.

> 2. Do it even work? We set permission to parent process. Our tests will work as producer and customer shares the same parent process. What if producer, and consumer are different parents? I.E are start from different bash processes.

It works for `MPI` ranks communication. That's the case covered in our tests where producer and consumer share the same parent process. As for the different parents case, it's not tested but I think it could work if they are using the same shared `umf` library.

examples/ipc_ipcapi/ipc_ipcapi_anon_fd.sh
PatKamin: Thanks, done

lplewa:
> As for the different parents case, it's not tested but I think it could work if they are using the same shared `umf` library.

This is about kernel level process security protection/permissions - common shared library between processes do not have to influence here.

ldorau (changes requested):
26 tests out of 56 fail on WSL with the following error:

```
umf_example_ipc_ipcapi_producer: unified-memory-framework/src/ipc_cache.c:136: umfIpcHandleMappedCacheCreate: Assertion `IPC_MAPPED_CACHE_GLOBAL != NULL' failed.
Aborted (core dumped)
```

lukaszstolarczuk:
just a note here: we've established to rather updated docs, examples and our tests, than use the current solution...

PatKamin:
Updated PR with `prctl()` call removed from the umf library but used in tests and mentioned in comments and `README`.

examples/ipc_ipcapi/ipc_ipcapi_producer.c
```diff
@@ -6,9 +6,11 @@
  */
```
lukaszstolarczuk, line 6: sry, one more date left to bump

test/common/ipc_common.c
```diff
@@ -6,9 +6,11 @@
  */
```
lukaszstolarczuk, line 6: and another date

src/utils/utils_posix_common.c
```diff
@@ -91,9 +91,8 @@ umf_result_t utils_duplicate_fd(int pid, int fd_in, int *fd_out) {
     return UMF_RESULT_ERROR_NOT_SUPPORTED;
 #else
     // pidfd_getfd(2) is used to obtain a duplicate of another process's file descriptor.
-    // Permission to duplicate another process's file descriptor
-    // is governed by a ptrace access mode PTRACE_MODE_ATTACH_REALCREDS check (see ptrace(2))
-    // that can be changed using the /proc/sys/kernel/yama/ptrace_scope interface.
+    // Calling prctl(PR_SET_PTRACER, getppid()) in a producer binary that creates IPC handle
+    // allow file descriptor duplication for parent process and its children
```
lukaszstolarczuk, line 95: `allow` -> `allows`; plus a full stop at the end of sentence

lukaszstolarczuk (commented):
beside minor issues, LGTM

examples/ipc_ipcapi/ipc_ipcapi_producer.c
```diff
@@ -69,6 +71,23 @@ int main(int argc, char *argv[]) {
...
+    // use the pidfd_getfd(2) system call to duplicate another process's file descriptor, which is
+    // governed by ptrace permissions. By default on Ubuntu /proc/sys/kernel/yama/ptrace_scope is
+    // set to 1 ("restricted ptrace"), which prevents pidfd_getfd from working unless ptrace_scope
+    // is set to 0.
+    // To overcome this limitation without requiring users to change the ptrace_scope
+    // setting (which requires root privileges), we use prctl() to allow the consumer process
+    // to copy producer's file descriptor, even when ptrace_scope is set to 1.
+    ret = prctl(PR_SET_PTRACER, getppid());
+    if (ret == -1) {
+        printf("prctl() call failed with errno %d (%s). This may indicate that "
```
lplewa, line 85: perror?

test/common/ipc_common.c
```diff
@@ -336,6 +338,14 @@ int run_producer(int port, umf_memory_pool_ops_t *pool_ops, void *pool_params,
     int producer_socket = -1;
     char consumer_message[MSG_SIZE];
 
+    ret = prctl(PR_SET_PTRACER, getppid());
+    if (ret == -1) {
+        printf("prctl() call failed with errno %d (%s). This may indicate that "
```
lplewa, line 343: perror

lplewa (changes requested):

examples/ipc_ipcapi/ipc_ipcapi_producer.c
lukaszstolarczuk: Done

test/common/ipc_common.c
lukaszstolarczuk: Done

lukaszstolarczuk:
Patryk is off at the moment, so finishing up the minor issues left here - contd in https://github.com/oneapi-src/unified-memory-framework/pull/1040

PR #1030, opened 2025-01-09
Make ICX enabled again

### LINKED ISSUE #843: Re-enable ICX test in CI

KFilipek:
### Description
There is a problem with linkage under icx/icpx due to linking against compiler libs.
When executing tests without sourced paths to compiler (prevents from loading compiler version of UMF) there are errors of missing symbols/libs.

### Considerations
There was 2 options:
- link statically against compiler dependencies using `-static-intel`
- omit linking Intel compiler extensions with `-no-intel-lib` ([more info](https://www.intel.com/content/www/us/en/docs/cpp-compiler/developer-guide-reference/2021-8/no-intel-lib-qno-intel-lib.html)) or using another option `-Wl,-as-needed'`

First works like a charm, but the UMF contains icx/icpx needed dependencies compiled in, I don't know it's a problem or not. 
Last option require to force `gtest` to compile the same way - without Intel compiler dependencies (e.g. `svml`, `imf` etc). According to `gtest` documentation we can share `CMAKE_(C/CXX)_FLAGS`. Another approach is to split steps and apply patch.

Fixes #843

### Checklist

- [x] All tests pass locally

### DIFF
--- a/.github/workflows/reusable_sanitizers.yml
+++ b/.github/workflows/reusable_sanitizers.yml
@@ -77,7 +77,6 @@ jobs:
         ASAN_OPTIONS: allocator_may_return_null=1
         TSAN_OPTIONS: allocator_may_return_null=1
       run: |
-        ${{ matrix.compiler.cxx == 'icpx' && '. /opt/intel/oneapi/setvars.sh' || true }}
         ctest --output-on-failure
 
   windows-build:
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -1,9 +1,15 @@
-# Copyright (C) 2022-2024 Intel Corporation
+# Copyright (C) 2022-2025 Intel Corporation
 # Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
 # SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED YES)
+
+if(CMAKE_C_COMPILER_ID STREQUAL "IntelLLVM")
+    # Compiler dependencies needs to be in library path or to be linked
+    # statically
+    add_link_options(-static-intel)
+endif()
 
 include(FetchContent)
 FetchContent_Declare(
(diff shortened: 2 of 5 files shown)


.github/workflows/reusable_sanitizers.yml
```diff
@@ -77,7 +77,6 @@ jobs:
         ASAN_OPTIONS: allocator_may_return_null=1
         TSAN_OPTIONS: allocator_may_return_null=1
       run: |
-        ${{ matrix.compiler.cxx == 'icpx' && '. /opt/intel/oneapi/setvars.sh' || true }}
         ctest --output-on-failure
```
ldorau, line 80: Why is `LD_LIBRARY_PATH=${{env.BUILD_DIR}}/lib/` not needed like it is above?

bratpiorka: @KFilipek please check if setting LD_LIBRARY_PATH is still needed if we don't source oneAPI env vars

KFilipek: It's still needed by shared library, this case above is compiled statically with the UMF, so there is no need to pass shared library.

ldorau (approved):

PatKamin (approved):

test/CMakeLists.txt
```diff
@@ -5,6 +5,12 @@
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED YES)
```
lukaszstolarczuk, line 7: it's this time of the year - pls updates copyright dates

lukaszstolarczuk (commented):
beside copyright and removing redundant commit - LGTM

KFilipek:
> heh, I appreciate using my branch for this PR, but it contains a test commit "[[CI] test only icx](https://github.com/oneapi-src/unified-memory-framework/pull/1030/commits/ac6682783b963d17bbb71d38992f5339eea46ee0)" - I guess it should be removed ;-)

I left it intentionally because it was also your work :P

test/CMakeLists.txt
KFilipek: Updated, thanks

lukaszstolarczuk (approved):

PR #1037, opened 2025-01-14
Add sles15 to multi numa c final

intel12232289:
Add SLES support.

Older conversation at https://github.com/oneapi-src/unified-memory-framework/pull/1007#pullrequestreview-2517114003

https://github.com/intel12232289/unified-memory-framework/actions/runs/12767806141/job/35587841005
https://github.com/intel12232289/unified-memory-framework/actions/runs/12767806141/job/35587841609
https://github.com/intel12232289/unified-memory-framework/actions/runs/12767806141/job/35587842189
https://github.com/intel12232289/unified-memory-framework/actions/runs/12767806141/job/35587842525

PR #1038, opened 2025-01-14
Add-SLES15-to-multi-NUMA-CI

intel12232289:
https://github.com/oneapi-src/unified-memory-framework/pull/1037


.github/workflows/reusable_multi_numa.yml
```diff
@@ -19,7 +19,7 @@ jobs:
 
     strategy:
       matrix:
-        os: [ubuntu-22.04, rhel-9.1]
+        os: [ubuntu-22.04, rhel-9.1, sles-15]
```
PatKamin, line 22: It would be nice to have a name containing an SP version of SLES

PR #1042, opened 2025-01-16
reduce code duplication in Benchmark by using Apply() to set arguments

lplewa:
### Checklist

- [ ] All tests pass locally

### DIFF
--- a/benchmark/benchmark.cpp
+++ b/benchmark/benchmark.cpp
@@ -1,10 +1,12 @@
 /*
- * Copyright (C) 2024 Intel Corporation
+ * Copyright (C) 2024-2025 Intel Corporation
  *
  * Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
  * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
  *
  */
+
+#include <benchmark/benchmark.h>
 
 #include "benchmark.hpp"
 
@@ -18,14 +20,8 @@
 
 #define UMF_BENCHMARK_REGISTER_F(BaseClass, Method)                            \
     BENCHMARK_REGISTER_F(BaseClass, Method)                                    \
-        ->ArgNames(                                                            \
-            BENCHMARK_PRIVATE_CONCAT_NAME(BaseClass, Method)::argsName())      \
-        ->Name(BENCHMARK_PRIVATE_CONCAT_NAME(BaseClass, Method)::name())       \
-        ->Iterations(                                                          \
-            BENCHMARK_PRIVATE_CONCAT_NAME(BaseClass, Method)::iterations())
-
-UMF_BENCHMARK_TEMPLATE_DEFINE(alloc_benchmark, glibc_fix, fixed_alloc_size,
-                              glibc_malloc);
+        ->Apply(                                                               \
+            &BENCHMARK_PRIVATE_CONCAT_NAME(BaseClass, Method)::defaultArgs)
 
 // Benchmarks scenarios:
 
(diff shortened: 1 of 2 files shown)


bratpiorka (approved):

benchmark/benchmark.cpp
```diff
@@ -109,71 +100,67 @@ UMF_BENCHMARK_TEMPLATE_DEFINE(alloc_benchmark, scalable_pool_fix,
...
+static void
+default_multiple_alloc_fix_size(benchmark::internal::Benchmark *benchmark) {
+    benchmark->Args({10000, 4096});
+    benchmark->Threads(4);
+    benchmark->Threads(1);
+}
+
+static void
+default_multiple_alloc_uniform_size(benchmark::internal::Benchmark *benchmark) {
+    benchmark->Args({10000, 0, 8, 64 * 1024, 8});
```
lukaszstolarczuk, line 122: fyi, some of these before was `({10000, 8, 64 * 1024, 8})` (one extra `0` here) - is that expected? ;)

lplewa: nope - this was copy paste error - nice catch

benchmark/benchmark.cpp
```diff
@@ -109,71 +102,67 @@ UMF_BENCHMARK_TEMPLATE_DEFINE(alloc_benchmark, scalable_pool_fix,
...
 
 // TODO: debug why this crashes
 /*UMF_BENCHMARK_TEMPLATE_DEFINE(multiple_malloc_free_benchmark,
                               disjoint_pool_uniform, uniform_alloc_size,
                               pool_allocator<disjoint_pool<os_provider>>);
 UMF_BENCHMARK_REGISTER_F(multiple_malloc_free_benchmark, disjoint_pool_uniform)
-    ->Args({10000, 0, 8, 64 * 1024, 8})
-    ->Threads(4)
-    ->Threads(1);
+    ->Apply(&default_multiple_alloc_uniform_size);
```
lukaszstolarczuk, line 165: soo, now this one is either wrongly copy-pasted (and should be `default_alloc_uniform_size`), or previously it was wrongly set...?

benchmark/benchmark.cpp
```diff
@@ -18,69 +20,65 @@
...
 
 UMF_BENCHMARK_TEMPLATE_DEFINE(alloc_benchmark, proxy_pool, fixed_alloc_size,
                               pool_allocator<proxy_pool<os_provider>>);
 
 UMF_BENCHMARK_REGISTER_F(alloc_benchmark, proxy_pool)
-    ->Args({1000, 0, 4096})
-    ->Args({1000, 100000, 4096})
-    ->Threads(4)
-    ->Threads(1);
+    ->Apply(&default_alloc_fix_size);
```
lukaszstolarczuk, line 67: soo this one had 1k instead of 10k iterations? perhaps was that on purpose before? if not, pls disregard

benchmark/benchmark.cpp
```diff
@@ -182,17 +171,13 @@ UMF_BENCHMARK_TEMPLATE_DEFINE(multiple_malloc_free_benchmark, jemalloc_pool_fix,
...
+    ->Apply(&default_multiple_alloc_fix_size);
 
 UMF_BENCHMARK_TEMPLATE_DEFINE(multiple_malloc_free_benchmark,
                               jemalloc_pool_uniform, uniform_alloc_size,
                               pool_allocator<jemalloc_pool<os_provider>>);
 UMF_BENCHMARK_REGISTER_F(multiple_malloc_free_benchmark, jemalloc_pool_uniform)
-    ->Args({1000, 8, 64 * 1024, 8})
-    ->Threads(4)
-    ->Threads(1);
+    ->Apply(&default_multiple_alloc_uniform_size);
```
lukaszstolarczuk, line 180: sam here, no extra `0` before

benchmark/benchmark.cpp
lplewa: This is exactly why i did this patch - so eliminate copypaste errors - now is correct

benchmark/benchmark.cpp
lplewa: Now is correct

benchmark/benchmark.cpp
lplewa: and still no zero - it is default_multiple_alloc_uniform_size function

lukaszstolarczuk: perfect, thx for double checking

lukaszstolarczuk (approved):

PR #1055, opened 2025-01-23
do not use global ctor/dtor for test params

bratpiorka:
Do not use global constructor/destructors for test params.

### Description
Currently, all provider and pool tests use parameters created during the instantiation phase, with automatic destructors called upon program exit. The problem occurs when the creation of some parameters requires our global base allocator (in the case for the Disjoint Pool in the C version - not merged yet). In this scenario, the global base allocator's free() function could be called after the deletion of the entire global base allocator, because we cannot guarantee the correct order of destructors. Additionally, in the failing scenario, the base allocator's debug checks will report memory leaks.

### Checklist

- [x] All tests pass locally

### DIFF
--- a/test/ipcAPI.cpp
+++ b/test/ipcAPI.cpp
@@ -1,4 +1,4 @@
-// Copyright (C) 2023-2024 Intel Corporation
+// Copyright (C) 2023-2025 Intel Corporation
 // Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
 // SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 // This file contains tests for UMF pool API
@@ -115,5 +115,6 @@ HostMemoryAccessor hostMemoryAccessor;
 
 INSTANTIATE_TEST_SUITE_P(umfIpcTestSuite, umfIpcTest,
                          ::testing::Values(ipcTestParams{
-                             umfProxyPoolOps(), nullptr, &IPC_MOCK_PROVIDER_OPS,
-                             nullptr, &hostMemoryAccessor}));
+                             umfProxyPoolOps(), nullptr, nullptr,
+                             &IPC_MOCK_PROVIDER_OPS, nullptr, nullptr,
+                             &hostMemoryAccessor}));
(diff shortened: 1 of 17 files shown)


vinser52:
> Currently, all provider and pool tests use parameters created during the instantiation phase, with automatic destructors called upon program exit. The problem occurs when the creation of some parameters requires our global base allocator (in the case for the Disjoint Pool in the C version - not merged yet). In this scenario, the global base allocator's free() function could be called after the deletion of the entire global base allocator, because we cannot guarantee the correct order of destructors. Additionally, in the failing scenario, the base allocator's debug checks will report memory leaks.

Does it mean that in real use cases, the same problem exists? I mean if some shared library (e.g. UR adaptor) will create disjoint pool params in the library ctor than we have the same issue as described?

bratpiorka:
> Does it mean that in real use cases, the same problem exists? I mean if some shared library (e.g. UR adaptor) will create disjoint pool params in the library ctor than we have the same issue as described?

library ctor should be fine. The real problem is with global ctors/dtors like here: https://github.com/oneapi-src/unified-memory-framework/pull/1055/files#diff-0884623826c0b9a057c7517f627f3f2f96062c7b14ff570599a115e46a2dd526L246

What I see on Windows is that the base alloc is destroyed before the global object destructors are called and because of this we see mem leaks in base alloc

src/provider/provider_level_zero.c
```diff
@@ -366,8 +366,15 @@ static umf_result_t ze_memory_provider_initialize(void *params,
...
-    if ((ze_params->memory_type == UMF_MEMORY_TYPE_HOST) ==
+    if ((ze_params->memory_type == UMF_MEMORY_TYPE_HOST) &&
         (ze_params->level_zero_device_handle != NULL)) {
+        LOG_ERR("For HOST memory type the Level Zero device handle "
+                "must be NULL");
+        return UMF_RESULT_ERROR_INVALID_ARGUMENT;
+    }
+
+    if ((ze_params->memory_type != UMF_MEMORY_TYPE_HOST) &&
+        (ze_params->level_zero_device_handle == NULL)) {
```
ldorau, line 377: Does this change should belong to this PR? What does it have in common?

PatKamin (approved):
Looks good

src/provider/provider_level_zero.c
bratpiorka: reverted

ldorau (approved):

Issue #700, opened 2024-08-27
Init TBB symbols lazily but only once


vinser52:
## Description

Today UMF loads TBB library via `dlopen` every time the scalable pool instance is created. It is done by `tbb_pool_initialize` function by calling `init_tbb_callbacks`. We can optimize it by loading TBB library and looking for appropriate TBB symbols only once (at the first access). 

## API Changes

N/A

## Implementation details

A scalable pool instance should not have a copy of TBB symbols, it should be a global structure of tbb symbols that is initialized lazily only once.

PR #1064, opened 2025-01-30
Init TBB symbols only once

wlemkows:
Fixes: https://github.com/oneapi-src/unified-memory-framework/issues/700

### Checklist

- [x] All tests pass locally

### DIFF
--- a/src/pool/pool_scalable.c
+++ b/src/pool/pool_scalable.c
@@ -20,6 +20,7 @@
 
 #include "base_alloc_global.h"
 #include "libumf.h"
+#include "pool_scalable_internal.h"
 #include "utils_common.h"
 #include "utils_concurrency.h"
 #include "utils_load_library.h"
@@ -33,6 +34,7 @@ static __TLS umf_result_t TLS_last_allocation_error;
 static __TLS umf_result_t TLS_last_free_error;
 
 static const size_t DEFAULT_GRANULARITY = 2 * 1024 * 1024; // 2MB
+
 typedef struct tbb_mem_pool_policy_t {
     raw_alloc_tbb_type pAlloc;
     raw_free_tbb_type pFree;
@@ -66,7 +68,6 @@ typedef struct tbb_callbacks_t {
 typedef struct tbb_memory_pool_t {
     umf_memory_provider_handle_t mem_provider;
     void *tbb_pool;
-    tbb_callbacks_t tbb_callbacks;
 } tbb_memory_pool_t;
 
 typedef enum tbb_enums_t {
@@ -81,6 +82,10 @@ typedef enum tbb_enums_t {
     TBB_POOL_MSIZE,
     TBB_POOL_SYMBOLS_MAX // it has to be the last one
 } tbb_enums_t;
+
+static UTIL_ONCE_FLAG tbb_initialized = UTIL_ONCE_FLAG_INIT;
+static int tbb_init_result = 0;
+static tbb_callbacks_t tbb_callbacks = {0};
 
 static const char *tbb_symbol[TBB_POOL_SYMBOLS_MAX] = {
 #ifdef _WIN32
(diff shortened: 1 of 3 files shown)


src/pool/pool_scalable.c
```diff
@@ -109,42 +111,46 @@ static const char *tbb_symbol[TBB_POOL_SYMBOLS_MAX] = {
...
 
-static int init_tbb_callbacks(tbb_callbacks_t *tbb_callbacks) {
-    assert(tbb_callbacks);
-
+static int init_tbb_callbacks() {
     const char *lib_name = tbb_symbol[TBB_LIB_NAME];
-    tbb_callbacks->lib_handle = utils_open_library(lib_name, 0);
-    if (!tbb_callbacks->lib_handle) {
-        LOG_ERR("%s required by Scalable Pool not found - install TBB malloc "
+    if (tbb_callbacks.initialized < 1) {
```
vinser52, line 116: This approach is not thread safe. The common practice is to use `init_once`. In UMF we have `utils_init_once` function for that. Please see how it is used in other places.

vinser52 (changes requested):

src/pool/pool_scalable.c
lplewa: init_once will not work in this case, unless we add dlclose() to umf destructor(what i'm not great fan). Otherwise you will get memleaks do not closing library.

Refcount implemented here is correct. We just need use either a lock, or fetch_and_add on refcounter. I opt for second option.

vinser52: In case of a sequence of `umfPoolCreate()/umfPoolDestroy()` calls the current implementation will load/unload TBB library. I think we should do it only once. What is the reason for unloading TBB lib immediately after the pool is destroyed?

lplewa: This is unlikely that someone will destroy all tbb pools and then create new one, and even so  pool create is not performance critical code. We have to close library at some point, otherwise vslgrind will report memleaks in our code. 
We have two options: 
- this PR approach - unload then last pool is destroyed.
- unload in UMF destructor. 
I do not like second approach. UMF modual design, should should enclose all tbb pool code in pool ops. If we want to destroy it in umf destructor we must do it under ifdef as umf can be compiled without TBB.  This is "leaking abstraction" while this PR approach encapsulates all tbb code in one module.

vinser52: > This is unlikely that someone will destroy all tbb pools and then create new one, and even so pool create is not performance critical code.

If something might happen, it will happen :). Different clients of UMF might create pool at a different stage of an application.

> * this PR approach - unload then last pool is destroyed.
> * unload in UMF destructor.

TBB lib is a dependency of UMF lib. I think the dependency should be loaded once. We are loading it lazy when it is needed first time and should keep until the UMF lib itself is unloaded.

> I do not like second approach. UMF modual design, should should enclose all tbb pool code in pool ops. If we want to destroy it in umf destructor we must do it under ifdef as umf can be compiled without TBB. This is "leaking abstraction" while this PR approach encapsulates all tbb code in one module.

I disagree that there is an “abstraction leak”. Code responsible to TBB symbols load resides in TBB-related source files. We should put appropriate function call (that will `dlclose(tbb_handle)`) into UMF destructor - and we already do that for other “modules”.

P.S.: this discussion makes me thinking that we might have issues with L0 provider as well. Today we do not `dlopen(“libze_loader.so”)` at all. We suppose that UMF’s client, who uses Level Zero provider, loaded `libze_loader.so` into the process. But what if there are two clients in the process:

1.  The first client loads `libze_loader.so` into the process and uses Level Zero provider. In that case UMF inits Level Zero symbols via `dlsym`. Then first client destroys the Level Zero Memory provider and unloads the `libze_loader.so`.
2. Then the second client loads `libze_loader.so` into the process and uses Level Zero provider. But our current implementation does not catch such situation and Level Zero symbols considered initialized, but the `dlsym` should be done again.

To fix that, I think UMF should acquire handle to the `libze_loader.so` via `dlopen`. There is `RTLD_NOLOAD` flag that tells `dlopen` not to load the library and succeed only if the library already loaded. It will allows to increase the refcount to the `libze_loader.so` and not unload it when the first client calls dlclose.

src/pool/pool_scalable.c
```diff
@@ -109,42 +111,46 @@ static const char *tbb_symbol[TBB_POOL_SYMBOLS_MAX] = {
 #endif
 };
 
-static int init_tbb_callbacks(tbb_callbacks_t *tbb_callbacks) {
-    assert(tbb_callbacks);
-
+static int init_tbb_callbacks() {
```
ldorau, line 114: `void` should be used here (`static int init_tbb_callbacks(void) {`) - there are compilers that require that (on MacOS AFAIR)

ldorau (changes requested):

src/pool/pool_scalable.c
lplewa: > > This is unlikely that someone will destroy all tbb pools and then create new one, and even so pool create is not performance critical code.
> 
> If something might happen, it will happen :). Different clients of UMF might create pool at a different stage of an application.
>
as i said even if it will happen, we just have to dlopen & dlsym once again, not huge deal
> 
> > * this PR approach - unload then last pool is destroyed.
> > * unload in UMF destructor.
> 
> TBB lib is a dependency of UMF lib. I think the dependency should be loaded once. We are loading it lazy when it is needed first time and should keep until the UMF lib itself is unloaded.
> 
> > I do not like second approach. UMF modual design, should should enclose all tbb pool code in pool ops. If we want to destroy it in umf destructor we must do it under ifdef as umf can be compiled without TBB. This is "leaking abstraction" while this PR approach encapsulates all tbb code in one module.
> 
> I disagree that there is an “abstraction leak”. Code responsible to TBB symbols load resides in TBB-related source files. We should put appropriate function call (that will `dlclose(tbb_handle)`) into UMF destructor - and we already do that for other “modules”.
>
We do not do it for other pool/providers - we have only "core" features there: memspaces, and memory tracker. 
Pools and providers are separate modules which interact with "core" UMF thru ops structure.  So if you want provider/pool module cleanup feature in the way that do not leak umf abstraction you must do it thru ops structure, by adding a destructor function to it, which is registered when first time pool is created with this structure, and will be called in umf destruction. But i just prefer this PR approach.
> 
> P.S.: this discussion makes me thinking that we might have issues with L0 provider as well. Today we do not `dlopen(“libze_loader.so”)` at all. We suppose that UMF’s client, who uses Level Zero provider, loaded `libze_loader.so` into the process. But what if there are two clients in the process:
> 
>     1. The first client loads `libze_loader.so` into the process and uses Level Zero provider. In that case UMF inits Level Zero symbols via `dlsym`. Then first client destroys the Level Zero Memory provider and unloads the `libze_loader.so`.
> 
>     2. Then the second client loads `libze_loader.so` into the process and uses Level Zero provider. But our current implementation does not catch such situation and Level Zero symbols considered initialized, but the `dlsym` should be done again.
> 
> 
> To fix that, I think UMF should acquire handle to the `libze_loader.so` via `dlopen`. There is `RTLD_NOLOAD` flag that tells `dlopen` not to load the library and succeed only if the library already loaded. It will allows to increase the refcount to the `libze_loader.so` and not unload it when the first client calls dlclose.
>
You are right - we should do it also for L0. The only question is Windows, and how it should be implemented there.

vinser52: > as i said even if it will happen, we just have to dlopen & dlsym once again, not a huge deal

I will reinitialize the TBB library state every time we load it. While you argue that this flow is not on the critical path (that might be not true for all cases) I still not convinced that we should not keep it openned till the UMF itself is unloaded. 
I understand that under modules you mean logical blocks of the UMF library but  in fact UMF is a single shared library that has an optional dependency to TBB lib. And it is common practice when different "modules" register callbacks to the library destructor (because there is only one destructor per library).
I also mentioned Level Zero as an example that TBB dependency is not unique, and it is better to use the same approach for all dependencies that are loaded via `dlopen`. And in common case calling dlopen & dlsym once again, might be huge deal: Some libraries are not designed to be unloaded and may leave behind lingering state, memory leaks, or corrupt internal state when reloaded.

lplewa: > I will reinitialize the TBB library state every time we load it. While you argue that this flow is not on the critical path (that might be not true for all cases) I still not convinced that we should not keep it openned till the UMF itself is unloaded.

If someone creates and destroys all tbb pools multiple times, in performance critical path  - we can recomend easy workaround just create one extra pool and never close it. Maybe lite bit dirty hack, but it's simple and it will work




> I understand that under modules you mean logical blocks of the UMF library but in fact UMF is a single shared library that has an optional dependency to TBB lib. And it is common practice when different "modules" register callbacks to the library destructor (because there is only one destructor per library).

Then we have to change provider ops (and pool ops, in the similar way)
```
From ecc5af1c900e9821bee7e77f63a0019d7ceee358 Mon Sep 17 00:00:00 2001
From: =?UTF-8?q?=C5=81ukasz=20Plewa?= <lukasz.plewa@intel.com>
Date: Fri, 31 Jan 2025 17:03:40 +0100
Subject: [PATCH] provider ops

---
 include/umf/memory_provider_ops.h | 8 +++++++-
 1 file changed, 7 insertions(+), 1 deletion(-)

diff --git a/include/umf/memory_provider_ops.h b/include/umf/memory_provider_ops.h
index a61e0aad..b247b873 100644
--- a/include/umf/memory_provider_ops.h
+++ b/include/umf/memory_provider_ops.h
@@ -1,6 +1,6 @@
 /*
  *
- * Copyright (C) 2023-2024 Intel Corporation
+ * Copyright (C) 2023-2025 Intel Corporation
  *
  * Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
@@ -78,6 +78,12 @@ typedef struct umf_memory_provider_ext_ops_t {
     umf_result_t (*allocation_split)(void *hProvider, void *ptr,
                                      size_t totalSize, size_t firstSize);
 
+    /// @brief This function is registered during the first provider's initialization
+    ///        and is invoked when UMF is being destroyed. It should clean up
+    ///        any global or static state maintained by the provider (e.g.,
+    ///        release global resources, close file handles, etc.).
+    ///
+    void (*provider_destructor)(void);
 } umf_memory_provider_ext_ops_t;
 
 ///
-- 
2.43.0

```

vinser52: > If someone creates and destroys all tbb pools multiple times, in performance critical path - we can recomend easy workaround - just create one extra pool and never close it. Maybe lite bit dirty hack, but it's simple and it will work.

I think we lost the main question of our discussion. The main question is whether should we keep UMF dependencies (TBB, L0, etc) till the UMF lib itself is unloaded or should we `dlclose` ASAP and `dlopen` again if needed. My strong position is the following: it is a bad choice to unload the UMF dependency immediately because of the reasons I mentioned above.

> Then we have to change provider ops (and pool ops, in the similar way)

No. Please do not confuse internal pools/providers (that are part of libumf.so) and external ones that are parts of other binaries. In both cases, it is implementation details:
* Scalable pool implementation depends on `libtbb.so` and Level Zero provider implementation depends on `libze_loader.so`. And both are parts of `libumf.so` - so their dependencies are dependencies of `libumf.so` and should be `dlclose`-ed when `libumf.so` is unloaded.
* In case some external pool/provider implementation is part of `libfoo.so` and depends on some `libbar.so`, the `libbar.so` should be unloaded when the `libfoo.so` is unloaded. There is nothing UMF should deal with, it is the implementation details of external pool/provider

lplewa: Ok, let's do it, as you propose. So init with pthreadonce at first pool creation, and close in the library destructor

src/pool/pool_scalable.c
```diff
@@ -109,46 +114,60 @@ static const char *tbb_symbol[TBB_POOL_SYMBOLS_MAX] = {
...
+}
 
-    return 0;
+void fini_tbb_global_state(void) {
+    if (tbb_callbacks.lib_handle) {
+        if (!utils_close_library(tbb_callbacks.lib_handle)) {
+            tbb_callbacks.lib_handle = NULL;
+            LOG_DEBUG("TBB library closed");
+        } else {
+            LOG_DEBUG("TBB library cannot be unloaded");
```
vinser52, line 168: nit: should it be a higher severity level, `LOG_ERROR`?

src/pool/pool_scalable.c
```diff
@@ -264,15 +283,15 @@ static umf_result_t tbb_pool_initialize(umf_memory_provider_handle_t provider,
         return UMF_RESULT_ERROR_OUT_OF_HOST_MEMORY;
     }
 
-    int ret = init_tbb_callbacks(&pool_data->tbb_callbacks);
+    int ret = init_tbb_callbacks();
```
vinser52, line 286: I just noted that we have a possible memory leak in this function (not related to this particular PR).

If the `init_tbb_callbacks()` fails the `tbb_pool_initialize` will exit at line 289 and the `pool_data` is leaked.
Could you please fix it in this PR as well or at least create an issue?

src/pool/pool_scalable.c
ldorau: 👍

src/pool/pool_scalable.c
ldorau: 👍 it is a one-line fix (just add `umf_ba_global_free(pool_data);` there), so it will be faster to fix it than to submit an issue ;-)

vinser52: Actually, it is a two-line fix :). There are two possible returns with errors.

src/pool/pool_scalable.c
```diff
@@ -109,46 +114,60 @@ static const char *tbb_symbol[TBB_POOL_SYMBOLS_MAX] = {
...
+static int init_tbb_callbacks(void) {
+    utils_init_once(&tbb_initialized, init_tbb_callbacks_once);
+    return tbb_init_result;
+}
+
+void fini_tbb_global_state(void) {
+    if (tbb_callbacks.lib_handle) {
+        if (!utils_close_library(tbb_callbacks.lib_handle)) {
+            tbb_callbacks.lib_handle = NULL;
+            LOG_ERR("TBB library closed");
```
vinser52, line 166: Why is it a `LOG_ERR`?

vinser52 (changes requested):

src/pool/pool_scalable.c
ldorau: IMHO it should be `LOG_DEBUG`

ldorau:
LGTM besides the above issue.

wlemkows:
> LGTM besides the above issue (@wlemkows please ping me when it is fixed)

@ldorau done

vinser52 (approved):

src/pool/pool_scalable.c
```diff
@@ -82,6 +83,10 @@ typedef enum tbb_enums_t {
     TBB_POOL_SYMBOLS_MAX // it has to be the last one
 } tbb_enums_t;
 
+static UTIL_ONCE_FLAG tbb_initialized = UTIL_ONCE_FLAG_INIT;
+int tbb_init_result = 0;
```
lplewa, line 87: static?

src/pool/pool_scalable.c
```diff
@@ -264,35 +283,41 @@ static umf_result_t tbb_pool_initialize(umf_memory_provider_handle_t provider,
         return UMF_RESULT_ERROR_OUT_OF_HOST_MEMORY;
     }
 
-    int ret = init_tbb_callbacks(&pool_data->tbb_callbacks);
+    umf_result_t res = UMF_RESULT_SUCCESS;
+    int ret = init_tbb_callbacks();
     if (ret != 0) {
         LOG_FATAL("loading TBB symbols failed");
-        return UMF_RESULT_ERROR_UNKNOWN;
+        res = UMF_RESULT_ERROR_UNKNOWN;
```
ldorau, line 290: Please use `UMF_RESULT_ERROR_DEPENDENCY_UNAVAILABLE` here.

ldorau (changes requested):

src/pool/pool_scalable.c
ldorau: ... and I will merge this PR.

ldorau: I will fix it in a separate PR.

ldorau (approved):

src/pool/pool_scalable.c
ldorau: Fixed in https://github.com/oneapi-src/unified-memory-framework/pull/1107

PR #1068, opened 2025-01-31
[L0 provider] log each call to L0

igchor:
In a format that is required by SYCL tests.

This is to match UR behavior that logs every call to L0.

### Checklist

- [ ] All tests pass locally

### DIFF
--- a/src/provider/provider_level_zero.c
+++ b/src/provider/provider_level_zero.c
@@ -201,6 +201,9 @@ static umf_result_t ze2umf_result(ze_result_t result) {
     }
 }
 
+#define ZE_CALL(ZeName, ZeArgs)                                                \
+    (LOG_INFO("ZE ---> %s%s", #ZeName, #ZeArgs), g_ze_ops.ZeName ZeArgs)
+
 static void init_ze_global_state(void) {
 #ifdef _WIN32
     const char *lib_name = "ze_loader.dll";
@@ -409,8 +412,9 @@ static umf_result_t ze_memory_provider_alloc(void *provider, size_t size,
             .stype = ZE_STRUCTURE_TYPE_HOST_MEM_ALLOC_DESC,
             .pNext = NULL,
             .flags = 0};
-        ze_result = g_ze_ops.zeMemAllocHost(ze_provider->context, &host_desc,
-                                            size, alignment, resultPtr);
+
+        ze_result = ZE_CALL(zeMemAllocHost, (ze_provider->context, &host_desc,
+                                             size, alignment, resultPtr));
         break;
     }
     case UMF_MEMORY_TYPE_DEVICE: {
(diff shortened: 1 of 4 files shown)


src/provider/provider_level_zero.c
```diff
@@ -201,6 +201,9 @@ static umf_result_t ze2umf_result(ze_result_t result) {
     }
 }
 
+#define ZE_CALL(ZeName, ZeArgs)                                                \
+    (LOG_INFO("ZE ---> %s%s", #ZeName, #ZeArgs), g_ze_ops.ZeName ZeArgs)
```
vinser52, line 205: Do you think it should be `INFO` not `DEBUG`? BTW, I have no strong opinion.

igchor: DEBUG is pretty verbose right now (e.g. all inserts to memory tracker are printed). This functionality (tracing all calls to L0) is often used for debugging SYCL programs. Having all this debug information from UMF would probably be too much.

bratpiorka (approved):

bratpiorka:
@igchor LGTM, but I think changes made for specific clients should be avoided or we should make it more configurable, but I don't have a simple idea how to do that

igchor:
> @igchor LGTM, but I think changes made for specific clients should be avoided or we should make it more configurable, but I don't have a simple idea how to do that

Yeah, I know this is not ideal. I will be trying to get rid of log parsing logic from SYCL so that we do not have to follow a specific format in future (but this might take some time, until all the tests are fixed).

vinser52:
> > @igchor LGTM, but I think changes made for specific clients should be avoided or we should make it more configurable, but I don't have a simple idea how to do that
> 
> Yeah, I know this is not ideal. I will be trying to get rid of log parsing logic from SYCL so that we do not have to follow a specific format in future (but this might take some time, until all the tests are fixed).

BTW, you are trying to log every Level Zero call. Why not to make such logging on the Level Zero side?

igchor:
> > > @igchor LGTM, but I think changes made for specific clients should be avoided or we should make it more configurable, but I don't have a simple idea how to do that
> > 
> > 
> > Yeah, I know this is not ideal. I will be trying to get rid of log parsing logic from SYCL so that we do not have to follow a specific format in future (but this might take some time, until all the tests are fixed).
> 
> BTW, you are trying to log every Level Zero call. Why not to make such logging on the Level Zero side?

Yes, eventually, I would like to have logging inside L0. However, this will require a few more fixes/changes in the loader. For now i just wanted have something that we can use with SYCL right away.

vinser52:
> > > > @igchor LGTM, but I think changes made for specific clients should be avoided or we should make it more configurable, but I don't have a simple idea how to do that
> > > 
> > > 
> > > Yeah, I know this is not ideal. I will be trying to get rid of log parsing logic from SYCL so that we do not have to follow a specific format in future (but this might take some time, until all the tests are fixed).
> > 
> > 
> > BTW, you are trying to log every Level Zero call. Why not to make such logging on the Level Zero side?
> 
> Yes, eventually, I would like to have logging inside L0. However, this will require a few more fixes/changes in the loader. For now i just wanted have something that we can use with SYCL right away.

I confused than.
It sounds like you want this PR to be merged as a temporary workaround to fix the SYCL test. is it right? Two things that confuse me:

1. There is only an issue with the SYCL malformed test.
2. This PR wants a temporary workaround to be merged into UMF. Furthermore, this changes are client specific as @bratpiorka  noticed.

Based on that I have the following questions:

1. What is the urgency and priority of the issue you are trying to fix?
2. Why not fix the test if it is malformed?
3. If such logs eventually should be part of L0, what we will do once they are merged into L0 implementation? The SYCL test will fail again?

igchor:
The logging is needed not only to make SYCL tests pass but also for debugging/development (and for v2 L0 adapter to have the same capabilities as the legacy adapter).

Let me mark this as draft for now and see if we can add this logging to L0 soon. If not, I will undraft the PR.

bratpiorka:
@igchor would you like to continue working on this PR?

igchor:
This is not needed anymore, functionality implemented in the loader.

PR #1079, opened 2025-02-06
add support for CUDA allocation flags

bratpiorka:
Add support for CUDA allocation flags.

This PR adds one new API call for CUDA:
- `umfCUDAMemoryProviderParamsSetAllocFlags()`

### Checklist

- [x] All tests pass locally

### DIFF
--- a/include/umf/providers/provider_cuda.h
+++ b/include/umf/providers/provider_cuda.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2024 Intel Corporation
+ * Copyright (C) 2024-2025 Intel Corporation
  *
  * Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
  * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
@@ -52,6 +52,13 @@ umf_result_t umfCUDAMemoryProviderParamsSetDevice(
 umf_result_t umfCUDAMemoryProviderParamsSetMemoryType(
     umf_cuda_memory_provider_params_handle_t hParams,
     umf_usm_memory_type_t memoryType);
+
+/// @brief  Set the allocation flags in the parameters struct.
+/// @param  hParams handle to the parameters of the CUDA Memory Provider.
+/// @param  flags valid combination of CUDA allocation flags.
+/// @return UMF_RESULT_SUCCESS on success or appropriate error code on failure.
+umf_result_t umfCUDAMemoryProviderParamsSetAllocFlags(
+    umf_cuda_memory_provider_params_handle_t hParams, unsigned int flags);
 
 umf_memory_provider_ops_t *umfCUDAMemoryProviderOps(void);
 
--- a/src/provider/provider_cuda.c
+++ b/src/provider/provider_cuda.c
@@ -55,6 +55,14 @@ umf_result_t umfCUDAMemoryProviderParamsSetMemoryType(
     return UMF_RESULT_ERROR_NOT_SUPPORTED;
 }
 
+umf_result_t umfCUDAMemoryProviderParamsSetAllocFlags(
+    umf_cuda_memory_provider_params_handle_t hParams, unsigned int flags) {
+    (void)hParams;
+    (void)flags;
+    LOG_ERR("CUDA provider is disabled (UMF_BUILD_CUDA_PROVIDER is OFF)!");
+    return UMF_RESULT_ERROR_NOT_SUPPORTED;
+}
+
 umf_memory_provider_ops_t *umfCUDAMemoryProviderOps(void) {
     // not supported
     LOG_ERR("CUDA provider is disabled (UMF_BUILD_CUDA_PROVIDER is OFF)!");
@@ -89,13 +97,22 @@ typedef struct cu_memory_provider_t {
     CUdevice device;
     umf_usm_memory_type_t memory_type;
     size_t min_alignment;
+    unsigned int alloc_flags;
 } cu_memory_provider_t;
 
 // CUDA Memory Provider settings struct
 typedef struct umf_cuda_memory_provider_params_t {
-    void *cuda_context_handle;         ///< Handle to the CUDA context
-    int cuda_device_handle;            ///< Handle to the CUDA device
-    umf_usm_memory_type_t memory_type; ///< Allocation memory type
+    // Handle to the CUDA context
+    void *cuda_context_handle;
+
+    // Handle to the CUDA device
+    int cuda_device_handle;
+
+    // Allocation memory type
+    umf_usm_memory_type_t memory_type;
+
+    // Allocation flags for cuMemHostAlloc/cuMemAllocManaged
+    unsigned int alloc_flags;
 } umf_cuda_memory_provider_params_t;
 
 typedef struct cu_ops_t {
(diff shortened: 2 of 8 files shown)


include/umf/providers/provider_cuda.h
```diff
@@ -53,6 +53,22 @@ umf_result_t umfCUDAMemoryProviderParamsSetMemoryType(
     umf_cuda_memory_provider_params_handle_t hParams,
     umf_usm_memory_type_t memoryType);
 
+/// @brief  Set the host allocation flags in the parameters struct.
+/// @param  hParams handle to the parameters of the CUDA Memory Provider.
+/// @param  flags combination of CU_MEMHOSTALLOC_PORTABLE,
+///         CU_MEMHOSTALLOC_DEVICEMAP and CU_MEMHOSTALLOC_WRITECOMBINED flags
+/// @return UMF_RESULT_SUCCESS on success or appropriate error code on failure.
+umf_result_t umfCUDAMemoryProviderParamsSetHostAllocFlags(
```
igchor, line 61: Why do we need 2 separate functions? A provider can allocate either host or managed memory, never both, right?
If It's only for checking if the flags are valid, we can move validation to initialize.

bratpiorka: merged, done

src/provider/provider_cuda.c
```diff
@@ -294,6 +327,24 @@ static umf_result_t cu_memory_provider_initialize(void *params,
...
+            return UMF_RESULT_ERROR_INVALID_ARGUMENT;
+        }
+    } else if (cu_params->memory_type == UMF_MEMORY_TYPE_HOST) {
+        if (cu_params->alloc_flags &
+            ~(CU_MEMHOSTALLOC_PORTABLE | CU_MEMHOSTALLOC_DEVICEMAP |
+              CU_MEMHOSTALLOC_WRITECOMBINED)) {
+            LOG_ERR("Invalid host allocation flags");
+            return UMF_RESULT_ERROR_INVALID_ARGUMENT;
+        }
+    }
```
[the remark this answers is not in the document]
ldrumm: I agree.  We shouldn't try and catch invalid combinations of flags since CUDA will always be more up to date than us and will do the error checking for us. As long as errors from the CUDA runtime are correctly propagated I think we should just pass through

src/provider/provider_cuda.c
```diff
@@ -196,12 +216,12 @@ static void init_cu_global_state(void) {
         utils_get_symbol_addr(0, "cuIpcCloseMemHandle", lib_name);
 
     if (!g_cu_ops.cuMemGetAllocationGranularity || !g_cu_ops.cuMemAlloc ||
-        !g_cu_ops.cuMemAllocHost || !g_cu_ops.cuMemAllocManaged ||
-        !g_cu_ops.cuMemFree || !g_cu_ops.cuMemFreeHost ||
-        !g_cu_ops.cuGetErrorName || !g_cu_ops.cuGetErrorString ||
-        !g_cu_ops.cuCtxGetCurrent || !g_cu_ops.cuCtxSetCurrent ||
-        !g_cu_ops.cuIpcGetMemHandle || !g_cu_ops.cuIpcOpenMemHandle ||
-        !g_cu_ops.cuIpcCloseMemHandle) {
+        !g_cu_ops.cuMemAllocHost || !g_cu_ops.cuMemHostAlloc ||
```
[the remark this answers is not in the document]
bratpiorka: removed

src/provider/provider_cuda.c
```diff
@@ -294,6 +324,16 @@ static umf_result_t cu_memory_provider_initialize(void *params,
...
     }
 
+    // If the memory type is shared (CUDA managed), the allocation flags must
+    // be set.
+    // NOTE: we do not check here if the flags are valid - this will be done in
+    // CUDA.
+    if (cu_params->memory_type == UMF_MEMORY_TYPE_SHARED &&
+        cu_params->alloc_flags == 0) {
+        // the default setting is CU_MEM_ATTACH_GLOBAL
+        cu_params->alloc_flags = CU_MEM_ATTACH_GLOBAL;
```
igchor, line 334: Hm, I think it would be better not to modify the params. In theory, following scenario is valid, right?

```
auto params = umfCUDAMemoryProviderParamsCreate();
params.setMemType(UMF_MEMORY_TYPE_SHARED);

provider1 = umfMemoryProviderCreate(..., params);

params.setMemType(UMF_MEMORY_TYPE_DEVICE);
provider2 = umfMemoryProviderCreate(..., params);
```

In this case, the alloc_flags passed for the second provider will be different than what user would expect.

vinser52: Good catch. We should do this for the `cu_provider->alloc_flags`.

bratpiorka: fixed - I modified only the provider copy of the parameters, not the user. BTW we should change the umfMemoryProviderCreate and make params a const void*

bratpiorka:
@ldrumm @igchor @vinser52 all tests pass, ready to re-review

igchor (approved):

vinser52 (approved):

ldrumm:
Apologies for the delay. I should be able to properly review this later today.

src/provider/provider_cuda.c
```diff
@@ -89,21 +97,30 @@ typedef struct cu_memory_provider_t {
...
 
 // CUDA Memory Provider settings struct
 typedef struct umf_cuda_memory_provider_params_t {
-    void *cuda_context_handle;         ///< Handle to the CUDA context
-    int cuda_device_handle;            ///< Handle to the CUDA device
-    umf_usm_memory_type_t memory_type; ///< Allocation memory type
+    // Handle to the CUDA context
+    void *cuda_context_handle;
+
+    // Handle to the CUDA device
```
ldrumm, line 108: These are no longer doxygen style comments. Is that intended since this is implementation and non-public?

ldrumm (commented):
I'm happy with this implementation of this feature as it pertains to CUDA, but do we need to better generalize for other accelerator APIs that also takes flags for allocations? CUDA is one, but HIP supports a different set of flags, and Intel OpenCL extensions also extend allocation to take hints.

As I discussed [here](https://github.com/oneapi-src/unified-runtime/pull/2480#issuecomment-2636466900) I've already got work in progress that does similar things for all accelerator targets, so if UMF can provide a generic way to do this, then I'd prefer that

Please note, I'm still not to be considered competent with understanding how UMF works so feel free to tell me I'm misunderstanding

igchor:
> I'm happy with this implementation of this feature as it pertains to CUDA, but do we need to better generalize for other accelerator APIs that also takes flags for allocations? CUDA is one, but HIP supports a different set of flags, and Intel OpenCL extensions also extend allocation to take hints.
> 
> As I discussed [here](https://github.com/oneapi-src/unified-runtime/pull/2480#issuecomment-2636466900) I've already got work in progress that does similar things for all accelerator targets, so if UMF can provide a generic way to do this, then I'd prefer that
> 
> Please note, I'm still not to be considered competent with understanding how UMF works so feel free to tell me I'm misunderstanding

UMF doesn't allow flags/hints to be passed to pool allocation APIs. This is because those flags/hints only make sense at the page-granularity level. If we would, for example, allow `umfPoolMalloc(1024, READ_MOSTLY_HINT)`, internally UMF would have to apply the hint to the entire memory page which would impact other, unrelated allocations.

Our current thinking is that the upper level (UR or SYCL) should decide what to do for every specific flag. For some flags (ones that are used often), it might make sense to create dedicated pools (for which entire memory is allocated with those flags). For others (or when the allocation size is big), it might be better to not do any pooling (just pass the request directly to the driver). UR spec gives as quite a lot of flexibility in how we can implement this in the adapters.

src/provider/provider_cuda.c
bratpiorka: Yes, this code was moved from the public include files to this location when we updated the UMF API to use setters instead of directly exposign the parameter structure. So this change is just a cleanup

Issue #926, opened 2024-11-22
L0 provider cannot find L0 symbols when dlopen is used.


igchor:
Problematic scenario:
L0 UR adapter uses L0 provider from UMF. L0 UR adapter is being dlopened by the loader (which application links to). When application itself also links with UMF and uses `umfPoolGetMemoryProvider` (or perahps any UMF symbol?) then L0 provider cannot find symbols.

How to reproduce:
```
git clone https://github.com/igchor/umf_repro
cd umf_repro
mkdir build
cd build
# uncomment include_directories and link_directories from CMakeLists.txt and change them to point to a proper directory
cmake ..
make
LD_DEBUG=files LD_LIBRARY_PATH=. ./test
```

Output:
```
...

   1159394:     /home/igchor/unified-memory-framework/build/lib/libumf.so.0: error: symbol lookup error: undefined symbol: zeMemAllocHost (fatal)
   1159394:     /home/igchor/unified-memory-framework/build/lib/libumf.so.0: error: symbol lookup error: undefined symbol: zeMemAllocDevice (fatal)
   1159394:     /home/igchor/unified-memory-framework/build/lib/libumf.so.0: error: symbol lookup error: undefined symbol: zeMemAllocShared (fatal)
   1159394:     /home/igchor/unified-memory-framework/build/lib/libumf.so.0: error: symbol lookup error: undefined symbol: zeMemFree (fatal)
   1159394:     /home/igchor/unified-memory-framework/build/lib/libumf.so.0: error: symbol lookup error: undefined symbol: zeMemGetIpcHandle (fatal)
   1159394:     /home/igchor/unified-memory-framework/build/lib/libumf.so.0: error: symbol lookup error: undefined symbol: zeMemPutIpcHandle (fatal)
   1159394:     /home/igchor/unified-memory-framework/build/lib/libumf.so.0: error: symbol lookup error: undefined symbol: zeMemOpenIpcHandle (fatal)
   1159394:     /home/igchor/unified-memory-framework/build/lib/libumf.so.0: error: symbol lookup error: undefined symbol: zeMemCloseIpcHandle (fatal)
   1159394:     /home/igchor/unified-memory-framework/build/lib/libumf.so.0: error: symbol lookup error: undefined symbol: zeContextMakeMemoryResident (fatal)
[...]

PR #1086, opened 2025-02-07
Increase refcount to ze_loader/CUDA libraries when Level Zero/CUDA providers are used

vinser52:
### Description
Prior to this PR, the Level Zero provider did not `dlopen(“libze_loader.so”)` at all. We suppose that UMF’s client, who uses Level Zero provider, loaded `libze_loader.so` into the process. But what if there are two clients in the process:

1. The first client loads `libze_loader.so` into the process and uses the Level Zero provider. In that case, UMF inits Level Zero symbols via `dlsym`. Then the first client destroys the Level Zero Memory provider and unloads the `libze_loader.so`.
2. Then the second client loads `libze_loader.so` into the process and uses the Level Zero provider. But our current implementation does not catch such situation and Level Zero symbols are considered initialized, but the `dlsym` should be done again.

This PR fixes that. The UMF acquires handle to the `libze_loader.so` via `dlopen`. There is a `RTLD_NOLOAD` flag that tells `dlopen` not to load the library and succeed only if the library is already loaded. It allows to increase the refcount to the `libze_loader.so` and not unload it when the first client calls `dlclose`.

This PR should fix the #926.

Fixes: https://github.com/intel/llvm/issues/16944 (confirmed by @ldorau)

### Checklist

- [x] All tests pass locally

### DIFF
--- a/src/provider/provider_cuda.c
+++ b/src/provider/provider_cuda.c
@@ -12,7 +12,18 @@
 #include <umf.h>
 #include <umf/providers/provider_cuda.h>
 
+#include "provider_cuda_internal.h"
+#include "utils_load_library.h"
 #include "utils_log.h"
+
+static void *cu_lib_handle = NULL;
+
+void fini_cu_global_state(void) {
+    if (cu_lib_handle) {
+        utils_close_library(cu_lib_handle);
+        cu_lib_handle = NULL;
+    }
+}
 
 #if defined(UMF_NO_CUDA_PROVIDER)
 
@@ -88,7 +99,6 @@ umf_memory_provider_ops_t *umfCUDAMemoryProviderOps(void) {
 #include "utils_assert.h"
 #include "utils_common.h"
 #include "utils_concurrency.h"
-#include "utils_load_library.h"
 #include "utils_log.h"
 #include "utils_sanitizers.h"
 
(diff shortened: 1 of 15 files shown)


bratpiorka:
I know this is probably not easy, but could we add the simplest test for this case?

src/provider/provider_level_zero.c
```diff
@@ -207,32 +217,41 @@ static void init_ze_global_state(void) {
...
 #endif
-    // check if Level Zero shared library is already loaded
-    // we pass 0 as a handle to search the global symbol table
+    // The Level Zero shared library should be already loaded by the user
+    // of the Level Zero provider. UMF just want to reuse it
+    // and increase the reference count to the the Level Zero shared library.
+    void *lib_handle =
+        utils_open_library(lib_name, UMF_UTIL_OPEN_LIBRARY_NO_LOAD);
+    if (!lib_handle) {
+        LOG_ERR("Failed to open Level Zero shared library");
```
PatKamin, line 226: This should be a fatal error as `Init_ze_global_state_failed == true` results in a fatal in `ze_memory_provider_initialize()`.

src/provider/provider_cuda.c
```diff
@@ -163,37 +173,45 @@ static void init_cu_global_state(void) {
 #else
     const char *lib_name = "libcuda.so";
 #endif
-    // check if CUDA shared library is already loaded
-    // we pass 0 as a handle to search the global symbol table
+    // The CUDA shared library should be already loaded by the user
+    // of the CUDA provider. UMF just want to reuse it
+    // and increase the reference count to the the CUDA shared library.
```
PatKamin, line 178: nit: double "the"

ldorau:
@vinser52 @bratpiorka @pbalcer This PR fixes https://github.com/intel/llvm/issues/16944

src/provider/provider_level_zero.c
```diff
@@ -207,32 +217,41 @@ static void init_ze_global_state(void) {
 #else
     const char *lib_name = "libze_loader.so";
 #endif
-    // check if Level Zero shared library is already loaded
-    // we pass 0 as a handle to search the global symbol table
+    // The Level Zero shared library should be already loaded by the user
+    // of the Level Zero provider. UMF just want to reuse it
+    // and increase the reference count to the the Level Zero shared library.
```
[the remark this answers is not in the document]
vinser52: Done

src/provider/provider_level_zero.c
vinser52: Done

src/provider/provider_cuda.c
vinser52: Done

vinser52:
> I know this is probably not easy, but could we add the simplest test for this case?

Ok, new tests have been added. Actually, I split the `umf-provider_level_zero_dlopen` and `umf-provider_cuda_dlopen` to the `umf-provider_level_zero_dlopen_global`/`umf-provider_level_zero_dlopen_local` and `umf-provider_cuda_dlopen_global`/`umf-provider_cuda_dlopen_local` tests correspondingly.

Without my changes the `umf-provider_level_zero_dlopen_local` and `umf-provider_cuda_dlopen_local` tests are failed.

vinser52:
> @vinser52 @bratpiorka @pbalcer This PR fixes [intel/llvm#16944](https://github.com/intel/llvm/issues/16944)

so this PR is ready for the review.

ldorau (approved):

ldorau:
@lukaszstolarczuk @bratpiorka @pbalcer Hanging on Windows CUDA CI builds ...

igchor (approved):

src/provider/provider_level_zero.c
```diff
@@ -244,7 +263,10 @@ static void init_ze_global_state(void) {
         // starting from Level Zero 1.6
         LOG_FATAL("Required Level Zero symbols not found.");
         Init_ze_global_state_failed = true;
+        utils_close_library(lib_handle);
```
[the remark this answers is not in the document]
ldorau: Right.

PR #1090, opened 2025-02-11
disable umf mt bench in GPU workflow

bratpiorka:
disable the UMF MT bench in the GPU workflow as it crashes on GitHub machines. 

see https://github.com/oneapi-src/unified-memory-framework/actions/runs/13255586163/job/37001810256 

it will only be built


lukaszstolarczuk:
side note, do we know why it fails? do we have an issue? ;)

lukaszstolarczuk (approved):

PR #1101, opened 2025-02-14
Call zeInitDrivers in L0 provider

igchor:
This PR is based on: https://github.com/oneapi-src/unified-memory-framework/pull/1086/files

And fixes problems in https://github.com/oneapi-src/unified-runtime/pull/2667

### DIFF
--- a/src/provider/provider_level_zero.c
+++ b/src/provider/provider_level_zero.c
@@ -14,6 +14,7 @@
 #include <umf/memory_provider_ops.h>
 #include <umf/providers/provider_level_zero.h>
 
+#include "base_alloc_global.h"
 #include "provider_level_zero_internal.h"
 #include "utils_load_library.h"
 #include "utils_log.h"
@@ -111,7 +112,6 @@ umf_memory_provider_ops_t *umfLevelZeroMemoryProviderOps(void) {
 
 #else // !defined(UMF_NO_LEVEL_ZERO_PROVIDER)
 
-#include "base_alloc_global.h"
 #include "libumf.h"
 #include "utils_assert.h"
 #include "utils_common.h"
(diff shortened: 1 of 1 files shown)


igchor:
The tests are failing on CI most likely because we have an older L0 loader version installed and zeInitDrivers might not work. I'm not sure if we need to support older version so perhas we can just update the runners?

If we do need to support older version, we will probably need to implement a fallback to zeInit (which is supported by older loaders but deprecated now).

src/libumf.c
```diff
@@ -63,12 +63,16 @@ void umfTearDown(void) {
 #endif
         umfIpcCacheGlobalTearDown();
 
+        fini_ze_global_state();
+        fini_cu_global_state();
+        LOG_DEBUG("UMF library finalized");
+
```
vinser52, line 69: Why do we need it earlier?

src/provider/provider_level_zero.c
```diff
@@ -267,6 +297,15 @@ static void init_ze_global_state(void) {
         return;
     }
     ze_lib_handle = lib_handle;
+
+    umf_result_t result = ze_init_drivers();
```
vinser52, line 301: Hmmm, I am thinking about who is responsible for calling `zeInitDrivers`.
I suppose that the client who uses UMF and Level Zero is responsible for that. At the moment, the client calls UMF to create Level Zero it should already create the context and get the device, so the driver should be initialized.

vinser52:
> The tests are failing on CI most likely because we have an older L0 loader version installed and zeInitDrivers might not work. I'm not sure if we need to support older version so perhas we can just update the runners?
> 
> If we do need to support older version, we will probably need to implement a fallback to zeInit (which is supported by older loaders but deprecated now).

So, I suppose it should be the client's responsibility to call `zeInit`/`zeInitDrivers`, as we have today in the tests.

src/provider/provider_level_zero.c
bratpiorka: I agree with @vinser52

igchor: As discussed, UMF needs to call init function, even if the client called it already.

src/libumf.c
igchor: Reverted, I need that to be able to free memory in fini_ze_global_state, but we don't actually need to keep the driver handles so it's not necessary.

src/provider/provider_level_zero.c
```diff
@@ -211,6 +211,49 @@ static umf_result_t ze2umf_result(ze_result_t result) {
...
+    *(void **)&zeInitDriversFunc =
+        utils_get_symbol_addr(lib_handle, "zeInitDrivers", lib_name);
+    if (!zeInitDriversFunc) {
+        return UMF_RESULT_ERROR_DEPENDENCY_UNAVAILABLE;
+    }
+
+    ze_init_driver_type_desc_t desc = {
+        .stype = ZE_STRUCTURE_TYPE_INIT_DRIVER_TYPE_DESC,
+        .pNext = NULL,
+        .flags = UINT32_MAX};
```
nrspruit, line 226: If you want only GPU Drivers, you need to set this to _GPU otherwise you will get back GPUs and NPUs.

igchor: Done.

nrspruit (approved):

kswiecicki (approved):

.github/workflows/reusable_basic.yml
```diff
@@ -443,67 +443,67 @@ jobs:
...
-    - name: Test UMF installation and uninstallation
-      run: >
-        python3 ${{github.workspace}}/test/test_installation.py
-        --build-dir ${{env.BUILD_DIR}}
-        --install-dir ${{env.INSTL_DIR}}
-        --build-type ${{env.BUILD_TYPE}}
-        --proxy
-        --umf-version ${{env.UMF_VERSION}}
-        --shared-library
+  # macos-build:
```
vinser52, line 446: Why is  MacOS CI disabled?

ldorau: It should not be disabled here.

ldorau (changes requested):

src/provider/provider_level_zero.c
```diff
@@ -266,6 +309,19 @@ static void init_ze_global_state(void) {
         utils_close_library(lib_handle);
         return;
     }
+
+    if (ze_init_drivers(lib_handle, lib_name) != UMF_RESULT_SUCCESS) {
+        LOG_INFO("Initializing Level Zero through zeInitDrivers failed. "
+                 "Falling back to zeInit.");
```
vinser52, line 315: Should we fall back to the `zeInit` only if `zeInitDrivers` is unavailable?
E.g. should we call `zeInit` if `zeInitDrivers` is available but returns some error?

src/provider/provider_level_zero.c
```diff
@@ -211,6 +211,49 @@ static umf_result_t ze2umf_result(ze_result_t result) {
...
+    *(void **)&zeInitDriversFunc =
+        utils_get_symbol_addr(lib_handle, "zeInitDrivers", lib_name);
+    if (!zeInitDriversFunc) {
+        return UMF_RESULT_ERROR_DEPENDENCY_UNAVAILABLE;
+    }
+
+    ze_init_driver_type_desc_t desc = {
+        .stype = ZE_STRUCTURE_TYPE_INIT_DRIVER_TYPE_DESC,
+        .pNext = NULL,
+        .flags = ZE_INIT_DRIVER_TYPE_FLAG_GPU};
```
vinser52, line 226: Why only GPUs?
That is why I was against calling `zeInit` or `zeInitDrivers` from UMF's code. It is the client's responsibility to initiate L0 properly.

igchor: If we don't call zeInit/zeInitDrivers, the entire provider won't work. As for what specifically we are initializing - GPU is currently the only thing that makes sense from the UMF perspective. If there will be some other device types in the future that we want to support we can extend params to allow user to specify that.

.github/workflows/reusable_basic.yml
igchor: Done.

src/provider/provider_level_zero.c
lplewa: I do not know how user can create level zero provider without zeInit on their side.  They need to provoide context/device handle, which need level zero inited?

vinser52: +1 for what @lplewa wrote.

@igchor are there any issues when adapter statically linked with libze_loader?

PR #1106, opened 2025-02-19
coverity fixes

bratpiorka:
** CID 475100:  Concurrent data access violations  (MISSING_LOCK)
/src/pool/pool_disjoint.c: 496 in disjoint_pool_print_stats()

** CID 475099:  Concurrent data access violations  (MISSING_LOCK)
/src/ipc_cache.c: 150 in umfIpcOpenedCacheDestroy()

** CID 475098:  Concurrent data access violations  (MISSING_LOCK)
/src/pool/pool_disjoint.c: 496 in disjoint_pool_print_stats()

** CID 475097:  Concurrent data access violations  (MISSING_LOCK)
/test/pools/disjoint_pool.cpp: 145 in test_internals_Test::TestBody()()


ldorau (approved):

src/ipc_cache.c
```diff
@@ -144,6 +144,8 @@ umfIpcOpenedCacheCreate(ipc_opened_cache_eviction_cb_t eviction_cb) {
 
 void umfIpcOpenedCacheDestroy(ipc_opened_cache_handle_t cache) {
     ipc_opened_cache_entry_t *entry, *tmp;
+    utils_mutex_lock(&(cache->global->cache_lock));
```
vinser52, line 147: I do not think that lock is needed here because destructor is always called at the end and there should not be any concurrent threads. I think this Coverity issue should be marked as a false-positive.

vinser52 (changes requested):

src/ipc_cache.c
bratpiorka: @vinser52 do we need to set cur_size here? Maybe we could remove this line so Coverity would not report any error

cache->global->cur_size -= 1;

test/pools/disjoint_pool.cpp
```diff
@@ -138,6 +138,8 @@ TEST_F(test, internals) {
     EXPECT_NE(bucket, nullptr);
 
     // check bucket stats
+    // NOTE: we have to lock the bucket to properly access its internals
+    utils_mutex_lock(&bucket->bucket_lock);
```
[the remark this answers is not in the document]
bratpiorka: right, reverted

src/ipc_cache.c
vinser52: ahh, wait, this is not a global cache destructor, this function is called every time a particular pool/ipc_handler is destroyed. If we have two or more ipc_handlers that are destroyed in parallel then this function is called concurrently. So looks like we need the lock here. We need a test for such flow.

Give me some time to add such a test in `ipcFixtures.hpp`

vinser52 (approved):

PR #1108, opened 2025-02-19
Benchmarks UR repo move

lukaszstolarczuk:
Update location of benchmarking scripts (prev. in UR repo, now in SYCL repo, unified-runtime dir).

### Checklist
- [x] All tests pass locally
- [x] CI workflows, not executed per PR (e.g. Nightly), execute properly // https://github.com/oneapi-src/unified-memory-framework/actions/runs/13410329249/job/37458859547

### DIFF
--- a/.github/workflows/reusable_benchmarks.yml
+++ b/.github/workflows/reusable_benchmarks.yml
@@ -1,6 +1,5 @@
-# Executes benchmarks implemented in this repository
-# using scripts for benchmark results visualization,
-# which are downloaded from Unified Runtime repository. 
+# Executes benchmarks implemented in this repository using scripts
+# for results visualization from intel/llvm (unified-runtime dir).
 name: Benchmarks
 
 on:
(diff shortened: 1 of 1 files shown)


.github/workflows/reusable_benchmarks.yml
```diff
@@ -130,19 +129,21 @@ jobs:
...
         --umf ${{env.BUILD_DIR}}
+        --compare baseline
         ${{ inputs.upload_report && '--output-html' || '' }}
+        ${{ inputs.pr_no != 0 && '--output-markdown' || '' }}
         ${{ inputs.bench_script_params }}
 
     - name: Print benchmark results
       if: ${{ always() }}
-      run: cat ${{env.BUILD_DIR}}/benchmark_results.md
+      run: cat ${{env.BUILD_DIR}}/benchmark_results.md || true
```
bratpiorka, line 146: IMO it is better to fail the workflow than to silently ignore any errors

lukaszstolarczuk: there's a legit use cases, where we only have html output (and it's pushed to web), so no markdown in such case will be found.

lplewa: If we have HTML output this step should be disabled by the if clause?

lplewa (approved):

bratpiorka (approved):

.github/workflows/reusable_benchmarks.yml
kswiecicki: `inputs.pr_no != 0` is tied to markdown output option, we could use that.

lukaszstolarczuk: Done.

PR #1111, opened 2025-02-19
Fix data race in the umfIpcOpenedCacheDestroy function

vinser52:
### Description
This PR fixes data race (found by Coverity) in the umfIpcOpenedCacheDestroy function.

### Checklist

- [x] All tests pass locally

### DIFF
--- a/src/provider/provider_tracking.c
+++ b/src/provider/provider_tracking.c
@@ -473,10 +473,6 @@ static void trackingFinalize(void *provider) {
 
     critnib_delete(p->ipcCache);
 
-#ifndef NDEBUG
-    check_if_tracker_is_empty(p->hTracker, p->pool);
-#endif /* NDEBUG */
-
     umf_ba_global_free(provider);
 }
 
--- a/test/ipcFixtures.hpp
+++ b/test/ipcFixtures.hpp
@@ -593,4 +593,73 @@ TEST_P(umfIpcTest, ConcurrentOpenCloseHandles) {
     EXPECT_EQ(stat.openCount, stat.closeCount);
 }
 
+TEST_P(umfIpcTest, ConcurrentDestroyIpcHandlers) {
+    constexpr size_t SIZE = 100;
+    constexpr size_t NUM_ALLOCS = 100;
+    constexpr size_t NUM_POOLS = 10;
+    void *ptrs[NUM_ALLOCS];
+    void *openedPtrs[NUM_POOLS][NUM_ALLOCS];
+    std::vector<umf::pool_unique_handle_t> consumerPools;
+    umf::pool_unique_handle_t producerPool = makePool();
+    ASSERT_NE(producerPool.get(), nullptr);
+
+    for (size_t i = 0; i < NUM_POOLS; ++i) {
+        consumerPools.push_back(makePool());
+    }
+
+    for (size_t i = 0; i < NUM_ALLOCS; ++i) {
+        void *ptr = umfPoolMalloc(producerPool.get(), SIZE);
+        ASSERT_NE(ptr, nullptr);
+        ptrs[i] = ptr;
+    }
+
+    for (size_t i = 0; i < NUM_ALLOCS; ++i) {
+        umf_ipc_handle_t ipcHandle = nullptr;
+        size_t handleSize = 0;
+        umf_result_t ret = umfGetIPCHandle(ptrs[i], &ipcHandle, &handleSize);
+        ASSERT_EQ(ret, UMF_RESULT_SUCCESS);
+
+        for (size_t poolId = 0; poolId < NUM_POOLS; poolId++) {
+            void *ptr = nullptr;
+            umf_ipc_handler_handle_t ipcHandler = nullptr;
+            ret =
+                umfPoolGetIPCHandler(consumerPools[poolId].get(), &ipcHandler);
+            ASSERT_EQ(ret, UMF_RESULT_SUCCESS);
+            ASSERT_NE(ipcHandler, nullptr);
+
+            ret = umfOpenIPCHandle(ipcHandler, ipcHandle, &ptr);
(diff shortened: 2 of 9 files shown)


lplewa (approved):

src/provider/provider_tracking.c
```diff
@@ -473,10 +473,6 @@ static void trackingFinalize(void *provider) {
 
     critnib_delete(p->ipcCache);
 
-#ifndef NDEBUG
-    check_if_tracker_is_empty(p->hTracker, p->pool);
-#endif /* NDEBUG */
-
```
vinser52, line 479: @ldorau I was required to delete this check because TSAN and Valgrind reported a lot of data races when several pools are destroyed concurrently. It looks like we never tested concurrent pools destroy.

ldorau: I insist on keeping it under additional ifdef not defined in our CI (maybe `UMF_DEBUG_MODE`?) - it can be very useful during debugging.

bratpiorka: why not, if this is the only place? 
these checks could help us catch bugs very early

test/supp/drd-umf_test-provider_devdax_memory_ipc.supp
```diff
@@ -6,3 +6,20 @@
    fun:umfOpenIPCHandle
    ...
 }
+
+{
+   False-positive ConflictingAccess in jemalloc
+   drd:ConflictingAccess
+   fun:atomic_*
```
bratpiorka, line 13: could we make these suppressions less generic?

src/provider/provider_tracking.c
vinser52: @ldorau I introduced additional macros to enable this check: `UMF_POOL_CHECK_TRACKER_IS_EMPTY`

test/supp/drd-umf_test-provider_devdax_memory_ipc.supp
vinser52: there are different atomic operations (`load`/`store`) for different data types (`b`/`n`/`zu` etc.). So valgrind reports the following functions:
`atomic_load_b`
`atomic_store_b`
`atomic_load_zu`
`atomic_store_zu`

etc.

vinser52: And as you can see it suppresses not all calls to the `atomic_*` functions but only those that was called by the jemalloc. Btw, as I understand, the `atomic_*` functions are from the jemalloc library as well

bratpiorka: ok

src/provider/provider_tracking.c
ldorau: Thanks. It's OK for me.

ldorau (approved):

src/provider/provider_tracking.c
ldorau: @bratpiorka ?

kswiecicki (approved):

test/ipcFixtures.hpp
```diff
@@ -593,4 +593,73 @@ TEST_P(umfIpcTest, ConcurrentOpenCloseHandles) {
     EXPECT_EQ(stat.openCount, stat.closeCount);
 }
 
+TEST_P(umfIpcTest, ConcurrentDestroyIpcHandlers) {
+    constexpr size_t SIZE = 100;
+    constexpr size_t NUM_ALLOCS = 100;
+    constexpr size_t NUM_POOLS = 10;
+    void *ptrs[NUM_ALLOCS];
+    void *openedPtrs[NUM_POOLS][NUM_ALLOCS];
```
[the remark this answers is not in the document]
vinser52: Done, now variables in the test use camel case and constants use capital letters.

src/provider/provider_tracking.c
lplewa: Imho is better to remove this check. It will be dead code, as no one will enable this macro

vinser52: > Is this code wrong, or this is only false positive? If it is false positive, than this is an issue that when we took critnib from PMDK we did not imported implementation of the "VALGRIND_HG_DRD_DISABLE_CHECKING" macros, which are used in critnib to fix false positives.

I am not 100% sure but I think it is a real issue rather than false-postive. Btw the same flow with critnib is used by the disjoint pool, see these issues for reference: #1114,  #1115.

The test do the following:
* Thread 1 destroys Pool A and as a result, the `check_if_tracker_is_empty` is called and it iterates over the records in the memory tracker (which `critnib` map).
* Thread 2 destroys Pool B and the pool deallocates memory blocks (that were cached) and these blocks are removed from the tracker.
* As a result Thread 2 deallocates from the tracker only records that corresponds to Pool B. But Thread 1 touches all entries in the tracker. And it is the problem because thread 1 might read the entry which is in a process of deletion by thread 2

vinser52: > why not, if this is the only place?
> these checks could help us catch bugs very early

@bratpiorka I do not want to introduce locks in other places just to support this debugging functionality. And I do not think that this debugging check is really needed because we are checking if the tracker is empty when the tracker itself is destructed (when libumf is unloaded). And this removed check is called when the pool is destroyed and checks that the tracker does not contain records from the corresponding pool. But without checking at the pool destruction stage we will be able to catch the same issue when the tracker itself is destructed.

bratpiorka: @vinser52 ok if we check the tracker at libumf unload then we could remove it from here

vinser52: > @vinser52 ok if we check the tracker at libumf unload then we could remove it from here

Done

bratpiorka (approved):

PR #1112, opened 2025-02-19
Benchmark improvements

lplewa:
Fix incorrect bucked size, causing all allocating going to single bucket

### DIFF
--- a/benchmark/benchmark_umf.hpp
+++ b/benchmark/benchmark_umf.hpp
@@ -199,7 +199,7 @@ struct disjoint_pool : public pool_interface<Provider> {
             return {nullptr, [](void *) {}};
         }
 
-        ret = umfDisjointPoolParamsSetMinBucketSize(raw_params, 4096);
+        ret = umfDisjointPoolParamsSetMinBucketSize(raw_params, 8);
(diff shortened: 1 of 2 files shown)


bratpiorka:
Why do we need these changes? What are the effects - performance improvements, better stability? 
Please fill in the PR description and provide any relevant data for this PR.

benchmark/benchmark_umf.hpp
```diff
@@ -199,7 +199,7 @@ struct disjoint_pool : public pool_interface<Provider> {
             return {nullptr, [](void *) {}};
         }
 
-        ret = umfDisjointPoolParamsSetMinBucketSize(raw_params, 4096);
+        ret = umfDisjointPoolParamsSetMinBucketSize(raw_params, 8);
```
bratpiorka, line 202: why?

bratpiorka (changes requested):

benchmark/benchmark_umf.hpp
lplewa: So we are using more then 1 bucket in benchmarks

bratpiorka (approved):

kswiecicki (approved):

PR #1073, opened 2025-02-04
[CTL] Add CTL by handle

KFilipek:
### Description
This PR introduces OS memory provider initial CTL support.

Fixes: #1062

### Checklist

- [x] Code compiles without errors locally
- [x] All tests pass locally
- [x] CI workflows execute properly

PR #1123, opened 2025-02-21
[CTL] Add variadic functions

### LINKED ISSUE #1085: CTL - add variadic arguments API
# Add variadic arguments API
This issue introduces generic CTL API for the following functions:
```cpp
umfCtlGet(const char* name, ...);
umfCtlSet(const char* name, ...);
umfCtlexec(const char* name, ...);
```
Current API contains explicit arguments like:
```cpp
int umfCtlGet(const char *name, void *ctx, void *arg);
```

KFilipek:
# [CTL] Add variadic functions

### Description
Since this PR all the CTL API functions (umfCtlSet/Get/Exec)
will be defined as a variadic functions to provide more elastic
approach in future development.

### Checklist
Depends on #1073
Resolves #1085

- [x] All tests pass locally


include/umf/base.h
```diff
@@ -63,29 +63,23 @@ typedef enum umf_ctl_query_type {
 ///
 /// @brief Get value of a specified attribute at the given name.
 /// @param name name of an attribute to be retrieved
-/// @param ctx pointer to the pool or the provider
```
PatKamin: Is the information on used params documented somewhere after these params description removal? I'm not sure if this is a good idea to place variadic arguments in this API. This approach would certainly require a very good documentation clearly stating how arguments are currently parsed.

What could be future arguments? Perhaps you could simply add unused `void *extra_args` for handling future ctl usage?

.github/workflows/reusable_compatibility.yml
```diff
@@ -110,7 +110,7 @@ jobs:
       run: >
         UMF_LOG="level:warning;flush:debug;output:stderr;pid:no" 
         LD_LIBRARY_PATH=${{github.workspace}}/latest_version/build/lib/ 
-        ctest --output-on-failure
+        ctest --output-on-failure -E "umf-mempolicy"
```
PatKamin, line 113: Why this test has to be disabled? If this is because of a compatibility break, we could merge PRs with a failing compatibility tests workflow since this PR until the next tag. After all, this is a change that breaks compatibility on purpose.

Otherwise, a clear TODO to restore this test after the next tag is created should be placed here.

include/umf/base.h
PatKamin: I agree on adding documentation later, after more of CTL implementation is done.

`extra_args` would be a replacement for variadic args that would allow to have a clear Doxygen comment on possible args. But if you would document them later anyway, it's better to stick with variadic args.

.github/workflows/reusable_compatibility.yml
KFilipek: It's not a part of this PR, please focus just on latest commit.

src/memory_pool_internal.h
```diff
@@ -26,14 +26,16 @@ extern "C" {
...
-    umf_memory_pool_ops_t ops;
     umf_pool_create_flags_t flags;
 
     // Memory provider used by the pool.
     umf_memory_provider_handle_t provider;
 
     utils_mutex_t lock;
     void *tag;
+
+    // ops should be the last due to possible change size in the future
```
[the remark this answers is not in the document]
KFilipek: Please review latest commit

src/memory_provider_internal.h
```diff
@@ -21,8 +21,9 @@ extern "C" {
 #endif
 
 typedef struct umf_memory_provider_t {
-    umf_memory_provider_ops_t ops;
     void *provider_priv;
+    // ops should be the last due to possible change size in the future
```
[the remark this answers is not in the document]
KFilipek: Please review latest commit

PatKamin (approved):

src/libumf.def
```diff
@@ -119,6 +119,9 @@ EXPORTS
     umfScalablePoolParamsSetKeepAllMemory
 ; Added in UMF_0.11
     umfCUDAMemoryProviderParamsSetAllocFlags
+    umfCtlExec
```
bratpiorka, line 122: please move CTL functions to 0.12 namespace

bratpiorka: also, bump OPS structure version to 0.12

KFilipek: I've updated structures for pool and provider, and also def+map files.

lplewa (changes requested):
This PR changes nothing. 

variadic args idea, for CTL was about replaceing ctx (and potentaily other args) with wildcards in "name". So you can do something like this
"umf.pool.byHandle.%p.arena.%d.something", ctx, 1

PR #1126, opened 2025-02-21
Add concurrent tests for IPC Get/Put functions

vinser52:
### Description
Add one more concurrent test for IPC Get/Put flow.

### Checklist

- [x] All tests pass locally

### DIFF
--- a/test/ipcFixtures.hpp
+++ b/test/ipcFixtures.hpp
@@ -15,8 +15,10 @@
 #include <umf/memory_provider.h>
 #include <umf/pools/pool_proxy.h>
 
+#include <algorithm>
 #include <cstring>
 #include <numeric>
+#include <random>
 #include <tuple>
 
 class MemoryAccessor {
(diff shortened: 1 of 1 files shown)


test/ipcFixtures.hpp
```diff
@@ -522,6 +522,44 @@ TEST_P(umfIpcTest, ConcurrentGetPutHandles) {
...
+        void *ptr = umfPoolMalloc(pool.get(), ALLOC_SIZE);
+        EXPECT_NE(ptr, nullptr);
+        ptrs.push_back(ptr);
+    }
+
+    umf_test::syncthreads_barrier syncthreads(NTHREADS);
+
+    auto getPutHandlesFn = [&ptrs, &syncthreads](size_t) {
+        syncthreads();
+        for (void *ptr : ptrs) {
```
igchor, line 542: It might be useful to have one test case where ptrs are shuffled differently for each thread

igchor (approved):

bratpiorka (approved):

test/ipcFixtures.hpp
vinser52: Yeah, but my idea is to achieve high contention on the IPC cache when every thread opens the same IPC handle and tries to put it in the cache, but only one should succeed. If `ptrs` are shuffled then the probability of the contention is decreased.

vinser52: Done. So I implemented both version shuffled and not-shuffled

igchor (approved):

Issue #1125, opened 2025-02-21
fix deadlock in Disjoint Pool MT benchmark on GPU + windows


bratpiorka:
fix deadlock in Disjoint Pool MT benchmark on GPU + windows

bratpiorka:
fixed by #1151

Issue #1115, opened 2025-02-20
sporadic TSAN failures in Disjoint Pool tests


bratpiorka:
https://github.com/oneapi-src/unified-memory-framework/actions/runs/13424920686/job/37506526701?pr=1111

```
[ RUN      ] disjointPoolTests/umfPoolTest.multiThreadedpow2AlignedAlloc/0
==================
WARNING: ThreadSanitizer: data race (pid=12083)
  Atomic write of size 8 at 0x7f46d243dec8 by thread T10 (mutexes: write M0, write M1):
    #0 store /home/runner/work/unified-memory-framework/unified-memory-framework/src/critnib/critnib.c:151:5 (umf_test-disjoint_pool+0x15ccb2) (BuildId: ee6b9c89d5b3e992f449624bd41880e02396a84e)
    #1 critnib_remove /home/runner/work/unified-memory-framework/unified-memory-framework/src/critnib/critnib.c:4[69](https://github.com/oneapi-src/unified-memory-framework/actions/runs/13424920686/job/37506526701?pr=1111#step:7:70):5 (umf_test-disjoint_pool+0x15d281) (BuildId: ee6b9c89d5b3e992f449624bd41880e02396a84e)

```

bratpiorka:
another case

https://github.com/ldorau/unified-memory-framework/actions/runs/13492334927/job/37692708148

PR #998, opened 2024-12-17
Implement size limit for the cache of opened IPC handles

vinser52:
### Description
It is a follow-up of the #736. Add an environment variable that allows setting a limit for the size of the cache of opened IPC handles.

### Checklist

- [x] Code compiles without errors locally
- [x] All tests pass locally
- [x] CI workflows execute properly
- [x] New tests added, especially if they will fail without my changes
- [x] Extended the README/documentation

Issue #886, opened 2024-11-11
Valgrind and Thread Sanitizer (TSan) instrumentation


vinser52:
During the investigation of the CI valgrind failures of PR #736, the concern raised in my mind if we correctly instrumented atomic-store-with-release and atomic-load-with-acquire operations.

## Current instrumentation in UMF

In `utils_concurrency.h` there are two macros:
```c
#define utils_atomic_load_acquire(object, dest)                                \
    do {                                                                       \
        utils_annotate_acquire((void *)object);                                \
        __atomic_load(object, dest, memory_order_acquire);                     \
    } while (0)

#define utils_atomic_store_release(object, desired)                            \
    do {                                                                       \
        __atomic_store_n(object, desired, memory_order_release);               \
        utils_annotate_release((void *)object);                                \
    } while (0)
```

The `utils_annotate_acquire` and `utils_annotate_release` functions defined in the `utils_sanitizers.h` as follows:
```c
static inline void utils_annotate_acquire(void *ptr) {
#if __SANITIZE_THREAD__
    __tsan_acquire(ptr);
#elif UMF_VG_HELGRIND_ENABLED || UMF_VG_DRD_ENABLED
    ANNOTATE_HAPPENS_AFTER(ptr);
#else
    (void)ptr;
#endif
}

static inline void utils_annotate_release(void *ptr) {
#if __SANITIZE_THREAD__
    __tsan_release(ptr);
#elif UMF_VG_HELGRIND_ENABLED || UMF_VG_DRD_ENABLED
    ANNOTATE_HAPPENS_BEFORE(ptr);
#else
    (void)ptr;
#endif
}
```

## My understanding of Annotation Placement:

> Note: My understanding after searching in internet

1. For Valgrind (using `ANNOTATE_HAPPENS_BEFORE` and `ANNOTATE_HAPPENS_AFTER`):
  - `ANNOTATE_HAPPENS_BEFORE` should be placed immediately before the actual atomic-store-with-release.
  - `ANNOTATE_HAPPENS_AFTER` should be placed immediately after the actual atomic_load with acquire semantics.
3. For Thread Sanitizer (TSan) (using `__tsan_release` and `__tsan_acquire`):
  - `__tsan_release` should be placed immediately after the actual atomic_store with release semantics.
  - `__tsan_acquire` should be placed immediately after the actual atomic_load with acquire semantics.

igchor:
[...]

PR #1127, opened 2025-02-21
Cleanup atomics and fix deadlock in DP bucket_can_pool() 

bratpiorka:
Cleanup atomics and replace while(true) loop in Disjoint Pool bucket_can_pool() with a pair of atomic add/sub.

fix for https://github.com/oneapi-src/unified-memory-framework/issues/1125 and https://github.com/oneapi-src/unified-memory-framework/issues/1115

This PR is required by https://github.com/oneapi-src/unified-memory-framework/pull/1143

### DIFF
--- a/src/coarse/coarse.c
+++ b/src/coarse/coarse.c
@@ -1170,10 +1170,13 @@ umf_result_t coarse_free(coarse_t *coarse, void *ptr, size_t bytes) {
     }
 
     block_t *block = get_node_block(node);
-    assert(block->used);
+    if (!block->used) {
+        LOG_ERR("double free");
+        utils_mutex_unlock(&coarse->lock);
+        return UMF_RESULT_ERROR_INVALID_ARGUMENT;
+    }
 
     if (bytes > 0 && bytes != block->size) {
-        // wrong size of allocation
         LOG_ERR("wrong size of allocation");
         utils_mutex_unlock(&coarse->lock);
         return UMF_RESULT_ERROR_INVALID_ARGUMENT;
--- a/test/coarse_lib.cpp
+++ b/test/coarse_lib.cpp
@@ -160,6 +160,13 @@ TEST_P(CoarseWithMemoryStrategyTest, coarseTest_basic_provider) {
     ASSERT_EQ(coarse_get_stats(ch).alloc_size, alloc_size);
     ASSERT_EQ(coarse_get_stats(ch).num_all_blocks, 1);
 
+    // test double free
+    umf_result = coarse_free(ch, ptr, 2 * MB);
+    ASSERT_EQ(umf_result, UMF_RESULT_ERROR_INVALID_ARGUMENT);
+    ASSERT_EQ(coarse_get_stats(ch).used_size, 0);
+    ASSERT_EQ(coarse_get_stats(ch).alloc_size, alloc_size);
+    ASSERT_EQ(coarse_get_stats(ch).num_all_blocks, 1);
+
     coarse_delete(ch);
     umfMemoryProviderDestroy(malloc_memory_provider);
 }
@@ -198,6 +205,13 @@ TEST_P(CoarseWithMemoryStrategyTest, coarseTest_basic_fixed_memory) {
 
     umf_result = coarse_free(ch, ptr, 2 * MB);
     ASSERT_EQ(umf_result, UMF_RESULT_SUCCESS);
+    ASSERT_EQ(coarse_get_stats(ch).used_size, 0);
+    ASSERT_EQ(coarse_get_stats(ch).alloc_size, buff_size);
+    ASSERT_EQ(coarse_get_stats(ch).num_all_blocks, 1);
+
+    // test double free
+    umf_result = coarse_free(ch, ptr, 2 * MB);
+    ASSERT_EQ(umf_result, UMF_RESULT_ERROR_INVALID_ARGUMENT);
     ASSERT_EQ(coarse_get_stats(ch).used_size, 0);
     ASSERT_EQ(coarse_get_stats(ch).alloc_size, buff_size);
     ASSERT_EQ(coarse_get_stats(ch).num_all_blocks, 1);


vinser52:
@bratpiorka we have the following issue in the backlog: #886.
Might it be related to the issue you are trying to fix?

src/utils/utils_posix_concurrency.c
```diff
@@ -70,3 +71,56 @@ int utils_read_unlock(utils_rwlock_t *rwlock) {
 int utils_write_unlock(utils_rwlock_t *rwlock) {
     return pthread_rwlock_unlock((pthread_rwlock_t *)rwlock);
 }
+
+void utils_atomic_load_acquire_u64(uint64_t *ptr, uint64_t *out) {
+    utils_annotate_acquire(ptr);
```
vinser52, line 76: Could you fix the annotations in this PR, as I did in #1138 ?

bratpiorka: ok I will check it

src/utils/utils_sanitizers.h
```diff
@@ -168,6 +168,24 @@ static inline void utils_annotate_memory_inaccessible(void *ptr, size_t size) {
 #endif
 }
 
+static inline void utils_annotate_memory_new(void *ptr, size_t size) {
+#ifdef UMF_VG_MEMCHECK_ENABLED
```
vinser52, line 172: Should it be `UMF_VG_DRD_ENABLED`?
The `ANNOTATE_NEW_MEMORY` is defined in `drd.h` header.

Were you able to compile this code?

vinser52 (changes requested):

igchor:
Why did you add ALIGNED_8 in a few places? All integers we use should be naturally-aligned already

src/pool/pool_disjoint.c
```diff
@@ -420,36 +431,26 @@ static void bucket_update_stats(bucket_t *bucket, int in_use, int in_pool) {
...
-}
-
 static bool bucket_can_pool(bucket_t *bucket) {
     size_t new_free_slabs_in_bucket;
 
     new_free_slabs_in_bucket = bucket->chunked_slabs_in_pool + 1;
 
     // we keep at most params.capacity slabs in the pool
     if (bucket_max_pooled_slabs(bucket) >= new_free_slabs_in_bucket) {
-        size_t pool_size = 0;
```
igchor, line 436: what was the deadlock exactly? looking at this code, it's not clear

bratpiorka:
> Why did you add ALIGNED_8 in a few places? All integers we use should be naturally-aligned already

to 8 bytes? this is a requirement for Windows Interlocked*, see https://learn.microsoft.com/en-us/windows/win32/api/winnt/nf-winnt-interlockedcompareexchange64 requirements for args

src/pool/pool_disjoint.c
bratpiorka: I'm not exactly sure how this happened, but the deadlock was caused by two issues: a) decreasing shared_limits->total_size when it was equal to 0, causing it to become SIZE_T_MAX, and b) the utils_compare_exchange() function, whose implementation on Windows does not return a boolean value. Eventually, I decided to remove the while(true) loop entirely.

igchor:
> > Why did you add ALIGNED_8 in a few places? All integers we use should be naturally-aligned already
> 
> to 8 bytes? this is a requirement for Windows Interlocked*, see https://learn.microsoft.com/en-us/windows/win32/api/winnt/nf-winnt-interlockedcompareexchange64 requirements for args

All integers are automatically naturally aligned (that is, 32 bit interger is 4B aligned and 64bit integer is 8B aligned). Getting an unaligned integer is only possible if you have an allocator that doesn't align allocations to 8B (which is not the case for ba_alloc).
Unaligned integer access is always UB in C++ (even normal non-atomic accesses); I'm not sure about C.

src/pool/pool_disjoint.c
igchor: Hm, but isn't decreasing total_size below zero still possible? I don't see any logic here that would cause underflow, so it's probably somewhere else?

test/supp/drd-umf_test-ipc.supp
```diff
@@ -5,3 +5,13 @@
    fun:pthread_cond_destroy@*
    ...
 }
+
+{
+   [false-positive] Double check locking pattern in trackingOpenIpcHandle
+   drd:ConflictingAccess
+   ...
```
vinser52, line 12: why did you replace `fun:utils_atomic_load_acquire_ptr` with `...`?

You need to add the same suppression to the `helgrind-umf_test-ipc.supp` to solve Valgrind failure in CI.

src/pool/pool_disjoint.c
```diff
@@ -381,10 +382,20 @@ static slab_list_item_t *bucket_get_avail_slab(bucket_t *bucket,
...
+            uint64_t total_size_check;
+            utils_atomic_load_acquire_u64(&bucket->shared_limits->total_size,
+                                          &total_size_check);
+            assert(total_size_check >= bucket_slab_alloc_size(bucket));
+#endif
             // If this was an empty slab, it was in the pool.
             // Now it is no longer in the pool, so update count.
             --bucket->chunked_slabs_in_pool;
-            bucket_decrement_pool(bucket);
+            ALIGNED_8 uint64_t size_to_add = bucket_slab_alloc_size(bucket);
```
vinser52, line 395: I think you put `ALIGNED_8` in many places where it is not needed (like here).

I think you can add debug assert to the `utils_*` functions for the atomic operations in the `utils_concurrency_posix.c` to make sure the variables are aligned. But I believe in almost all cases in this PR the `ALIGNED_8` is not needed.

src/pool/pool_disjoint.c
bratpiorka: the code that decreases total_size is here: https://github.com/oneapi-src/unified-memory-framework/pull/1127/files#diff-7d792b84303f5dbbfba19a0b300b3645a0455df415f4246568dc0b1a5b846c73R396 

However, I do not understand how this could happen. This issue was very sporadic and occurred only in the multithreaded Disjoint Pool benchmark on Windows.

src/utils/utils_sanitizers.h
bratpiorka: fixed

test/supp/drd-umf_test-ipc.supp
bratpiorka: done

src/pool/pool_disjoint.c
bratpiorka: ok I removed all of them

src/pool/pool_disjoint.c
igchor: Hm, okay. I don't think your change should break anything, and if it avoids the deadlock, I'm fine with it. However, perhaps you could add some extra asserts? For example on line 452, where you do utild_fetch_and_sub, you could verify that the original value is big enough (and perhaps there are other places as well).

bratpiorka:
@ldorau @igchor @vinser52 please re-review

bratpiorka:
> clang ThreadSanitizer still detects data race between:
> 
> 1. Atomic write of size 8 at 0x7fb2f98bd6d8 by thread T15 (mutexes: write M0):
>    `critnib_remove()` at /home/runner/work/unified-memory-framework/unified-memory-framework/src/critnib/critnib.c:450
>    ` utils_atomic_store_release_ptr((void **)&n->child[slice_index(key, n->shift)], NULL);`
> 
> and
> 
> 2. Previous read of size 8 at 0x7fb2f98bd6d8 by thread T16:
>    `find_predecessor` at /home/runner/work/unified-memory-framework/unified-memory-framework/src/critnib/critnib.c:526
>    `if (n->child[nib]) {`
> 
> https://github.com/ldorau/unified-memory-framework/actions/runs/13562899026/job/37909488408

this should be fixed in the newest version of this PR

src/pool/pool_disjoint.c
bratpiorka: I already added such check here: https://github.com/oneapi-src/unified-memory-framework/pull/1127/files#diff-7d792b84303f5dbbfba19a0b300b3645a0455df415f4246568dc0b1a5b846c73R386

on line 452, I sub the value that I added just a few lines earlier. Since this is atomic, there is no easy way to assert this.

ldorau (approved):

benchmark/benchmark.cpp
```diff
@@ -96,9 +96,8 @@ UMF_BENCHMARK_TEMPLATE_DEFINE(multiple_malloc_free_benchmark,
                               pool_allocator<disjoint_pool<os_provider>>);
 UMF_BENCHMARK_REGISTER_F(multiple_malloc_free_benchmark, disjoint_pool_uniform)
     ->Apply(&default_multiple_alloc_uniform_size)
-    ->Apply(&singlethreaded);
-// TODO: change to multithreaded
-//->Apply(&multithreaded);
+    ->Apply(&singlethreaded)
```
lplewa, line 99: apply only multithreded.

src/pool/pool_disjoint.c
```diff
@@ -420,36 +431,26 @@ static void bucket_update_stats(bucket_t *bucket, int in_use, int in_pool) {
...
-                bucket_update_stats(bucket, -1, 1);
-                return true;
-            }
+
+        uint64_t size_to_add = bucket_slab_alloc_size(bucket);
+        size_t previous_size = utils_fetch_and_add_u64(
+            &bucket->shared_limits->total_size, size_to_add);
+
+        if (previous_size + size_to_add <= bucket->shared_limits->max_size) {
+
```
lplewa, line 447: new line

src/utils/utils_concurrency.h
```diff
@@ -27,10 +29,17 @@
 #else /* __cplusplus */
 #include <atomic>
 #define _Atomic(X) std::atomic<X>
+
+using std::memory_order_acq_rel;
+using std::memory_order_acquire;
+using std::memory_order_relaxed;
+using std::memory_order_release;
+
```
lplewa, line 37: Why this is needed. This looks like this should not bellong to this file

src/utils/utils_concurrency.h
```diff
@@ -27,10 +29,17 @@
 #else /* __cplusplus */
 #include <atomic>
 #define _Atomic(X) std::atomic<X>
+
+using std::memory_order_acq_rel;
```
vinser52, line 33: Why do we need C++ code here?

ldorau:
This PR is required by https://github.com/oneapi-src/unified-memory-framework/pull/1143

vinser52:
This PR is required by #998 as well.

src/pool/pool_disjoint.c
```diff
@@ -523,8 +524,8 @@ static void disjoint_pool_print_stats(disjoint_pool_t *pool) {
         utils_mutex_unlock(&bucket->bucket_lock);
     }
 
-    LOG_DEBUG("current pool size: %zu",
-              disjoint_pool_get_limits(pool)->total_size);
+    LOG_DEBUG("current pool size: %llu",
+              (unsigned long long)disjoint_pool_get_limits(pool)->total_size);
```
lplewa, line 528: why?

src/critnib/critnib.c
```diff
@@ -128,29 +129,9 @@ struct critnib {
     struct critnib_node *pending_del_nodes[DELETED_LIFE];
     struct critnib_leaf *pending_del_leaves[DELETED_LIFE];
```
lplewa, line 131: why?

src/critnib/critnib.c
```diff
@@ -653,7 +638,9 @@ static struct critnib_leaf *find_successor(struct critnib_node *__restrict n) {
     while (1) {
         unsigned nib;
         for (nib = 0; nib <= NIB; nib++) {
-            if (n->child[nib]) {
+            struct critnib_node *m;
+            utils_atomic_load_acquire_ptr((void **)&n->child[nib], (void **)&m);
+            if (m) {
                 break;
             }
         }
```
lplewa, line 646: What is a rationale for this change. Was old code wrong / not safe?

src/utils/utils_concurrency.h
```diff
@@ -92,57 +101,162 @@ static __inline unsigned char utils_mssb_index(long long value) {
...
 
-// NOTE: windows version have different order of args
-#define utils_compare_exchange(object, desired, expected)                      \
-    InterlockedCompareExchange64((LONG64 volatile *)object, *expected, *desired)
+static __inline bool utils_compare_exchange_u64(uint64_t *ptr,
+                                                uint64_t *expected,
+                                                uint64_t *desired) {
+    ASSERT_IS_ALIGNED((uintptr_t)ptr, 8);
+    ASSERT_IS_ALIGNED((uintptr_t)desired, 8);
+    ASSERT_IS_ALIGNED((uintptr_t)expected, 8);
```
lplewa, line 167: Are you sure that this ptrs has to by 8b aligned?

src/utils/utils_concurrency.h
```diff
@@ -92,57 +101,162 @@ static __inline unsigned char utils_mssb_index(long long value) {
...
+static __inline uint64_t utils_fetch_and_add_u64(uint64_t *ptr, uint64_t val) {
+    ASSERT_IS_ALIGNED((uintptr_t)ptr, 8);
+    ASSERT_IS_ALIGNED((uintptr_t)&val, 8);
+    // return the value that had previously been in *ptr
+    return InterlockedExchangeAdd64((LONG64 volatile *)(ptr), val);
+}
+
+static __inline uint64_t utils_fetch_and_sub_u64(uint64_t *ptr, uint64_t val) {
+    ASSERT_IS_ALIGNED((uintptr_t)ptr, 8);
+    ASSERT_IS_ALIGNED((uintptr_t)&val, 8);
```
lplewa, line 156: why?

src/utils/utils_concurrency.h
```diff
@@ -92,57 +101,162 @@ static __inline unsigned char utils_mssb_index(long long value) {
...
+
+static __inline uint64_t utils_atomic_decrement_u64(uint64_t *ptr) {
+    ASSERT_IS_ALIGNED((uintptr_t)ptr, 8);
+    // return decremented value
+    return InterlockedDecrement64((LONG64 volatile *)ptr);
+}
+
+static __inline uint64_t utils_fetch_and_add_u64(uint64_t *ptr, uint64_t val) {
+    ASSERT_IS_ALIGNED((uintptr_t)ptr, 8);
+    ASSERT_IS_ALIGNED((uintptr_t)&val, 8);
```
lplewa, line 149: why?

lplewa (changes requested):

src/critnib/critnib.c
ldorau: This is it: https://github.com/oneapi-src/unified-memory-framework/actions/runs/13587677512/job/37987102380

vinser52: @ldorau sanitizer might report false positives. Are we sure it is a real issue?

src/utils/utils_concurrency.h
bratpiorka: https://learn.microsoft.com/en-us/windows/win32/api/winnt/nf-winnt-interlockedcompareexchange64

The variables for this function must be aligned on a 64-bit boundary; otherwise, this function will behave unpredictably on multiprocessor x86 systems and any non-x86 systems

src/utils/utils_concurrency.h
bratpiorka: this code was indirectly included in c++ tests..
moved to test/base.hpp

benchmark/benchmark.cpp
bratpiorka: done

src/pool/pool_disjoint.c
bratpiorka: done

src/utils/utils_concurrency.h
bratpiorka: moved to test/base.hpp

src/pool/pool_disjoint.c
bratpiorka: total_size is now uint 64bit as other atomics. This is needed for MacOS. Please suggest better change

src/critnib/critnib.c
bratpiorka: reverted

src/utils/utils_concurrency.h
bratpiorka: removed

src/utils/utils_concurrency.h
```diff
@@ -92,57 +101,162 @@ static __inline unsigned char utils_mssb_index(long long value) {
...
+static inline void utils_atomic_load_acquire_ptr(void **ptr, void **out) {
+    ASSERT_IS_ALIGNED((uintptr_t)ptr, 8);
+    ASSERT_IS_ALIGNED((uintptr_t)out, 8);
+    *out = (void *)__atomic_load_n((uintptr_t *)ptr, memory_order_acquire);
+    utils_annotate_acquire((void *)ptr);
+}
 
-#define utils_atomic_increment(object)                                         \
-    __atomic_add_fetch(object, 1, memory_order_acq_rel)
+static inline void utils_atomic_store_release_u64(uint64_t *ptr,
```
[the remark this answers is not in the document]
bratpiorka: removed

src/utils/utils_concurrency.h
lplewa: We are dereferencing this pointers, so it do not matter what alignment, it has here, as anyway it their value will be copied to the register/stack before call.

src/critnib/critnib.c
ldorau: It was needed only by #1143. After adding a lock to #1143 it is not required any more.

src/utils/utils_concurrency.h
```diff
@@ -92,57 +88,144 @@ static __inline unsigned char utils_mssb_index(long long value) {
...
-    do {                                                                       \
-        *(LONG64 *)dest =                                                      \
-            InterlockedOr64Acquire((LONG64 volatile *)object, 0);              \
-    } while (0)
+static __inline void utils_atomic_load_acquire_u64(uint64_t *ptr,
+                                                   uint64_t *out) {
+    // NOTE: Windows cl complains about direct accessing 'ptr' which is next
+    // accessed using Interlocked* functions (warning 28112 - disabled)
+    ASSERT_IS_ALIGNED((uintptr_t)ptr, 8);
+    LONG64 ret = InterlockedCompareExchange64((LONG64 volatile *)ptr, 0, 0);
```
lplewa, line 96: This need comment how and why it works

src/utils/utils_concurrency.h
```diff
@@ -92,57 +88,144 @@ static __inline unsigned char utils_mssb_index(long long value) {
...
+    utils_annotate_acquire(ptr);
```
lplewa, line 97: why we need acquire after compareandswap, with full memory fence

src/utils/utils_concurrency.h
```diff
@@ -92,57 +88,144 @@ static __inline unsigned char utils_mssb_index(long long value) {
...
+    *out = *(uint64_t *)&ret;
+}
 
-#define utils_atomic_store_release(object, desired)                            \
-    InterlockedExchange64((LONG64 volatile *)object, (LONG64)desired)
+static __inline void utils_atomic_load_acquire_ptr(void **ptr, void **out) {
+    ASSERT_IS_ALIGNED((uintptr_t)ptr, 8);
+    uintptr_t ret = (uintptr_t)InterlockedCompareExchangePointer(ptr, 0, 0);
+    utils_annotate_acquire((void *)ptr);
```
lplewa, line 104: dito

lplewa (changes requested):

src/pool/pool_disjoint.c
igchor: Well, you could at least add a check like this:

```
size_t old = utils_fetch_and_sub_u64(&bucket->shared_limits->total_size,
                                    size_to_add);
assert(old >= size_to_add);
```

src/critnib/critnib.c
vinser52: Which lock? What about scalability?

src/pool/pool_disjoint.c
bratpiorka: Great idea, but I don't want to mix PRI macros with what we currently have in our code. If we want to use PRI, it should be implemented in a whole project with a separate PR.

src/utils/utils_concurrency.h
bratpiorka: Are you 100% sure that the compiler will copy these arguments and not use them directly? By the way, I really don't understand why we're discussing asserts here. In my opinion, there can never be too many asserts in the code.

src/utils/utils_concurrency.h
```diff
@@ -92,57 +88,144 @@ static __inline unsigned char utils_mssb_index(long long value) {
...
+    ASSERT_IS_ALIGNED((uintptr_t)ptr, 8);
+    ASSERT_IS_ALIGNED((uintptr_t)out, 8);
+    __atomic_load(ptr, out, memory_order_acquire);
+    utils_annotate_acquire(ptr);
+}
+
+static inline void utils_atomic_load_acquire_ptr(void **ptr, void **out) {
+    ASSERT_IS_ALIGNED((uintptr_t)ptr, 8);
+    ASSERT_IS_ALIGNED((uintptr_t)out, 8);
+    *out = (void *)__atomic_load_n((uintptr_t *)ptr, memory_order_acquire);
```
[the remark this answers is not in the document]
bratpiorka: changed to atomic_load()

src/utils/utils_concurrency.h
bratpiorka: removed

src/utils/utils_concurrency.h
bratpiorka: done

src/utils/utils_concurrency.h
```diff
@@ -92,57 +101,148 @@ static __inline unsigned char utils_mssb_index(long long value) {
...
+    ASSERT_IS_ALIGNED((uintptr_t)ptr, 8);
+    ASSERT_IS_ALIGNED((uintptr_t)desired, 8);
+    ASSERT_IS_ALIGNED((uintptr_t)expected, 8);
+
+    // if (*ptr == *desired)
+    //   *ptr = *expected
+    //   return true
+    // else
+    //  *expected = *ptr
+    // return false
```
vinser52, line 168: I think we do not need this comment. In case of questions it is better to read documentation for the `InterlockedCompareExchange64` API.

src/utils/utils_concurrency.h
```diff
@@ -92,57 +101,148 @@ static __inline unsigned char utils_mssb_index(long long value) {
...
-#define utils_atomic_load_acquire(object, dest)                                \
-    do {                                                                       \
-        utils_annotate_acquire((void *)object);                                \
-        __atomic_load(object, dest, memory_order_acquire);                     \
-    } while (0)
+static inline void utils_atomic_load_acquire_u64(uint64_t *ptr, uint64_t *out) {
+    ASSERT_IS_ALIGNED((uintptr_t)ptr, 8);
+    ASSERT_IS_ALIGNED((uintptr_t)out, 8);
+    __atomic_load(ptr, out, memory_order_acquire);
+    //utils_annotate_acquire(ptr);
```
vinser52, line 190: Why is it commented?

src/utils/utils_concurrency.h
```diff
@@ -92,57 +101,148 @@ static __inline unsigned char utils_mssb_index(long long value) {
...
+    ASSERT_IS_ALIGNED((uintptr_t)ptr, 8);
+    ASSERT_IS_ALIGNED((uintptr_t)expected, 8);
+    ASSERT_IS_ALIGNED((uintptr_t)desired, 8);
+
+    // if (*ptr == *expected)
+    //   *ptr = *desired
+    //   return true
+    // else
+    //  *expected = *ptr
+    // return false
```
vinser52, line 241: I do not think this comment is needed, because everyone knows or can google how the `__atomic_compare_exchange` works.

src/pool/pool_disjoint.c
bratpiorka: ok, done

src/utils/utils_concurrency.h
lplewa: Guys, this function, takes second and third argument as a *Value*. So this do not matter PTR is aligned or not it do not matter at all, as function will get a *copy* values instead of the address. 

``` c
LONG64 InterlockedCompareExchange64(
  [in, out] LONG64 volatile *Destination,
  [in]      LONG64          ExChange,
  [in]      LONG64          Comperand
);```

vinser52: ahh, sorry, I did not saw that your comment was not about destination. Of coarse, only destination should be aligned.

src/pool/pool_disjoint.c
bratpiorka: added

src/utils/utils_concurrency.h
bratpiorka: removed

src/utils/utils_concurrency.h
bratpiorka: removed

src/utils/utils_concurrency.h
bratpiorka: removed

igchor (approved):

src/critnib/critnib.c
ldorau: It is still needed now: https://github.com/oneapi-src/unified-memory-framework/actions/runs/13650462662/job/38158581307

bratpiorka:
@lplewa please re-review

src/utils/utils_concurrency.h
vinser52: @bratpiorka but C++ code is still there and in the `test/base.hpp` only Copyright is changed

bratpiorka: Yes, I tried moving it to base.hpp, but that broke the building of examples. I also attempted to remove the __cplusplus code entirely, but it's not that simple. So, in fact, I'm not adding new C++ code here; I'm extending the existing code. Further refactoring should be done in a separate PR.

src/critnib/critnib.c
lplewa: Can anyone explain rationale why this change is needed? I need explanation why from logical point of view, of this code. That it suppress/fixes an sanitizer bug do not explain anything. From my perspective this change is like doing random changes until sanitizer is fine.

src/utils/utils_concurrency.h
lplewa: > Yes, I tried moving it to base.hpp, but that broke the building of examples. 
Why? Our examples should not use any code either from umf internals, nor tests

>I also attempted to remove the __cplusplus code entirely, but it's not that simple. So, in fact, I'm not adding new C++ code here; I'm extending the existing code. Further refactoring should be done in a separate PR.
You are adding extra using, and stil we do not know why they are needed.

src/critnib/critnib.c
bratpiorka: removed
@ldorau please add this change to @1143 as it is needed (or not?) there

bratpiorka:
continue in #1151

src/critnib/critnib.c
ldorau: > removed @ldorau please add this change to #1143 as it is needed (or not?) there

OK

PR #1134, opened 2025-02-25
Stack the same values in the tracker

ldorau:
### Description

Stack the same values in the tracker.

It is an alternative solution to the #1121 (without splitting and merging tracker entries)

Ref: #1121

### Checklist

- [x] All tests pass locally


ldorau:
@vinser52 @igchor @bratpiorka @lplewa @kswiecicki Please review also this alternative solution without splitting and merging tracker entries.

test/provider_fixed_memory.cpp
vinser52, line 1: Could you please enable the `ipcFixture.hpp` tests for the Fixed provider on top of L0 pool.

ldorau: I cannot, because the Fixed provider does not support IPC at all.

ldorau:
It will be implemented in another way soon.

PR #1136, opened 2025-02-25
Add tests for pool from pointer to poolFixtures.hpp

ldorau:
### Description

Add tests for pool from pointer to poolFixtures.hpp.

Depends on: https://github.com/oneapi-src/unified-memory-framework/pull/1134

### Checklist

- [x] All tests pass locally

### DIFF
--- a/src/provider/provider_tracking.c
+++ b/src/provider/provider_tracking.c
@@ -37,68 +37,143 @@ struct umf_memory_tracker_t {
 typedef struct tracker_alloc_info_t {
     umf_memory_pool_handle_t pool;
     size_t size;
+    // list of previous entries with the same address (LIFO)
+    struct tracker_alloc_info_t *prev;
 } tracker_alloc_info_t;
+
+static umf_result_t
+umfMemoryTrackerAddValue(umf_memory_tracker_handle_t hTracker, const void *ptr,
+                         tracker_alloc_info_t *new_value) {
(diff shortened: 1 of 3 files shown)


vinser52:
Should we create a separate PR with the test (e.g. the last commit in this PR) and merge it. As I understand the test itself is generic and can be used with different pools/providers.

src/provider/provider_tracking.c
```diff
@@ -37,68 +37,143 @@ struct umf_memory_tracker_t {
...
-                size);
-        return UMF_RESULT_ERROR_OUT_OF_HOST_MEMORY;
-    }
-
-    value->pool = pool;
-    value->size = size;
-
-    int ret =
-        critnib_insert(hTracker->alloc_segments_map, (uintptr_t)ptr, value, 0);
+    umf_result_t umf_result = UMF_RESULT_ERROR_UNKNOWN;
```
KFilipek, line 50: Move this declaration to be sticked to this if:
```cpp
    if (ret != EEXIST) {
```
Similarly to the `int ret = ...`

ldorau:
> Should we create a separate PR with the test (e.g. the last commit in this PR) and merge it. As I understand the test itself is generic and can be used with different pools/providers.

This PR is incorporated into https://github.com/oneapi-src/unified-memory-framework/pull/1143

PR #1138, opened 2025-02-25
Fix Valgrind and TSAN instrumentation

vinser52:
### Description
This PR fixes Valgrind and TSAN instrumentation.

Ref. #886 

### Checklist

- [x] All tests pass locally

### DIFF
--- a/src/utils/utils_concurrency.h
+++ b/src/utils/utils_concurrency.h
@@ -121,14 +121,14 @@ static __inline unsigned char utils_mssb_index(long long value) {
 
 #define utils_atomic_load_acquire(object, dest)                                \
     do {                                                                       \
-        utils_annotate_acquire((void *)object);                                \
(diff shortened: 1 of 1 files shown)


bratpiorka:
I see that TSAN passed .. maybe there annotation should be placed differently for TSAN and Valgrind?
Also, should we remove some suppressions with this PR?

lplewa:
I'm 99% sure that issue is that this two macros are "empty" https://github.com/oneapi-src/unified-memory-framework/blob/113f3e097afee8a956c20ddb0313bbff48cb93bd/src/utils/utils_common.h#L56-L57

While we should call ANNOTATE_NEW_MEMORY and VALGRIND_HG_DISABLE_CHECKING and ANNOTATE_NEW_MEMORY

bratpiorka:
@vinser52 I added this code to #1127 so you can close this PR

PR #1139, opened 2025-02-26
Use `LOG_FATAL()` in case of critical errors

ldorau:
### Description

Use `LOG_FATAL()` in case of critical errors.

Ref: #1095

### Checklist

- [x] All tests pass locally


KFilipek (approved):
I've restarted workflows due to a connection issue. LGTM.

ldorau:
> I think that there should be `no value left` instead of `no left value`, but never mind.

No, it should be `no the left one value` - this is the **right** meaning :-)))

lukaszstolarczuk (approved):

PR #1142, opened 2025-02-27
Remove old SHM files left from the previous runs because of crashes

ldorau:
### Description

Remove old SHM files `/tmp/umf_file_provider_*`
(at the beginning of the test)
left from the previous runs, because of crashes
of the `ipc_file_prov` test.

### Checklist

- [x] All tests pass locally

### DIFF
--- a/test/ipc_file_prov.sh
+++ b/test/ipc_file_prov.sh
@@ -9,7 +9,12 @@
 
 set -e
 
-FILE_NAME="/tmp/umf_file_provider_$$"
+FILE_BASE="/tmp/umf_file_provider"
+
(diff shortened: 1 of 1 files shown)


vinser52 (approved):

lukaszstolarczuk (approved):

lukaszstolarczuk:
> We cannot use `tmpfile()` nor `mkstemp()`, because those functions create and open a temporary file on their own, but we pass only a name of the file to the file provider and it opens this file on its own - we cannot force file memory provider to use only temporary files.

`mktemp` is would you're looking for, but the current solution is as good, I believe.
`/tmp` files are deleted on reboot... but we almost never do it 😉

PR #1146, opened 2025-02-28
Make umf-* prefix for tests instead of test_*

KFilipek:
### Checklist

- [ ] All tests pass locally


bratpiorka (approved):

lplewa:
If we are renaming test names, can we rename them in that way that we use only '-' or '_', not both. I find it really annoying when typing name of the test.

KFilipek:
> If we are renaming test names, can we rename them in that way that we use only '-' or '_', not both. I find it really annoying when typing name of the test.

According to GTest documentation each test should be named using camel-case - it should solve each problem.

bratpiorka (approved):

lukaszstolarczuk:
I vote for combining both - Patryk's and Lukasz's proposals: `test_<test_bin_name>` (fixing test_bin_name's is totally different story)

lukaszstolarczuk (approved):

PR #1143, opened 2025-02-27
Use eight level of critnibs in the tracker

ldorau:
### Description

Use eight level of critnibs in the tracker.
Multilevel maps are needed to support the case
when one memory pool acts as a memory provider
for another memory pool (nested memory pooling).

Depends on:
- [x] https://github.com/oneapi-src/unified-memory-framework/pull/1151

Requires:
- [x] https://github.com/oneapi-src/unified-memory-framework/pull/1176

### Checklist

- [x] Code compiles without errors locally
- [x] All tests pass locally
- [x] CI workflows execute properly

PR #1157, opened 2025-03-05
Fix disjoint_pool_free() and proxy_free()

ldorau:
### Description

Fix disjoint_pool_free() and proxy_free() - error out on pool mismatch, when we try to free the pointer
belonging to a different pool.

### Checklist

- [x] All tests pass locally

### DIFF
--- a/src/pool/pool_disjoint.c
+++ b/src/pool/pool_disjoint.c
@@ -17,6 +17,7 @@
 #include <umf/memory_provider.h>
 
 #include "base_alloc_global.h"
+#include "memory_pool_internal.h"
 #include "pool_disjoint_internal.h"
 #include "provider/provider_tracking.h"
 #include "uthash/utlist.h"
@@ -817,6 +818,11 @@ umf_result_t disjoint_pool_free(void *pool, void *ptr) {
             TLS_last_allocation_error = ret;
             LOG_ERR("failed to get allocation info from the memory tracker");
             return ret;
+        }
+
+        if (pool != umfPoolGetPoolPriv(allocInfo.pool)) {
+            LOG_ERR("pool mismatch");
+            return UMF_RESULT_ERROR_INVALID_ARGUMENT;
         }
 
         size_t size = allocInfo.baseSize;
--- a/src/memory_pool.c
+++ b/src/memory_pool.c
@@ -208,3 +208,10 @@ umf_result_t umfPoolGetTag(umf_memory_pool_handle_t hPool, void **tag) {
     utils_mutex_unlock(&hPool->lock);
     return UMF_RESULT_SUCCESS;
 }
+
+void *umfPoolGetPoolPriv(umf_memory_pool_handle_t hPool) {
+    if (hPool == NULL) {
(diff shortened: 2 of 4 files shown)


src/pool/pool_disjoint.c
```diff
@@ -819,6 +820,11 @@ umf_result_t disjoint_pool_free(void *pool, void *ptr) {
             return ret;
         }
 
+        if (pool != umfPoolGetPoolPriv(allocInfo.pool)) {
+            LOG_ERR("pool mismatch");
+            return UMF_RESULT_ERROR_INVALID_ARGUMENT;
```
bratpiorka, line 825: There should be added a negative test for this, even simplest

src/pool/pool_disjoint.c
```diff
@@ -819,6 +820,11 @@ umf_result_t disjoint_pool_free(void *pool, void *ptr) {
             return ret;
         }
 
+        if (pool != umfPoolGetPoolPriv(allocInfo.pool)) {
```
vinser52, line 823: Why do we need this check? It breaks encapsulation because now pool implementation calls `umfPoolGetPoolPriv`.

ldorau: This check prevents from freeing a pointer from a different pool.

ldorau:
> The name of this PR is misleading because it does not fix anything, it just introduces an additional check. But do we really need this check?

`proxy_pool_free()` can free a pointer from a different pool than it was called with now. I think it is a bug, so this is a fix for this bug.

src/pool/pool_disjoint.c
ldorau: @vinser52 It is tested in https://github.com/oneapi-src/unified-memory-framework/pull/1143
See https://github.com/oneapi-src/unified-memory-framework/pull/1143#discussion_r1981344299

ldorau: @vinser52 It is needed by https://github.com/oneapi-src/unified-memory-framework/pull/1143
See https://github.com/oneapi-src/unified-memory-framework/pull/1143#discussion_r1981344299

vinser52:
> > The name of this PR is misleading because it does not fix anything, it just introduces an additional check. But do we really need this check?
> 
> `proxy_pool_free()` can free a pointer from a different pool than it was called with now. I think it is a bug, so this is a fix for this bug.

it is the ill-formed client.

src/pool/pool_disjoint.c
ldorau: This PR was replaced with https://github.com/oneapi-src/unified-memory-framework/pull/1161 - please review this new PR instead.

ldorau:
> I'm not sure if adding this check in the pool is a good idea. As I understand, all other pools will NOT return an error if we try to free memory from a different pool (which is inconsistent). And even if we added this check to proxy_pool, jemalloc_pool, what about custom user pools?

This PR was replaced with https://github.com/oneapi-src/unified-memory-framework/pull/1161 - please review this new PR instead.

PR #1161, opened 2025-03-06
Do not free a pointer from another pool

ldorau:
### Description

Do not free a pointer from another pool.
Error out when `free()` tries to free a pointer used by another pool
or a pointer belonging to another pool.
Add the `UMF_RESULT_ERROR_INVALID_FREE_OP` error for this case.

### Checklist

- [x] All tests pass locally

### DIFF
--- a/src/memory_pool.c
+++ b/src/memory_pool.c
@@ -180,6 +180,14 @@ size_t umfPoolMallocUsableSize(umf_memory_pool_handle_t hPool, void *ptr) {
 
 umf_result_t umfPoolFree(umf_memory_pool_handle_t hPool, void *ptr) {
     UMF_CHECK((hPool != NULL), UMF_RESULT_ERROR_INVALID_ARGUMENT);
+    umf_memory_pool_handle_t poolTracker = umfMemoryTrackerGetPool(ptr);
+    if (poolTracker != NULL && poolTracker != hPool) {
+        LOG_ERR("invalid free operation: trying to free a pointer %p used by "
+                "another pool or belonging to another pool %p than the given "
+                "pool %p",
+                ptr, poolTracker, hPool);
+        return UMF_RESULT_ERROR_INVALID_FREE_OP;
+    }
     return hPool->ops.free(hPool->pool_priv, ptr);
 }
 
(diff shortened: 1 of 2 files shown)


src/memory_pool.c
```diff
@@ -180,6 +180,14 @@ size_t umfPoolMallocUsableSize(umf_memory_pool_handle_t hPool, void *ptr) {
 
 umf_result_t umfPoolFree(umf_memory_pool_handle_t hPool, void *ptr) {
     UMF_CHECK((hPool != NULL), UMF_RESULT_ERROR_INVALID_ARGUMENT);
+    umf_memory_pool_handle_t poolTracker = umfMemoryTrackerGetPool(ptr);
```
bratpiorka, line 183: with this call, we would search the tracker 2 times during the free?

ldorau: AFAIK in case of Proxy Pool and Disjoint Pool - yes

ldorau:
This PR is tested in https://github.com/oneapi-src/unified-memory-framework/pull/1143
by the following tests:
`TEST_P(TrackingProviderTest, reverted_free_half_size)`
and
`TEST_P(TrackingProviderTest, reverted_free_the_same_size)`

src/memory_pool.c
bratpiorka: so this could significantly hurt the performance. Could you run the benchmarks and ask @lplewa for interpretation of results?

vinser52: Is this problem specific to the disjoint and proxy pools? Does the issue exist with the scalable or with the jemalloc pools?

ldorau: I think it is theoretically possible for every pool.

vinser52: As @bratpiorka mentioned it might impact performance.
so it is obvious that if an incorrect pool handle is passed by the client it is an ill-formed client's code. The question is do we want to catch such issues (and probably pay for the performance impact) or we should just say it is "undefined behavior" if an incorrect pool handle is passed as a parameter?

ldorau: I vote for UB in such case.

ldorau:
Passing an incorrect pool handle to `umfPoolFree()` is an Undefined Behavior.

PR #1176, opened 2025-03-10
add missing nullcheck in critnib

lplewa:
### Checklist

- [ ] All tests pass locally

### DIFF
--- a/src/critnib/critnib.c
+++ b/src/critnib/critnib.c
@@ -537,6 +537,11 @@ find_predecessor(struct critnib_node *__restrict n) {
         }
 
         utils_atomic_load_acquire_ptr((void **)&n->child[nib], (void **)&n);
+
+        if (!n) {
+            return NULL;
+        }
+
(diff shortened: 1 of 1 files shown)


ldorau (approved):

ldorau:
@vinser52 This is the fix we need - please review

vinser52 (approved):

src/critnib/critnib.c
```diff
@@ -650,6 +655,11 @@ static struct critnib_leaf *find_successor(struct critnib_node *__restrict n) {
         }
 
         n = n->child[nib];
```
lukaszstolarczuk, line 657: just asking - shouldn't we also have atomic_load_acquire here?

vinser52: @ldoray did it in the scope of #1143. See this [commit](https://github.com/oneapi-src/unified-memory-framework/pull/1143/commits/44e498193139bfed65b874619c24c06bc0196c52)

Issue #1171, opened 2025-03-07
CUDA Runtime API support


irozanova:
# CUDA Runtime API support

## Rationale

Some applications are written on pure CUDA Runtime API without using Driver API. But currently UMF requires to provide cuda context and device from Driver API to create UMF CUDA provider. Runtime API does not have any contexts and device datatype is different. 

## Description

As a possible solution device and context values can be initialized in umfCUDAMemoryProviderParamsCreate with the "current" values. It will allow to support CUDA Runtime and Driver API at the same time. In this case an application will need just to call umfCUDAMemoryProviderParamsCreate -> umfCUDAMemoryProviderOps -> umfMemoryProviderCreate. 

```
CUresult cuCtxGetCurrent ( CUcontext* pctx )
Returns the CUDA context bound to the calling CPU thread.

CUresult cuCtxGetDevice ( CUdevice* device )
Returns the device handle for the current context.
```

PR #1180, opened 2025-03-12
use current ctx and dev by default in CUDA prov

bratpiorka:
use the current ctx and dev by default in the CUDA provider 

implements https://github.com/oneapi-src/unified-memory-framework/issues/1171

### DIFF
--- a/include/umf/providers/provider_cuda.h
+++ b/include/umf/providers/provider_cuda.h
@@ -20,7 +20,8 @@ typedef struct umf_cuda_memory_provider_params_t
     *umf_cuda_memory_provider_params_handle_t;
 
 /// @brief  Create a struct to store parameters of the CUDA Memory Provider.
-/// @param  hParams [out] handle to the newly created parameters struct.
+/// @param  hParams [out] handle to the newly created parameters structure,
+///         initialized with the default (current) context and device ID.
 /// @return UMF_RESULT_SUCCESS on success or appropriate error code on failure.
 umf_result_t umfCUDAMemoryProviderParamsCreate(
     umf_cuda_memory_provider_params_handle_t *hParams);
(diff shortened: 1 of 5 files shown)


include/umf/providers/provider_cuda.h
```diff
@@ -34,13 +34,17 @@ umf_result_t umfCUDAMemoryProviderParamsDestroy(
 /// @brief  Set the CUDA context handle in the parameters struct.
 /// @param  hParams handle to the parameters of the CUDA Memory Provider.
 /// @param  hContext handle to the CUDA context.
+/// @note   This function is optional - if no context is set in the params
```
vinser52, line 37: `This function is optional` sounds confusing. I understand what you meant, but others might be confused.
Maybe we should add a note the `umfCUDAMemoryProviderParamsCreate` function and say that returned params structure is initialized with default value, and default context is a current context and the same for device.

bratpiorka: is it better now?

vinser52 (approved):

src/provider/provider_cuda.c
```diff
@@ -260,8 +263,29 @@ umf_result_t umfCUDAMemoryProviderParamsCreate(
...
+        params_data->cuda_context_handle = NULL;
+    }
+
+    CUdevice current_device = -1;
+    cu_result = g_cu_ops.cuCtxGetDevice(&current_device);
+    if (cu_result == CUDA_SUCCESS) {
+        params_data->cuda_device_handle = current_device;
+    } else {
+        params_data->cuda_device_handle = -1;
+    }
```
lplewa, line 287: nit: Whouln't it better to keep NULL/-1, and in provider_create function if params are NULL/-1 try to get default one?

not nit: What happents if getCurrent/getDevice fails - we silently ignore error here, so user gets error message during create about invalid argument? Not sure if it's userfrendly.

bratpiorka: I added LOG_ERR messages to cu_memory_provider_initialize().
Note that the user can set the arguments multiple times before they are passed to initialize(), where they will be finally checked.

lplewa (approved):

PR #1183, opened 2025-03-12
Fix segfault in `cu_memory_provider_get_last_native_error()`

ldorau:
### Description

Fix segfault in `cu_memory_provider_get_last_native_error()` when it is called after a CUDA device was destroyed.

### Checklist

- [x] All tests pass locally

### DIFF
--- a/src/provider/provider_cuda.c
+++ b/src/provider/provider_cuda.c
@@ -542,22 +542,41 @@ static void cu_memory_provider_get_last_native_error(void *provider,
         return;
     }
 
-    const char *error_name = 0;
-    const char *error_string = 0;
-    g_cu_ops.cuGetErrorName(TLS_last_native_error.native_error, &error_name);
-    g_cu_ops.cuGetErrorString(TLS_last_native_error.native_error,
-                              &error_string);
-
+    CUresult result;
     size_t buf_size = 0;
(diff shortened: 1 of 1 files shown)


ldorau:
@bratpiorka @lukaszstolarczuk it is required for the CUDA fix: https://github.com/intel/llvm/pull/17411

PatKamin (approved):

src/provider/provider_cuda.c
```diff
@@ -546,15 +546,27 @@ static void cu_memory_provider_get_last_native_error(void *provider,
                               &error_string);
 
     size_t buf_size = 0;
-    strncpy(TLS_last_native_error.msg_buff, error_name, TLS_MSG_BUF_LEN - 1);
+    if (error_name) {
+        strncpy(TLS_last_native_error.msg_buff, error_name,
+                TLS_MSG_BUF_LEN - 1);
+    } else {
+        strncpy(TLS_last_native_error.msg_buff, "cuGetErrorName() failed",
```
lplewa, line 553: nit: can we extend this error message, or/and add comment, that it can happen if cuda is allready destroyed. 

documentation for this functions, says that they never return NULL, so this might be misleading

ldorau: Done

bratpiorka (approved):

KFilipek (approved):

PR #1188, opened 2025-03-13
Move C++ helper header to utils

KFilipek:
# Nothing interesting
Just move C++ header to tests since it's never used by any code in `src/`. Minor deletes with trailing whitespaces and bumps date due to annoying checkers that forced me to do that.


PatKamin (approved):

bratpiorka:
+1 we should have no C++ code in src

bratpiorka (approved):

lukaszstolarczuk (commented):
I can see in cpp_helpers we use `umf` namespace, probably to be updated to `umf_test`. Also btw, could also update `#define UMF_HELPERS_HPP` -> `#define UMF_TEST_HELPERS_HPP`

PR #1197, opened 2025-03-17
Suppress false-positive in IPC tests under helgrind

vinser52:
### Description
Suppress false-positive in IPC tests under helgrind.

### Checklist

- [ ] All tests pass locally

### DIFF
--- a/test/supp/helgrind-test_ipc.supp
+++ b/test/supp/helgrind-test_ipc.supp
@@ -19,6 +19,7 @@
    [false-positive] Double check locking pattern in trackingOpenIpcHandle
    Helgrind:Race
    fun:utils_atomic_store_release_ptr
+   fun:upstreamOpenIPCHandle
    fun:trackingOpenIpcHandle
    fun:umfMemoryProviderOpenIPCHandle
    fun:umfOpenIPCHandle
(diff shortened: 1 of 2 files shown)


ldorau (approved):
This patch fixes issues like this one:
https://github.com/oneapi-src/unified-memory-framework/actions/runs/13894590051/job/38874344510?pr=1185

test/supp/helgrind-test_ipc.supp
```diff
@@ -19,6 +19,7 @@
    [false-positive] Double check locking pattern in trackingOpenIpcHandle
    Helgrind:Race
    fun:utils_atomic_store_release_ptr
+   fun:upstreamOpenIPCHandle
```
bratpiorka, line 22: Wouldn't this disable suppression for false positive issues with a backtrace that lacks upstreamOpenIPCHandle? Shouldn't we use "..." here?

vinser52: But the `utils_atomic_store_release_ptr` is called only inside the `upstreamOpenIPCHandle` function.
Probably, I did not get your point.

KFilipek (approved):

PR #1198, opened 2025-03-17
fix jemalloc benchmark with fixedprovider

lplewa:
### Checklist

- [ ] All tests pass locally


ldorau (changes requested):
Could you describe in commit message what this patch changes and why?

lplewa:
> Could you describe in commit message what this patch changes and why?

Is just fix for a typo

ldorau:
> > Could you describe in commit message what this patch changes and why?
> 
> Is just fix for a typo

so why did it work with this "typo" ?

bratpiorka (approved):

lplewa:
> > > Could you describe in commit message what this patch changes and why?
> > 
> > 
> > Is just fix for a typo
> 
> so why did it work with this "typo" ?

We had benchmarks with os_provider. I copied them, and updated to use fixed provider. In one place i did not renamed benchmark object in register function. So we registered jemalloc<os_provider> twice instead, and benchmark object for jemalloc<fixed_provider> was not registered/used.

lplewa:
Before the patch
image
After the patch
image

ldorau (approved):

PR #1199, opened 2025-03-18
limit benchmark test to singlethreaded one only

lplewa:
### Checklist

- [ ] All tests pass locally


kswiecicki:
Does this filter those benchmarks? https://github.com/oneapi-src/unified-memory-framework/blob/4c543661c9810d2ff7b013a6ca6229f36044d7c3/benchmark/benchmark.cpp#L62
 If so, it might be more appropriate to remove `->Apply(&multithreaded);` when registering the benchmarks.

lplewa:
> Does this filter those benchmarks?
> 
> https://github.com/oneapi-src/unified-memory-framework/blob/4c543661c9810d2ff7b013a6ca6229f36044d7c3/benchmark/benchmark.cpp#L62
> 
> If so, it might be more appropriate to remove `->Apply(&multithreaded);` when registering the benchmarks.

Well we want to have multithreded benchmark as part of benchmarks workflow. This PR is only to disable MT benchmarks as part of "ctest" command

benchmark/CMakeLists.txt
```diff
@@ -40,8 +40,9 @@ function(add_umf_benchmark)
     # * SRCS - source files
     # * LIBS - libraries to be linked with
     # * LIBDIRS - directories of libraries to be linked with
+    # * EXTRARGS - additional arguments to be passed to the add_test
     set(oneValueArgs NAME)
-    set(multiValueArgs SRCS LIBS LIBDIRS)
+    set(multiValueArgs SRCS LIBS LIBDIRS EXTRARGS)
```
bratpiorka, line 45: please update .cmake-format file

benchmark/CMakeLists.txt
```diff
@@ -148,7 +149,7 @@ add_umf_benchmark(
     NAME benchmark
     SRCS benchmark.cpp
     LIBS ${LIBS_OPTIONAL} benchmark::benchmark
-    LIBDIRS ${LIB_DIRS})
+    LIBDIRS ${LIB_DIRS} EXTRARGS --benchmark_filter=threads:1$)
```
bratpiorka, line 152: please add a comment "limit running benchmarks in CI tests to single-threaded"

bratpiorka (approved):

benchmark/CMakeLists.txt
lplewa: Done

KFilipek (approved):

PatKamin (approved):
Much shorter benchmark tests, thanks!

PR #1208, opened 2025-03-19
[CI] Bump L0 version

lukaszstolarczuk:
It contains static loader now. It should work with any loader installed in the system (e.g. via apt) as long as it's >= v1.15.17

// ref. https://github.com/intel/llvm/pull/17523


lukaszstolarczuk:
they discarded this change in SYCL, due to some errors, so closing this here for now
// https://github.com/intel/llvm/pull/17531

PR #1220, opened 2025-03-25
Small fixes around CI

lukaszstolarczuk:
Each commit with its own fix/change.

Most importantly - update the tokens to the least privileged ones by default.


.github/workflows/coverity.yml
```diff
@@ -35,11 +35,11 @@ jobs:
 
     - name: Download Coverity
       run: |
```
PatKamin, line 37: You could use `>` instead for a multiline command. This would increase readability even more.

PatKamin (approved):

bratpiorka (approved):

KFilipek (approved):

PR #1229, opened 2025-03-31
Check for autoreconf prerequisite in case hwloc is fetched from sources

PatKamin:
### Description
Check for `autoreconf` prerequisite in case hwloc is fetched from sources.

### Checklist

- [x] All tests pass locally

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -245,37 +245,45 @@ else()
     )
 endif()
 
+if(NOT UMF_DISABLE_HWLOC AND (NOT UMF_LINK_HWLOC_STATICALLY))
+    pkg_check_modules(LIBHWLOC hwloc>=2.3.0)
+    if(NOT LIBHWLOC_FOUND)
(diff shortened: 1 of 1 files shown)


bratpiorka (approved):

CMakeLists.txt
```diff
@@ -245,37 +245,45 @@ else()
...
+        message(WARNING "autoreconf is not installed. Disabling hwloc.")
+        set(UMF_DISABLE_HWLOC ON)
+        set(UMF_LINK_HWLOC_STATICALLY OFF)
     endif()
+endif()
 
+if(UMF_DISABLE_HWLOC)
+    message(STATUS "hwloc is disabled, hence OS provider, memtargets, "
+                   "topology discovery, examples won't be available!")
+else()
```
KFilipek, line 278: if user decided to `DISABLE_HWLOC` but `LINK_HWLOC_STATICALLY` is enabled then, screw it, let's link it. It's another example why this flags should be unified.

PR #1247, opened 2025-04-09
Fix file_alloc_aligned() and add the allocFreeAligned test to poolFixtures.hpp

ldorau:
### Description

Fix file_alloc_aligned() and add the allocFreeAligned test to poolFixtures.hpp.

### Checklist

- [x] All tests pass locally

### DIFF
--- a/src/provider/provider_file_memory.c
+++ b/src/provider/provider_file_memory.c
@@ -429,6 +429,8 @@ static umf_result_t file_alloc_aligned(file_memory_provider_t *file_provider,
         return UMF_RESULT_ERROR_UNKNOWN;
     }
 
+    assert(file_provider->offset_mmap <= file_provider->size_mmap);
+
     if (file_provider->size_mmap - file_provider->offset_mmap < size) {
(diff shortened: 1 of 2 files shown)


src/provider/provider_file_memory.c
```diff
@@ -454,7 +454,7 @@ static umf_result_t file_alloc_aligned(file_memory_provider_t *file_provider,
     size_t new_offset_fd =
         file_provider->offset_fd + new_offset_mmap - file_provider->offset_mmap;
 
-    if (file_provider->size_mmap - new_offset_mmap < size) {
+    if (file_provider->size_mmap < size + new_offset_mmap) {
```
bratpiorka, line 457: so there was an underflow here? I see similar case at line 432 - should it also be fixed?

ldorau: 1) Yes, there was an underflow here.
2) No, there should be always: `file_provider->offset_mmap <= file_provider->size_mmap` - I added an assert for that.

Issue #1078, opened 2025-02-05
Backward Compatibility for OPS Structures


lplewa:
Our OPS structures for pools and providers are currently not backward compatible, even though they include a `version` field. The following are four separate requirements we must fulfill:

1. **User calls `umfDisjointPoolOps()`** and passes the resulting structure to `umfCreate()`.  
   *(No manual struct creation—just calling the UMF-provided function.)*

2. **User creates their own provider/pool structure** (manually defining the fields) and passes it to `umfCreate()`.  
   *(We need to detect the old structure version and handle it properly.)*

3. **User obtains our OPS structure**, duplicates it (it SHOULD be const), and modifies it.  
   *(Challenge: if the new UMF version’s OPS structure has a different size, the user doesn’t know how much memory to allocate.)*

4. **User compiles a program (written against an older UMF release) without any source changes**, with a new UMF version.  
   *(The program should compile without problems and continue to work.)*

---

### Option 1: Make OPS Structure Opaque

- **Concept**  
  - Remove the public definition of the OPS structure.  
  - Provide creation, duplication, and freeing functions (e.g., `umfPoolOpsCreate()`, `umfPoolOpsDup()`, `umfPoolOpsDestroy()`).  
  - Provide setter/getter functions for each field (there are currently ~27 fields).

- **Advantages**  
  - Future changes do not break binary compatibility: fields can be added or removed internally.  
  - Removing a field just turns the associated getter/setter into a no-op.

- **Disadvantages**  
  - **Backward Compatibility**: This **breaks** existing code if it directly references the struct fields.  
  - **User Adoption Effort**: **High**. Users must replace direct struct assignments with multiple setter/getter calls.  
  - A large number of new APIs (at least 2× the number of fields, and we have 27 atm).

---

### Option 2: Keep Old Versions Internally

We maintain multiple versions of the OPS structure and the functions that return them. That way, older user code can call older “versions” and still work. There are two sub-approaches:

#### Option 2a: Symbol Versioning (Linker Script)

- **Concept**  
  - Use a linker script with versioned symbols. For instance, `umfDisjointPoolOps` might exist in multiple versions: `_umfDisjointPoolOps@UMF_1.0`, `_umfDisjointPoolOps@UMF_2.0`, etc.  
  - Older binaries link to the older version automatically, newer binaries link to the newer symbol.

- **Advantages**  
[...]

PR #1265, opened 2025-04-15
make ops structure flat

lplewa:
fixes: #1078

### Checklist

- [ ] All tests pass locally


include/umf/memory_pool_ops.h
```diff
@@ -28,6 +28,9 @@ extern "C" {
 /// pointers.
 ///
 typedef struct umf_memory_pool_ops_t {
+    /// Size of this structure.
+    /// Should be initialized with sizeof(umf_memory_pool_ops_t) in the current umf version
+    size_t size;
```
vinser52, line 33: Why `version` is not enough?

lplewa: As user do not know mapping between size and version. As this structure might be bigger then user is aware (as it might be extended, in future umf versions), size field is needed if user needs to copy this structure.  This is why this pr adds flexible array member at the end of the struct. 

Adding size field was a conclusion of the linked issue.

vinser52: > As user do not know mapping between size and version. As this structure might be bigger then user is aware (as it might be extended, in future umf versions), size field is needed if user needs to copy this structure. This is why this pr adds flexible array member at the end of the struct.
> 
> Adding size field was a conclusion of the linked issue.

If I remember we already discussed this issue, and if I am not wrong we have not found the use case where it is needed. Could you please post a use case here where this size variable is used?

vinser52: BTW, I am not against that, but I just do not understand why it is needed.

lplewa: TBH i found this action step in final notes after discusion, in the issue, so i implemented it.

The use case is when the user want to add some extra function to our provider/pool. 
To do so:
```c
const umf_memory_pool_ops_t *ops = umfDisjointPoolOps();

umf_memory_pool_ops_t *my_ops = malloc(ops.size);
memcpy(my_ops, ops, ops.size);
my_ops.malloc = createMyCustomMallocWrapper(ops.alloc);
```

ofc this is also possible without size, but to do it in backward compatible way, user must reset version field (which let's be honest no one will do, and after umf update we will be blamed for compatibility break)

vinser52: > ```c
> const umf_memory_pool_ops_t *ops = umfDisjointPoolOps();
> 
> umf_memory_pool_ops_t *my_ops = malloc(ops.size);
> memcpy(my_ops, ops, ops.size);
> my_ops.malloc = createMyCustomMallocWrapper(ops.alloc);
> ```

I think it is not a proper way to implement a wrapper on top of some memory provider. The proper way is to create a wrapper memory provider that will use a handle to another memory provider. This how our tracking provider is implemented. Another example is a tracing provider in our tests.

We should keep in mind that the proper way to use the memory provider is by creating a memory provider handle and use it via Memory Provider API, e.g. `umfMemoryProviderAlloc`, `umfMemoryProviderFree`.

The `ops` structures are just an API for Memory provider/pools developers provided by UMF (something like a plugin interface). The `ops` structures are intended to be used by UMF, not by someone else. Via `ops` structure we are telling to  Memory provider/pools developers "if you want UMF to use your custom provider please give us the `ops` structure".

lplewa: i removed size - Done

src/memory_provider.c
```diff
@@ -120,67 +120,78 @@ static umf_result_t umfDefaultCtlHandle(void *provider, int operationType,
...
 
-static bool validateOpsMandatory(const umf_memory_provider_ops_t *ops) {
-    // Mandatory ops should be non-NULL
-    return ops->alloc && ops->free && ops->get_recommended_page_size &&
-           ops->get_min_page_size && ops->initialize && ops->finalize &&
-           ops->get_last_native_error && ops->get_name;
-}
+#define CHECK_OP(ops, fn)                                                      \
+    if (!(ops)->fn) {                                                          \
+        LOG_ERR("Error: missing function pointer: %s\n", #fn);                 \
```
bratpiorka, line 160: nit: LOG_ERR already print "error" string, so you could leave only "missing... "

ldorau (approved):

src/memory_pool.c
```diff
@@ -84,8 +85,8 @@ static umf_result_t umfPoolCreateInternal(const umf_memory_pool_ops_t *ops,
     pool->ops = *ops;
     pool->tag = NULL;
 
-    if (NULL == pool->ops.ctl) {
-        pool->ops.ctl = umfDefaultCtlPoolHandle;
+    if (NULL == pool->ops.ext_ctl) {
+        pool->ops.ext_ctl = umfDefaultCtlPoolHandle;
```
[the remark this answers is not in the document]
lplewa: I was considering this - when we have only one default imho this is overkill to introduce extra function, and it will be better to refactor it when we will add second/third one.

src/memory_provider.c
lplewa: Done

ldorau:
@lplewa please resolve conflicts

Issue #1233, opened 2025-04-02
AddressSanitizer: unknown-crash in umfMemoryTrackerAdd (in utils_atomic_load_acquire_u64())


ldorau:
**ERROR: AddressSanitizer: unknown-crash on address** :
https://github.com/ldorau/unified-memory-framework/actions/runs/14172971689/job/39700871936
```
[ RUN      ] disjointPoolTests/umfPoolTest.multiThreadedpow2AlignedAlloc/0
=================================================================
==1159214==ERROR: AddressSanitizer: unknown-crash on address 0x7f3e9feac248 at pc 0x5621d13ce718 bp 0x7f3e9a9fea00 sp 0x7f3e9a9fe9f0
READ of size 8 at 0x7f3e9feac248 thread T9
    #0 0x5621d13ce717 in utils_atomic_load_acquire_u64 /home/runner/work/unified-memory-framework/unified-memory-framework/src/utils/utils_concurrency.h:166
    #1 0x5621d13cfe1c in umfMemoryTrackerAdd /home/runner/work/unified-memory-framework/unified-memory-framework/src/provider/provider_tracking.c:202
    #2 0x5621d13d29b0 in trackingAlloc /home/runner/work/unified-memory-framework/unified-memory-framework/src/provider/provider_tracking.c:481
    #3 0x5621d13caa1e in umfMemoryProviderAlloc /home/runner/work/unified-memory-framework/unified-memory-framework/src/memory_provider.c:245
    #4 0x5621d13eaf23 in disjoint_pool_aligned_malloc /home/runner/work/unified-memory-framework/unified-memory-framework/src/pool/pool_disjoint.c:745
    #5 0x5621d13c7ea2 in umfPoolAlignedMalloc /home/runner/work/unified-memory-framework/unified-memory-framework/src/memory_pool.c:195
    #6 0x5621d12b6d99 in pow2AlignedAllocHelper(umf_memory_pool_t*) /home/runner/work/unified-memory-framework/unified-memory-framework/test/poolFixtures.hpp:190
    #7 0x5621d12b8cf1 in operator() /home/runner/work/unified-memory-framework/unified-memory-framework/test/poolFixtures.hpp:250
[...]

PR #1270, opened 2025-04-15
Postpone freeing a tracker entry until it is really removed from tracker

ldorau:
### Description

Postpone freeing a tracker entry until it is really removed from tracker.

Ref: #1233

### Checklist

- [ ] All tests pass locally


ldorau:
@lplewa @lukaszstolarczuk _"Exception: The directory /home/test-user/bench_workdir_umf exists but is not a benchmark work directory."_
See: https://github.com/oneapi-src/unified-memory-framework/actions/runs/14473372398/job/40592824780
@lplewa @lukaszstolarczuk Could you fix it?

ldorau:
The last commit https://github.com/oneapi-src/unified-memory-framework/pull/1270/commits/fa661de4f672d2555d2132f3ba4b5bc7b8fd4138 is not finished yet !

vinser52:
What is the purpose of this PR, could you please describe the scenario and how these changes help?

lplewa:
> What is the purpose of this PR, could you please describe the scenario and how these changes help?

We have a race, between insert and delete of entry in tracker

image
Top screen in screen one, bottom screen is thread two. Numbers shows order of the operations.

lplewa:
Pool nesting detection. We are checking if there is region in the tracker that contains region that we are adding. So we are looking for region with address lower then our - we are finding one, and in the exact moment other thread removes this region which we found.

vinser52:
> Pool nesting detection. We are checking if there is region in the tracker that contains region that we are adding. So we are looking for region with address lower then our - we are finding one, and in the exact moment other thread removes this region which we found.

That is exactly my question. How such a scenario might happen: if T1 has a pointer P, that belongs to some region R, and T2 removes this region R.

ldorau:
> > Pool nesting detection. We are checking if there is region in the tracker that contains region that we are adding. So we are looking for region with address lower then our - we are finding one, and in the exact moment other thread removes this region which we found.
> 
> That is exactly my question. How such a scenario might happen: if T1 has a pointer P, that belongs to some region R, and T2 removes this region R.

@vinser52 @lplewa The explanation is that the pointer P (of T1) does not belong to the region R (of T2):

a) thread **T2** allocates memory region R (for example: `0x100-0x300` - address `0x100` and size `0x200`) and adds it to the tracker (this is the only entry in the tracker yet) and gets preempted ...
b) thread **T1** allocates memory of size `0x500` and receives pointer P (address `0x400`, size `0x500` - range `0x400-0x900`) and tries to add it to the tracker, so it calls `umfMemoryTrackerAdd()`, so `critnib_find(FIND_LE)` finds and returns the only region R existing in the tracker R: `0x100-0x300` (step 1 on the picture) and gets preempted (step 2 on the picture) ...
c) thread **T2** removes the region R from the tracker (step 3 on the picture), **frees its memory** (step 4 on the picture) and gets preempted ...
d) thread **T1** tries to read a size of the already freed region R (step 5 on the picture) and blows up ...

ldorau:
@bratpiorka please review

ldorau:
See looped sanitizers and valgrind builds:
https://github.com/ldorau/unified-memory-framework/actions/runs/14905978766
https://github.com/ldorau/unified-memory-framework/actions/runs/14905981234
https://github.com/ldorau/unified-memory-framework/actions/runs/14906349518

ldorau:
See looped sanitizers and valgrind builds:
https://github.com/ldorau/unified-memory-framework/actions/runs/14905978766
https://github.com/ldorau/unified-memory-framework/actions/runs/14906349518
https://github.com/ldorau/unified-memory-framework/actions/runs/14923529489
https://github.com/ldorau/unified-memory-framework/actions/runs/14923882709

ldorau:
See looped sanitizers and valgrind builds:
https://github.com/ldorau/unified-memory-framework/actions/runs/14905978766
https://github.com/ldorau/unified-memory-framework/actions/runs/14906349518
https://github.com/ldorau/unified-memory-framework/actions/runs/14923529489
https://github.com/ldorau/unified-memory-framework/actions/runs/14923882709

ldorau:
See looped sanitizers and valgrind builds **with this fix** (all 24 jobs timed out after 6 hours, none of them failed):
https://github.com/ldorau/unified-memory-framework/actions/runs/14905978766
https://github.com/ldorau/unified-memory-framework/actions/runs/14906349518
https://github.com/ldorau/unified-memory-framework/actions/runs/14923529489
https://github.com/ldorau/unified-memory-framework/actions/runs/14923882709

While **on the current main** branch (21 jobs failed, only 3 jobs timed out after 6 hours):
https://github.com/ldorau/unified-memory-framework/actions/runs/14930716058
https://github.com/ldorau/unified-memory-framework/actions/runs/14930734240
https://github.com/ldorau/unified-memory-framework/actions/runs/14930742598
https://github.com/ldorau/unified-memory-framework/actions/runs/14930750871

ldorau:
Please review

ldorau:
@bratpiorka @lplewa please review

src/critnib/critnib.c
```diff
@@ -162,7 +168,7 @@ static inline unsigned slice_index(word key, sh_t shift) {
 /*
  * critnib_new -- allocates a new critnib structure
  */
-struct critnib *critnib_new(void) {
+struct critnib *critnib_new(free_leaf_t cb_free_leaf, void *leaf_allocator) {
```
bratpiorka, line 171: please add a description of args in the comment above

src/critnib/critnib.c
```diff
@@ -189,6 +197,14 @@ struct critnib *critnib_new(void) {
  */
 static void delete_node(struct critnib *c, struct critnib_node *__restrict n) {
     if (is_leaf(n)) {
+        if (c->cb_free_leaf && to_leaf(n)) {
```
bratpiorka, line 200: please add a comment that we are calling a callback here

src/critnib/critnib.c
```diff
@@ -290,8 +315,20 @@ static struct critnib_leaf *alloc_leaf(struct critnib *__restrict c) {
...
     c->deleted_leaf = k->value;
+
+    uint64_t ref_count;
+    void *to_be_freed;
+    void *key;
+    utils_atomic_load_acquire_u64(&k->ref_count, &ref_count);
+    utils_atomic_load_acquire_ptr(&k->to_be_freed, &to_be_freed);
+    utils_atomic_load_acquire_ptr((void **)&k->key, &key);
+    if (ref_count && to_be_freed) {
+        LOG_DEBUG("WARNING: reusing probably non-freed leaf: key = %p, value = "
```
bratpiorka, line 327: should we also put an assert here that would trigger in debug mode?

src/critnib/critnib.c
```diff
@@ -399,9 +446,14 @@ int critnib_insert(struct critnib *c, word key, void *value, int update) {
 /*
  * critnib_remove -- delete a key from the critnib structure, return its value
  */
-void *critnib_remove(struct critnib *c, word key) {
+void *critnib_remove(struct critnib *c, word key, void **ref) {
```
bratpiorka, line 449: add info about "ref" to the comment above

src/critnib/critnib.c
```diff
@@ -399,9 +446,14 @@ int critnib_insert(struct critnib *c, word key, void *value, int update) {
...
     struct critnib_leaf *k;
     void *value = NULL;
+    word kkey;
+
+    if (!ref) {
+        return NULL;
```
bratpiorka, line 455: please comment

src/critnib/critnib.c
```diff
@@ -410,16 +462,34 @@ void *critnib_remove(struct critnib *c, word key) {
...
 
-    word del =
-        (utils_atomic_increment_u64(&c->remove_count) - 1) % DELETED_LIFE;
+    word del;
+    int i_del = 0;
+    uint64_t ref_count = 0;
+    // We look for a leaf with ref_count == 0.
+    // If we do not find one we take the first one that was checked.
+    // In such case (extremely unlikely) we might not be able to free the value
+    // and a memory leak might happen.
```
bratpiorka, line 471: add a log message in this case

src/critnib/critnib.c
```diff
@@ -626,19 +775,42 @@ static struct critnib_leaf *find_le(struct critnib_node *__restrict n,
...
         utils_atomic_load_acquire_u64(&c->remove_count, &wrs2);
     } while (wrs1 + DELETED_LIFE <= wrs2);
 
+    if (res) {
+        uint64_t ref_count;
+        utils_atomic_load_acquire_u64(&k->ref_count, &ref_count);
+        if (ref_count == 0) {
+            return NULL;
+        }
+        if (utils_atomic_increment_u64(&k->ref_count) == 1) {
```
bratpiorka, line 806: ???? what is happening here

src/critnib/critnib.h
```diff
@@ -27,19 +28,25 @@ enum find_dir_t {
...
+// The following 4 functions:
+void *critnib_remove(critnib *c, uintptr_t key, void **ref);
+void *critnib_get(critnib *c, uintptr_t key, void **ref);
+void *critnib_find_le(critnib *c, uintptr_t key, void **ref);
+int critnib_find(critnib *c, uintptr_t key, enum find_dir_t dir,
+                 uintptr_t *rkey, void **rvalue, void **ref);
+// ... returns a reference (void *ref) to the key,
+// that MUST be released by calling critnib_release():
+int critnib_release(struct critnib *c, void *ref);
+// when the returned value is no longer used and can be freed using the cb_free_leaf callback.
```
bratpiorka, line 48: IMO these comments should be grouped into one continuous section

src/pool/pool_disjoint.c
```diff
@@ -233,7 +231,11 @@ static umf_result_t pool_unregister_slab(disjoint_pool_t *pool, slab_t *slab) {
     // TODO ASSERT_IS_ALIGNED((uintptr_t)slab_addr, bucket->size);
     LOG_DEBUG("slab: %p, start: %p", (void *)slab, slab_addr);
 
-    critnib_remove(slabs, (uintptr_t)slab_addr);
+    void *ref_slabs = NULL;
+    critnib_remove(slabs, (uintptr_t)slab_addr, &ref_slabs);
```
bratpiorka, line 235: What do you think about an API change where, if I pass NULL as the last argument, the `critnib_release` would be called automatically at the end of the `critnib_remove` call? This would mean the change here would simply be `critnib_remove(slabs, (uintptr_t)slab_addr, NULL);`

src/pool/pool_disjoint.c
```diff
@@ -570,6 +572,13 @@ static void *disjoint_pool_allocate(disjoint_pool_t *pool, size_t size) {
     return ptr;
 }
 
+static void free_slab(void *unused, void *slab) {
```
bratpiorka, line 575: please comment when this function would be called

src/critnib/critnib.c
ldorau: I could but I am not sure if we should. 1) It is extremely unlikely and 2) it might cause only a memory leak of one value that IMHO is a good trade-off versus further complicating and decreasing performance of critnib. What do you think?

src/critnib/critnib.c
ldorau: It is verified and logged in `alloc_leaf()`.

src/critnib/critnib.c
ldorau: We check if the value has just been removed (i.e. if `(ref_count == 0)`).
If so, we return NULL.

A comment added.

src/pool/pool_disjoint.c
ldorau: It would be not consistent and misleading. We can add `critnib_remove_release(slabs, (uintptr_t)slab_addr);` instead.

src/provider/provider_tracking.c
```diff
@@ -76,25 +81,60 @@ static tracker_alloc_info_t *get_most_nested_alloc_segment(
...
+                break;
+            }
+
+            level++;
+
+            if (_parent_key) {
                 parent_key = rkey;
+            }
+
+            if (_parent_value) {
```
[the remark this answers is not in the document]
ldorau: Done

src/pool/pool_disjoint.c
ldorau: Done

src/critnib/critnib.h
ldorau: Done

src/critnib/critnib.c
ldorau: Done in the comment to `critnib_remove()`

src/critnib/critnib.c
ldorau: Done

src/critnib/critnib.c
ldorau: Done

src/critnib/critnib.c
ldorau: Done

src/pool/pool_disjoint.c
bratpiorka: I'm ok with critnib_remove_release

src/critnib/critnib.c
ldorau: Done

src/pool/pool_disjoint.c
ldorau: `critnib_remove_release()` added.
Done

src/critnib/critnib.c
ldorau: Hmm ... yes, right, it may lead to UB, so even a crash cannot be excluded.
I added an assert here.
Done.

src/critnib/critnib.c
```diff
@@ -111,15 +111,21 @@ struct critnib_node {
     struct critnib_node *child[SLNODES];
     word path;
     sh_t shift;
+    // padding to 8 bytes so that shift can be stored atomically as uint64_t
+    sh_t unused[7];
```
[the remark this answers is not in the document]
ldorau: It works now.
Done.

src/critnib/critnib.c
```diff
@@ -290,8 +318,34 @@ static struct critnib_leaf *alloc_leaf(struct critnib *__restrict c) {
...
+    // In such case (extremely unlikely) a memory leak might happen.
+    uint64_t ref_count;
+    utils_atomic_load_acquire_u64(&k->ref_count, &ref_count);
+    if (ref_count > 0) {
+        void *to_be_freed;
+        utils_atomic_load_acquire_ptr(&k->to_be_freed, &to_be_freed);
+        if (to_be_freed != NULL) {
+            void *key;
+            utils_atomic_load_acquire_ptr((void **)&k->key, &key);
+            LOG_DEBUG("WARNING: reusing probably non-freed leaf: key = %p, "
```
bratpiorka, line 333: nit: this message could be moved to #else so in debug there would be only one message

bratpiorka (approved):

bratpiorka:
@lplewa please review

src/critnib/critnib.c
ldorau: This code has been removed

ldorau: Done

ldorau:
@lplewa please review

ldorau:
@lplewa please review

ldorau:
@lplewa please review

src/critnib/critnib.c
```diff
@@ -472,13 +555,90 @@ void *critnib_remove(struct critnib *c, word key) {
...
+    utils_atomic_load_acquire_u64(&k->ref_count, &ref_count);
+
+    if (ref_count == 0) {
+        return -1;
+    }
+
+    /* decrement the reference count */
+    if (utils_atomic_decrement_u64(&k->ref_count) == 0) {
+        void *to_be_freed = NULL;
+        utils_atomic_load_acquire_ptr(&k->to_be_freed, &to_be_freed);
```
lplewa, line 588: Do we need atomic oprations here. We drop ref_count to zero no one will do anything with this ptrs

src/critnib/critnib.c
```diff
@@ -472,13 +555,90 @@ void *critnib_remove(struct critnib *c, word key) {
...
+        if (to_be_freed) {
+            utils_atomic_store_release_ptr(&k->to_be_freed, NULL);
+            if (c->cb_free_leaf) {
```
lplewa, line 591: Why we do refcounting if CB is NULL? Is it just a waste of the cpu cycles for all refcounting? 

imho If CB is null, we just should never update refcount.

src/critnib/critnib.c
```diff
@@ -472,13 +555,90 @@ void *critnib_remove(struct critnib *c, word key) {
...
+    uint64_t ref_count;
+    utils_atomic_load_acquire_u64(&k->ref_count, &ref_count);
+    if (ref_count == 0) {
+        return -1;
+    }
+
+    if (utils_atomic_increment_u64(&k->ref_count) == 1) {
+        utils_atomic_decrement_u64(&k->ref_count);
+        return -1;
+    }
```
lplewa, line 637: Why we check twice if ref_count is zero? 


also isn't it a race, between increment and decrement? If between both operation we have very long sleep, this node might be all ready reused, and count be different.

src/critnib/critnib.c
```diff
@@ -472,13 +555,90 @@ void *critnib_remove(struct critnib *c, word key) {
...
+
+#ifndef NDEBUG
+    // check if the reference count is overflowed
+    utils_atomic_load_acquire_u64(&k->ref_count, &ref_count);
+    assert((ref_count & (1ULL << 63)) == 0);
+    assert(ref_count != (uint64_t)(0 - 1ULL));
+#endif
+
+    return 0;
```
lplewa, line 604: What happens with the leaf if ref count drop to zero, shouldn't we put it on the list and  reuse it?

lplewa (changes requested):
to be continued....

src/critnib/critnib.c
```diff
@@ -90,7 +90,7 @@
 #define SLNODES (1 << SLICE)
 
 typedef uintptr_t word;
-typedef unsigned char sh_t;
+typedef uint64_t sh_t;
```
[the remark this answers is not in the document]
ldorau: So that we could use 64-bit atomic load/store on it.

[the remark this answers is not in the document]
lplewa: There are atomic functions for each variable size. If we are missing one, in our utils, we can always add it

src/critnib/critnib.c
ldorau: Yes, we do. Sanitizers show data race here (maybe false-positive but the build fails):
https://github.com/ldorau/unified-memory-framework/actions/runs/15147983132/job/42588195837

src/critnib/critnib.c
ldorau: It is done in the `free_leaf()` function.
Done.

src/critnib/critnib.c
```diff
@@ -189,6 +200,15 @@ struct critnib *critnib_new(void) {
  */
 static void delete_node(struct critnib *c, struct critnib_node *__restrict n) {
     if (is_leaf(n)) {
+        // call the callback freeing the leaf
+        if (c->cb_free_leaf && to_leaf(n)) {
```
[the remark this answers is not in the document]
ldorau: Right. `to_leaf(n)` removed.
Done

src/critnib/critnib.c
```diff
@@ -116,14 +116,19 @@ struct critnib_node {
 struct critnib_leaf {
     word key;
     void *value;
+    void *to_be_freed;
```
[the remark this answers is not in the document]
ldorau: Done

src/critnib/critnib.c
ldorau: OK, fixed.
Done

src/critnib/critnib.c
ldorau: `ref_count` can be incremented only if `ref_count > 0` - that's why the first check is needed.
After having incremented `ref_count`, we have to check if `ref_count == 1` to make sure it wasn't equal 0 before incrementing. If `ref_count == 1` we have to decrement it and exit. There is no other way to handle it.

ldorau:
@lplewa re-review please

src/critnib/critnib.c
lplewa: So why we have first check second check anyway

lplewa: Also you should use compare and swap instead increment and decrement to eliminate  race

ldorau: Do you mean like that?
```c
static inline int increment_ref_count(struct critnib_leaf *k) {
    uint64_t expected;
    uint64_t desired;

    do {
        utils_atomic_load_acquire_u64(&k->ref_count, &expected);
        if (expected == 0) {
            return -1;
        }
        desired = expected + 1;
    } while (!utils_compare_exchange_u64(&k->ref_count, &expected, &desired));

    return 0;
}
```

src/critnib/critnib.c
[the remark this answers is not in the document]
ldorau: Done

ldorau:
@lplewa re-review please

src/critnib/critnib.c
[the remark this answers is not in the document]
ldorau: OK, so please tell me how can I do 8-bit atomic load on Windows?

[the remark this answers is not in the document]
lplewa: There is _InterlockedCompareExchange8

src/critnib/critnib.c
ldorau: Done

ldorau:
@lplewa re-review please

src/critnib/critnib.c
```diff
@@ -277,8 +320,22 @@ static void free_leaf(struct critnib *__restrict c,
...
+            return;
+        }
+
+        // ref_count > 0, so add k to the pending_del_leaf list
+        utils_atomic_store_release_ptr((void **)&k->key, c->pending_del_leaf);
+        utils_atomic_store_release_ptr((void **)&c->pending_del_leaf, k);
+    } else {
+        utils_atomic_store_release_ptr((void **)&k->value, c->deleted_leaf);
+        utils_atomic_store_release_ptr((void **)&c->deleted_leaf, k);
+    }
```
lplewa, line 338: if (!cb_free_leaf)
ref_count = 0
else 
atomic_load......

src/critnib/critnib.c
```diff
@@ -277,8 +320,22 @@ static void free_leaf(struct critnib *__restrict c,
...
     }
 
-    utils_atomic_store_release_ptr((void **)&k->value, c->deleted_leaf);
-    utils_atomic_store_release_ptr((void **)&c->deleted_leaf, k);
+    if (c->cb_free_leaf) {
+        uint64_t ref_count;
+        utils_atomic_load_acquire_u64(&k->ref_count, &ref_count);
+        if (ref_count == 0) {
+            utils_atomic_store_release_ptr(&k->value, c->deleted_leaf);
+            utils_atomic_store_release_ptr((void **)&c->deleted_leaf, k);
```
lplewa, line 328: This is not how corectly we update lock free single headed list. 

If two threads adds to this list concurently they can have the same "next" node

src/critnib/critnib.c
```diff
@@ -410,16 +488,40 @@ void *critnib_remove(struct critnib *c, word key) {
...
+                break;
+            }
+
+            k = c->pending_del_leaves[del];
+            if (k) {
+                utils_atomic_load_acquire_u64(&k->ref_count, &ref_count);
+            }
+        } while (k && (ref_count > 0));
+    }
+
```
lplewa, line 515: I'm not huge fan of this algoritm. 
- As comment says it can memleak
- Causes that remove count can grow faster, what can reduce performance, as it will cause more retries. 

Can we just use old algorithm then:
- if refcount is non zero we can just remove this node from deleted array(but not free value yet)
- and when refcount == 0, at last unref, we can add this node back to the deleted(free) list.

src/critnib/critnib.c
```diff
@@ -116,14 +116,27 @@ struct critnib_node {
...
     void *value;
+    /*
+     * When cb_free_leaf() is set and values of leaves should be freed,
+     * critnib_remove() sets to_be_freed to the value of the leaf
+     * and sets value to NULL, because the value of a removed leaf
+     * is not valid anymore and should not be returned to a user.
+     * The value of the leaf saved in to_be_freed is freed in critnib_release()
+     * when the reference count of the leaf drops to 0.
+     */
+    void *to_be_freed;
```
lplewa, line 127: i do not see a reason/usecase to have two separate ptrs to keep the same data.  Are you sure that we need two ptrs?

lplewa (changes requested):

src/critnib/critnib.c
ldorau: The reason is described above in the comment. If value is set it will be returned to a user. We have to move it somewhere else to be able to free it when the reference counter drops to 0. Yes, I am sure.

src/critnib/critnib.c
ldorau: This is the way critnib has been doing it so far and everybody have considered critnib very thread-safe so far.
How do you propose to fix it?

src/critnib/critnib.c
ldorau: Done

src/critnib/critnib.c
ldorau: Done

src/critnib/critnib.c
```diff
@@ -116,14 +116,27 @@ struct critnib_node {
...
 
 struct critnib {
     struct critnib_node *root;
+    free_leaf_t cb_free_leaf; // callback for freeing a leaf
+    void *leaf_allocator;     // handle of allocator for leaves
 
     /* pool of freed nodes: singly linked list, next at child[0] */
     struct critnib_node *deleted_node;
     struct critnib_leaf *deleted_leaf;
+    struct critnib_leaf *pending_del_leaf;
```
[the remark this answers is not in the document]
ldorau: Removed.
Done.

src/critnib/critnib.c
ldorau: Removed.
Done.

src/critnib/critnib.c
ldorau: Done

ldorau:
@lplewa re-review please

ldorau:
Rebased

lplewa (approved):

PR #1271, opened 2025-04-15
[CI] Install hwloc from apt instead of building from sources

PatKamin:
Get rid of the hwloc installation script runs in CI. Save ca. 25s in each of 14 jobs using this script.

### Checklist

- [ ] All tests pass locally

### DIFF
--- a/.github/workflows/reusable_basic.yml
+++ b/.github/workflows/reusable_basic.yml
@@ -126,6 +126,11 @@ jobs:
         sudo apt-get update
         sudo apt-get install -y clang cmake libnuma-dev lcov
 
+    - name: Install hwloc
+      if: matrix.disable_hwloc == 'OFF'
+      run: |
+        sudo apt-get install -y libhwloc-dev
+
     - name: Install TBB apt package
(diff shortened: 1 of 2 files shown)


.github/workflows/reusable_basic.yml
```diff
@@ -124,7 +124,7 @@ jobs:
     - name: Install apt packages
       run: |
         sudo apt-get update
-        sudo apt-get install -y clang cmake libnuma-dev lcov
+        sudo apt-get install -y clang cmake libhwloc-dev libnuma-dev lcov
```
bratpiorka, line 127: I think HWLOC should be installed only when disable_hwloc == OFF ?

PatKamin: It should, but it wasn't. It was always installed, so I've left the logic intact in this PR. Anyway, the proposed change is a small one, I'll include it in this PR.

PatKamin: Done

bratpiorka (approved):

PR #1294, opened 2025-04-30
Fix UMF build on Alpine OS

wlemkows:
### Checklist

- [ ] All tests pass locally

### DIFF
--- a/.github/docker/alpine-3.20.Dockerfile
+++ b/.github/docker/alpine-3.20.Dockerfile
@@ -1,0 +1,46 @@
+# Copyright (C) 2025 Intel Corporation
+# Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+#
+# Dockerfile - a 'recipe' for Docker to build an image of Alpine
+#              environment for building the Unified Memory Framework project.
+#
+
+# Pull base Alpine image version 3.20
+FROM alpine:3.20
+
+# Set environment variables
+ENV OS alpine
+ENV OS_VER 3.20
+
+# Base development packages
+ARG BASE_DEPS="\
+	cmake \
+	git \
+    g++ \
+    make"
+
+# UMF's dependencies
+ARG UMF_DEPS="\
+	hwloc-dev"
+
+# Add a new (non-root) 'test_user'
+ENV USER test_user
+ENV USERPASS pass
+RUN useradd -m "${USER}" -g sudo -p "$(mkpasswd ${USERPASS})"
+USER test_user
+
+# Update and install required packages
+RUN apk update \
+ && apk add \
+	${BASE_DEPS} \
+	${UMF_DEPS}
+
+# clone the repository
+RUN git clone https://github.com/oneapi-src/unified-memory-framework.git
+
+# build the project
+RUN cd unified-memory-framework \
+cmake -B build -DCMAKE_BUILD_TYPE=Release -DUMF_BUILD_TESTS=ON -DUMF_BUILD_EXAMPLES=OFF \
+cmake --build build
--- a/.github/docker/alpine-3.20.Dockerfile
+++ b/.github/docker/alpine-3.20.Dockerfile
@@ -18,29 +18,20 @@ ENV OS_VER 3.20
 ARG BASE_DEPS="\
 	cmake \
 	git \
-    g++ \
-    make"
+	g++ \
+	make"
 
 # UMF's dependencies
 ARG UMF_DEPS="\
 	hwloc-dev"
 
-# Add a new (non-root) 'test_user'
-ENV USER test_user
-ENV USERPASS pass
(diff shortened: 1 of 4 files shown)


src/utils/utils_log.c
```diff
@@ -153,13 +152,21 @@ static void utils_log_internal(utils_log_level_t level, int perror,
...
-            char err_buff[1024]; // max size according to manpage.
             int saveno = errno;
             errno = 0;
+#if (_POSIX_C_SOURCE >= 200112L || _XOPEN_SOURCE >= 600) && !_GNU_SOURCE
+            char err[1024];
+            int err_ret = strerror_r(saveno, err, sizeof(err));
+            if (err_ret == ERANGE) {
+                postfix = "[truncated...]";
+            }
+#else
```
lplewa, line 163: is it diffrent from the apple version? can you just update apple ifdef

.github/workflows/nightly.yml
```diff
@@ -370,3 +372,25 @@ jobs:
...
+    - name: Checkout repository
+      uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      with:
+        fetch-depth: 0
+
+    - name: Build Alpine image
+      run: |
+        docker build . -f .github/docker/alpine-3.20.Dockerfile -t alpine_image
+
+    - name: Run UMF build on Alpine image
```
lukaszstolarczuk, line 394: we could also run tests on the build (not sure if as part of `build.sh` or as a separate step here)

.github/docker/alpine-3.20.Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+# Dependencies for tests
+ARG TEST_DEPS="\
+	numactl-dev"
+
+# Update and install required packages
+RUN apk update \
+ && apk add \
+	${BASE_DEPS} \
+	${TEST_DEPS} \
+	${UMF_DEPS}
```
lukaszstolarczuk, line 37: pls add a non-root user (see other dockers as an example)

.github/scripts/alpine_build.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+# Copyright (C) 2025 Intel Corporation
+# Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
+# SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+
+# alpine_build.sh - Script for building UMF on Alpine image
+
+set -e
+
+cd unified-memory-framework
```
lukaszstolarczuk, line 10: I somehow dislike this random `cd`... but I'm not sure how to better handle this... At the moment I can only think of updating `HOST_WORKDIR` and `WORKDIR` to already include it... but not sure if this is better...?

.github/workflows/nightly.yml
```diff
@@ -370,3 +372,25 @@ jobs:
...
+
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      with:
+        fetch-depth: 0
+
+    - name: Build Alpine image
+      run: |
+        docker build . -f .github/docker/alpine-3.20.Dockerfile -t alpine_image
```
lukaszstolarczuk, line 392: tag could be probably something more customized, e.g. `umf-alpine-3.20`

.github/workflows/nightly.yml
wlemkows: Setting up the necessary dependencies on Alpine is not worth the effort (at least for the purpose of fixing this issue)

.github/scripts/alpine_build.sh
wlemkows: One or the other method is equally bizarre, so I did not make changes.

.github/workflows/reusable_trivy.yml
```diff
@@ -19,13 +19,20 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Manual Trivy Setup
+        uses: aquasecurity/setup-trivy@9ea583eb67910444b1f64abf338bd2e105a0a93d
```
lukaszstolarczuk, line 23: pls add a comment which version of the action is it (e.g. similarly to the one below: `# v0.30.0`)

.github/workflows/reusable_trivy.yml
```diff
@@ -19,13 +19,20 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Manual Trivy Setup
```
lukaszstolarczuk, line 22: you could add a simple comment why to set it up manually

.github/workflows/reusable_trivy.yml
wlemkows: It appears that we are now using the correct version by default, and I no longer see the error in the logs; therefore, I decided to remove the manual setup.

src/utils/utils_log.c
wlemkows: Done.

lukaszstolarczuk (approved):

PR #1296, opened 2025-05-05
Check global state destruction in destructors

npmiller:
In some cases the global state may have been destroyed when we reach `umfDestroyPool` or `umfDestroyProvider`, in which case actually going through with the destruction will cause segmentation faults.

The way the global state is managed should maybe be re-worked, but this should be an okay workaround to avoid segmentation faults until then.

### Checklist

- [x] All tests pass locally


ldorau (approved):
LGTM

bratpiorka:
@npmiller thank you very much for this fix. Please don't worry about some of the failing tests in our CI - they are known sporadic issues

PR #1305, opened 2025-05-09
Fix maximum number of arenas in op_initialize()

ldorau:
### Description

Number of arenas greater than 253 causes
the "Resource temporarily unavailable" error
on a machine with 96 cores.

### Checklist

- [ ] All tests pass locally

### DIFF
--- a/src/pool/pool_jemalloc.c
+++ b/src/pool/pool_jemalloc.c
@@ -49,6 +49,11 @@ umfJemallocPoolParamsSetNumArenas(umf_jemalloc_pool_params_handle_t hParams,
 
 #define MALLOCX_ARENA_MAX (MALLCTL_ARENAS_ALL - 1)
 
+// Number of arenas greater than 253 causes
+// the "Resource temporarily unavailable" error
+// on a machine with 96 cores.
+#define MALLOCX_ARENA_MAX_SOFT 250
+
 typedef struct umf_jemalloc_pool_params_t {
     size_t n_arenas;
 } umf_jemalloc_pool_params_t;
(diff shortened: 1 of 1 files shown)


ldorau:
@lplewa please review

src/pool/pool_jemalloc.c
```diff
@@ -461,7 +470,9 @@ static umf_result_t op_initialize(umf_memory_provider_handle_t provider,
         err = je_mallctl("arenas.create", (void *)&arena_index, &unsigned_size,
                          NULL, 0);
         if (err) {
-            LOG_ERR("Could not create arena.");
+            LOG_ERR("Could not create a jemalloc arena (n_arenas = %zu, i = "
```
bratpiorka, line 473: nit: another solution to consider would be to start with utils_get_num_cores() * 4 and then try lower numbers in a loop until we got no error. I'm not sure if we need such a complicated solution, but what if there would be 1000 cores machine in the future?

ldorau: Done

bratpiorka (approved):

src/pool/pool_jemalloc.c
```diff
@@ -461,7 +467,19 @@ static umf_result_t op_initialize(umf_memory_provider_handle_t provider,
         err = je_mallctl("arenas.create", (void *)&arena_index, &unsigned_size,
                          NULL, 0);
         if (err) {
-            LOG_ERR("Could not create arena.");
+            // EAGAIN - means that a memory allocation failure occurred
+            // (2 * utils_get_num_cores()) is the required minimum number of arenas
+            if (err == EAGAIN && (i >= (2 * utils_get_num_cores()))) {
```
lplewa, line 472: If user set arenas number param, we should create all of them or fail.

ldorau: Done

lplewa (approved):

PR #1309, opened 2025-05-09
retrieve numanode object by os index instead of logical index.

lplewa:
physical_id field stores os index, so we should use correct function to get hwloc numanode object.

fixes: #1289

### Checklist

- [ ] All tests pass locally


bratpiorka:
@lplewa please add tests that would fail without this change and pass with it

lukaszstolarczuk (approved):

lukaszstolarczuk:
> @lplewa please add tests that would fail without this change and pass with it

as agreed test will be delivered in a sep. PR

bratpiorka (approved):

Issue #971, opened 2024-12-05
`cmake --install . --prefix <some_path>` failed in some cases


vinser52:
## Environment Information

- UMF version (hash commit or a tag):                         main branch
- OS(es) version(s):                                          Ubuntu 22.04
- cmake version :    3.22.1

## Please provide a reproduction of the bug:

```bash
git clone https://github.com/oneapi-src/unified-memory-framework.git
cd unified-memory-framework
mkdir build && cd build
cmake .. \
    -DUMF_BUILD_EXAMPLES=OFF \
    -DUMF_BUILD_BENCHMARKS=OFF \
    -DUMF_BUILD_BENCHMARKS_MT=OFF \
    -DUMF_FORMAT_CODE_STYLE=OFF \
    -DUMF_DEVELOPER_MODE=ON \
    -DUMF_BUILD_SHARED_LIBRARY=ON \
    -DUMF_BUILD_LIBUMF_POOL_JEMALLOC=ON \
    -DUMF_BUILD_LIBUMF_POOL_DISJOINT=ON \
    -DUMF_BUILD_LEVEL_ZERO_PROVIDER=ON \
    -DUMF_BUILD_CUDA_PROVIDER=ON \
    -DUMF_BUILD_GPU_TESTS=OFF \
    -DUMF_BUILD_GPU_EXAMPLES=OFF \
    -DCMAKE_INSTALL_LOCAL_ONLY=TRUE

make -j$(nproc)

cmake --install . --prefix umf_install
```

## How often bug is revealed:

Always when these two conditions are true:

1.  CMake version < 3.28
2. The `UMF_LEVEL_ZERO_INCLUDE_DIR` is not specified and we fetch the Level Zero loader at the build time.

## Actual behavior:

The `cmake --install . --prefix umf_install` fails with the following error:

```bash
-- Install configuration: "Release"
-- Installing: /workdir/umf_repo/build/umf_install/./include/level_zero/ze_api.h
-- Installing: /workdir/umf_repo/build/umf_install/./include/level_zero/ze_ddi.h
-- Installing: /workdir/umf_repo/build/umf_install/./include/level_zero/zes_api.h
-- Installing: /workdir/umf_repo/build/umf_install/./include/level_zero/zes_ddi.h
-- Installing: /workdir/umf_repo/build/umf_install/./include/level_zero/zet_api.h
-- Installing: /workdir/umf_repo/build/umf_install/./include/level_zero/zet_ddi.h
[...]

PR #1310, opened 2025-05-12
use --install in installation tests on Windows

bratpiorka:
This is a fix for the failing test_installation.py on Windows, introduced by #1285. The failures are related to the `--target install` command, which is not properly recognized on Windows. This pull request reverts the installation command to the previous version, which uses `--install` for Windows.

example of passed Windows Ninja job:
https://github.com/oneapi-src/unified-memory-framework/actions/runs/14972212328/job/42055589833?pr=1310

and Windows NMake job:
https://github.com/oneapi-src/unified-memory-framework/actions/runs/14972212328/job/42055589878?pr=1310

full nightly run:
https://github.com/oneapi-src/unified-memory-framework/actions/runs/15117728112

### DIFF
--- a/.github/workflows/reusable_basic.yml
+++ b/.github/workflows/reusable_basic.yml
@@ -239,6 +239,7 @@ jobs:
     name: Windows
     env:
       VCPKG_PATH: "${{github.workspace}}/build/vcpkg/packages/hwloc_x64-windows;${{github.workspace}}/build/vcpkg/packages/tbb_x64-windows;${{github.workspace}}/build/vcpkg/packages/jemalloc_x64-windows"
+      VCPKG_PATH_BIN: "${{github.workspace}}/build/vcpkg/packages/hwloc_x64-windows/bin;${{github.workspace}}/build/vcpkg/packages/tbb_x64-windows/bin;${{github.workspace}}/build/vcpkg/packages/jemalloc_x64-windows/bin"
     strategy:
       matrix:
         os: ['windows-2019', 'windows-2022']
@@ -247,6 +248,7 @@ jobs:
         shared_library: ['ON', 'OFF']
         level_zero_provider: ['ON']
         cuda_provider: ['ON']
+        cmake_ver: ['default']
         include:
           - os: 'windows-2019'
             # clang build fails on Windows 2022
@@ -256,6 +258,7 @@ jobs:
             level_zero_provider: 'ON'
             cuda_provider: 'ON'
             toolset: "-T ClangCL"
+            cmake_ver: '3.14.0-win64-x64'
           - os: 'windows-2022'
             build_type: Release
             compiler: {c: cl, cxx: cl}
@@ -263,12 +266,14 @@ jobs:
             level_zero_provider: 'ON'
             cuda_provider: 'ON'
             umfd_lib: 'ON'
+            cmake_ver: '3.28.0-windows-x86_64'
           - os: 'windows-2022'
             build_type: Release
             compiler: {c: cl, cxx: cl}
             shared_library: 'ON'
             level_zero_provider: 'OFF'
             cuda_provider: 'OFF'
+            cmake_ver: 'default'
 
     runs-on: ${{matrix.os}}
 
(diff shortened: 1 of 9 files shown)


test/test_installation.py
```diff
@@ -183,6 +183,16 @@ def install_umf(self) -> None:
 
         install_cmd = f"cmake --build {self.build_dir} --config {self.build_type.title()} --target install"
 
+        cmake_cache_file = self.build_dir / "CMakeCache.txt"
+        if cmake_cache_file.exists():
+            with cmake_cache_file.open() as f:
+                for line in f:
+                    if "CMAKE_GENERATOR:INTERNA" in line:
```
ldorau, line 190: "INTERNA" -> "INTERNA**L**"

test/test_installation.py
```diff
@@ -181,7 +181,10 @@ def install_umf(self) -> None:
                     f"Error: Installation directory '{self.install_dir}' is not empty"
                 )
 
-        install_cmd = f"cmake --build {self.build_dir} --config {self.build_type.title()} --target install"
+        if platform.system() == "Windows":
+            install_cmd = f"cmake --install {self.build_dir} --config {self.build_type.title()} --prefix {self.install_dir}"
```
lukaszstolarczuk, line 185: hmm didn't we just circle back to the original issue: https://github.com/oneapi-src/unified-memory-framework/issues/971 ?

Perhaps we should have various CMakes tested also on Windows...?

bratpiorka: no, the #971 issue is related to Ubuntu. I added several tests in previous PRs to reproduce this case in our CI. But I'll add testing of various CMake on Windows

ldorau (approved):

.github/workflows/reusable_basic.yml
```diff
@@ -318,10 +345,23 @@ jobs:
 
     - name: Run tests
       working-directory: ${{env.BUILD_DIR}}
-      run: ctest -C ${{matrix.build_type}} --output-on-failure --test-dir test
+      # For CMake versions < 3.22 we have to add the build directory to the PATH
```
lukaszstolarczuk, line 348: that's actually super crazy, as I can see the same issue would be with examples... so perhaps we should handle this in CMake...? (warn users? fix the paths already in CMake...? enforce newer versions if test/examples are enabled...?)

// no need to do this in this PR - I can see you marked this as TODO in CMake 👍

.github/workflows/reusable_basic.yml
```diff
@@ -286,7 +310,10 @@ jobs:
         vcpkgJsonGlob: '**/vcpkg.json'
 
     - name: Install dependencies
-      run: vcpkg install --triplet x64-windows
+      run: |
+        vcpkg install --triplet x64-windows        
+        $env:Path = "${{env.VCPKG_PATH_BIN}};$env:Path"
```
lukaszstolarczuk, line 315: perhaps it's worth noting why it's needed (a comment here?)

test/test_installation.py
```diff
@@ -181,7 +181,24 @@ def install_umf(self) -> None:
...
+        try:
+            cmake_ver_cmd = "cmake --version"
+            cmake_ver = subprocess.run(
+                cmake_ver_cmd.split(), capture_output=True, text=True
+            )  # nosec B603
+            cmake_ver = cmake_ver.stdout.splitlines()[0].split(" ")[2]
+            cmake_ver = Version(cmake_ver)
+            print(f"CMake version: {cmake_ver}", flush=True)
+        except subprocess.CalledProcessError:
+            sys.exit(f"Error: CMake check version command '{install_cmd}' failed")
```
lukaszstolarczuk, line 194: did you mean `{cmake_ver_cmd}` ?

lukaszstolarczuk (commented):
in general LGTM

.github/workflows/reusable_basic.yml
bratpiorka: I'm not sure about the examples - they are just regular executables, so the user needs to take care of the correct PATH, as usual. This issue is related only with ctest, as it contains some 'magic' that sets up paths and runs the tests. Anyway, if we encounter more problems here, I will create a separate PR

.github/workflows/reusable_basic.yml
bratpiorka: done

lukaszstolarczuk (approved):

.github/workflows/reusable_basic.yml
lukaszstolarczuk: I think we use "ENVIRONMENT_MODIFICATION" in examples' CMake, but well... we don't execute them here, so we should be good for now

PR #1346, opened 2025-06-02
fix __builtin_strncpy truncated compilation error

lplewa:
error: ‘__builtin_strncpy’ output may be truncated copying 4095 bytes from a string of length 4095 [-Werror=stringop-truncation]

### Checklist

- [ ] All tests pass locally


bratpiorka:
@lplewa how did you find this? I'm curious why it wasn't caught by Coverity. could you add a test for it?

lplewa:
> @lplewa how did you find this? I'm curious why it wasn't caught by Coverity. could you add a test for it?

I got an compilation error on my system when compiling with gcc + tsan - imho it's false positive

bratpiorka (approved):

Issue #1333, opened 2025-05-26
Missing tests for umfFixedMemoryProviderParamsSetMemory


lplewa:
This function in public header but do not have a single test case

PR #1350, opened 2025-06-03
Add tests for umfFixedMemoryProviderParamsSetMemory

lukaszstolarczuk:
Fixes: #1333

### Checklist

- [x] All tests pass locally

### DIFF
--- a/test/provider_fixed_memory.cpp
+++ b/test/provider_fixed_memory.cpp
@@ -240,6 +240,11 @@ TEST_P(FixedProviderTest, alloc_page64_align_one_half_pages_WRONG_ALIGNMENT_2) {
                        UMF_RESULT_ERROR_INVALID_ALIGNMENT, 0);
 }
 
+TEST_P(FixedProviderTest, alloc_size_exceeds_buffer) {
+    size_t size = memory_size + page_size;
+    test_alloc_failure(size, 0, UMF_RESULT_ERROR_OUT_OF_HOST_MEMORY, 0);
+}
+
 // Other positive tests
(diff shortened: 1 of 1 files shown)


test/provider_fixed_memory.cpp
```diff
@@ -334,11 +339,55 @@ TEST_F(test, create_with_zero_size) {
...
-TEST_P(FixedProviderTest, alloc_size_exceeds_buffer) {
-    size_t size = memory_size + page_size;
-    test_alloc_failure(size, 0, UMF_RESULT_ERROR_OUT_OF_HOST_MEMORY, 0);
+TEST_F(test, params_several_set_memory) {
+    constexpr size_t memory_size1 = 100, memory_size2 = 1, memory_size3 = 250;
+    char memory_buffer1[memory_size1];
+    char memory_buffer2[memory_size2];
+    char memory_buffer3[memory_size3];
+
+    umf_fixed_memory_provider_params_handle_t params = nullptr;
```
bratpiorka, line 348: this scenario tests almost nothing. Please test following scenario:
* create fixed provider params with ptr/size set to 0
* then create buffer A and call SetMemory with &A
* then create provider, make allocation and check if the allocation is valid and is a part of A,
* then create buffer B and repeat above steps for B with the same params struct

lukaszstolarczuk: can't set size/ptr to 0, but otherwise done

lplewa (approved):

bratpiorka (approved):

PR #1358, opened 2025-06-06
Fix L0 dlopen

lukaszstolarczuk:


ldorau (approved):

pbalcer:
> Just curious, what is the reason?

Unversioned .so files are typically only installed alongside -dev packages. In this case, UMF required that level-zero-dev loader was installed. Normal runtime level-zero package wasn't enough (because it only installed the .so.1 and .so.1.minor files).

PR #1374, opened 2025-06-12
memtarget & memcapacity: tests and fixes

lukaszstolarczuk:
- add more tests (mostly negative cases)
- change `if`s to asserts in internal functions (i.a. to increase the coverage)
- change return status in `umfMemspaceMemtargetRemove`, when trying to remove the last target - this is aligned with other functions, e.g. when trying to create memspace with `size == 0` it throws the same error


lplewa:
> * `mfMemspaceMemtargetRemove`, when trying to remove the last target - this is aligned with other functions, e.g. when trying to create memspace with `size == 0` it throws the same error

we have function
```c 
/// \brief Creates new empty memspace, which can be populated with umfMemspaceMemtargetAdd()
/// \param hMemspace [out] handle to the newly created memspace
///
umf_result_t umfMemspaceNew(umf_memspace_handle_t *hMemspace);|
```

So this sentence is not true.

lukaszstolarczuk:
> > * `mfMemspaceMemtargetRemove`, when trying to remove the last target - this is aligned with other functions, e.g. when trying to create memspace with `size == 0` it throws the same error
> 
> we have function
> 
> ```c
> /// \brief Creates new empty memspace, which can be populated with umfMemspaceMemtargetAdd()
> /// \param hMemspace [out] handle to the newly created memspace
> /// \return UMF_RESULT_SUCCESS on success or appropriate error code on failure.
> ///
> umf_result_t umfMemspaceNew(umf_memspace_handle_t *hMemspace);|
> ```
> 
> So this sentence is not true.

Fair point. Changed.

lplewa (approved):
Nit: Add test where you remove all targets  from the memspace, and then add some.

lukaszstolarczuk:
> Nit: Add test where you remove all targets from the memspace, and then add some.

I'll add more tests in another iteration, thx

PR #1397, opened 2025-06-30
move memspace api to experimental

lplewa:
### Checklist

- [ ] All tests pass locally


docs/config/api.rst
```diff
@@ -136,9 +136,12 @@ Memspace
 
 TODO: Add general information about memspaces.
 
+.. note::
+   The memspace, memtarget and mempolicy APIs are experimental and may change in future releases.
```
lukaszstolarczuk, line 140: this should be repeated in every section - one might scroll down to memtarget and not see this message at all

README.md
```diff
@@ -327,6 +327,8 @@ Packages required for using this pool and executing tests/benchmarks (not requir
 
 ### Memspaces (Linux-only)
 
+*Note: The memspace, memtarget and mempolicy APIs are experimental and may change in future releases.*
```
lukaszstolarczuk, line 330: I guess it'd be better to use syntax like `> **Note**: ...` - the italics you proposed is not very well visible in the readme, imho

include/umf/experimental/mempolicy.h
```diff
@@ -1,6 +1,6 @@
 /*
  *
- * Copyright (C) 2024 Intel Corporation
+ * Copyright (C) 2024-2025 Intel Corporation
  *
  * Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
  * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
```
lukaszstolarczuk, line 6: question: do we want to leave the guards as they are or change e.g. `UMF_MEMPOLICY_H` -> `UMF_EXPERIMENTAL_MEMPOLICY_H`, etc.

docs/config/api.rst
lplewa: Done

PR #1387, opened 2025-06-24
return umf_result in (almost) all umf functions

lplewa:
### Checklist

- [ ] Code compiles without errors locally
- [ ] All tests pass locally
- [ ] CI workflows execute properly

- [ ] CI workflows, not executed per PR (e.g. Nightly), execute properly 
- [ ] New tests added, especially if they will fail without my changes
- [ ] Added/extended example(s) to cover this functionality
- [ ] Extended the README/documentation
- [ ] All newly added source files have a license
- [ ] All newly added source files are referenced in CMake files
- [ ] Logger (with debug/info/... messages) is used
- [ ] All API changes are reflected in docs and def/map files, and are tested

PR #1398, opened 2025-06-30
Unexperiment pool get name

lplewa:
~~depends on #1387~~

### DIFF
--- a/include/umf/base.h
+++ b/include/umf/base.h
@@ -28,7 +28,7 @@ extern "C" {
 #define UMF_MINOR_VERSION(_ver) (_ver & 0x0000ffff)
 
 /// @brief Current version of the UMF headers
-#define UMF_VERSION_CURRENT UMF_MAKE_VERSION(0, 11)
+#define UMF_VERSION_CURRENT UMF_MAKE_VERSION(0, 12)
 
 /// @brief Operation results
 typedef enum umf_result_t {
--- a/include/umf/memory_pool_ops.h
+++ b/include/umf/memory_pool_ops.h
@@ -130,8 +130,21 @@ typedef struct umf_memory_pool_ops_t {
     umf_result_t (*get_last_allocation_error)(void *pool);
 
     ///
-    /// Following functions, with ext prefix, are optional and memory pool implementation
-    /// can keep them NULL.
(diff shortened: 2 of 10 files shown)


include/umf/base.h
```diff
@@ -28,7 +28,7 @@ extern "C" {
 #define UMF_MINOR_VERSION(_ver) (_ver & 0x0000ffff)
 
 /// @brief Current version of the UMF headers
-#define UMF_VERSION_CURRENT UMF_MAKE_VERSION(0, 11)
+#define UMF_VERSION_CURRENT UMF_MAKE_VERSION(0, 12)
```
lukaszstolarczuk, line 31: soo.. now as we established we'll be doing the 1.0.0 version, is this about to be changed...?

include/umf/memory_pool_ops.h
```diff
@@ -163,7 +163,7 @@ typedef struct umf_memory_pool_ops_t {
     ///   otherwise the pool's name is returned.
     /// @return UMF_RESULT_SUCCESS on success or appropriate error code on failure.
     ///
-    umf_result_t (*ext_get_name)(void *pool, const char **name);
+    umf_result_t (*get_name)(void *pool, const char **name);
```
lukaszstolarczuk, line 166: I get that you updated the comment above, but shouldn't this function be now moved above the `The following function is optional...` section?

src/pool/pool_jemalloc.c
```diff
@@ -557,6 +557,12 @@ static umf_result_t op_get_last_allocation_error(void *pool) {
     return TLS_last_allocation_error;
 }
 
+static umf_result_t op_get_name(void *pool, const char **name) {
```
lukaszstolarczuk, line 560: shouldn't we add some tests with that...?

bratpiorka: I have some draft code for improving tests in ops area .... let's do it in a separate PR

include/umf/base.h
lplewa: For now we have 0.12 tag - imho this variable should be updated to 1.0 with a release

include/umf/memory_pool_ops.h
lplewa: done

PR #1402, opened 2025-07-01
fix incompatible function pointer on windows

lplewa:
// Failing nightly CI: https://github.com/oneapi-src/unified-memory-framework/actions/runs/15989654467/job/45100430704

### Checklist

- [ ] All tests pass locally

### DIFF
--- a/src/libumf_windows.c
+++ b/src/libumf_windows.c
@@ -33,6 +33,8 @@ void libumfInit(void) {
 
 INIT_ONCE init_once_flag = INIT_ONCE_STATIC_INIT;
 
+static void umfTearDownWrapper(void) { (void)umfTearDown(); }
+
 BOOL CALLBACK initOnceCb(PINIT_ONCE InitOnce, PVOID Parameter,
                          PVOID *lpContext) {
     (void)InitOnce;  // unused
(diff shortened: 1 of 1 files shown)


src/libumf_windows.c
```diff
@@ -33,14 +33,16 @@ void libumfInit(void) {
 
 INIT_ONCE init_once_flag = INIT_ONCE_STATIC_INIT;
 
+static void umfTearDownWrapper(void) { (void)umfTearDown(); }
```
lukaszstolarczuk, line 36: Pls add info somewhere, perhaps, that it is related to ICX on Win...?

lukaszstolarczuk, line 36: did you test it (e.g. on your fork)? pls paste a link to a passing build, if you have any

KFilipek: It's not necessary, because it's not related strictly with the icx, the icx compiler just trigger warning.

The rules regarding how to determine if two function types are compatible are spelled out in section 6.7.6.3p15 of the [C standard](https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1570.pdf):

```
For two function types to be compatible, both shall specify compatible return types.
Moreover, the parameter type lists, if both are present, shall agree in the number of
parameters and in use of the ellipsis terminator; corresponding parameters shall have 
compatible types.
```

lplewa: I did not. 

Imho we can test it during nightly. Is simple enough patch to assume that it is for sure is working, and if not i can just then make another fix tomorrow.

PR #1404, opened 2025-07-02
Add comments about redundant annotations

ldorau:
### Description

Add comments about redundant annotations.

### Checklist

- [x] All tests pass locally


lplewa:
Is it all missing annotations? We should mark memory as inaccessible after mmap, and only then mark it as undefined/defined, when given to the user.

KFilipek (approved):

PR #1415, opened 2025-07-04
1.0.0-rc1 release

lukaszstolarczuk:
please review if anything is missing - this will be the commit for the tag/release.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,33 @@
+XXX Jul XX 2025 Łukasz Stolarczuk <lukasz.stolarczuk@intel.com>
+
+	* Version 1.0.0-rc1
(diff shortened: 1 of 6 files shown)


ChangeLog
```diff
@@ -1,3 +1,25 @@
+XXX Jul XX 2025 Łukasz Stolarczuk <lukasz.stolarczuk@intel.com>
+
+	* Version 1.0.0-rc1
+
+	This is the first release with the backward compatibility guarantee.
```
lplewa, line 5: In 1.0.0 we should include here  broader explanation what UMF exactly is. 

Also please indicate that some api is experimental, and it not stable.

PR #1419, opened 2025-07-04
Fix: do not use the value field as a next pointer

ldorau:
Do not use the value field as a next pointer,
because it sometimes can lead to race conditions
and reusing already used leaves like here:
https://github.com/ldorau/unified-memory-framework/actions/runs/16074523240/job/45366469415

### Checklist

- [x] All tests pass locally


ldorau:
@lplewa @bratpiorka please review

lplewa (approved):

PR #1441, opened 2025-07-15
[disjoint] Add CTL memory used/reserved metrics

lplewa:
cont. #1430 
### Checklist

- [ ] All tests pass locally


kswiecicki (approved):
nit: You could comment out the name ctl tests that are disabled.
Or will they be filtered out with gtest filter when running CI?

PR #1463, opened 2025-07-24
Add assert in `poolCreateExtUnique()` after `EXPECT_*()`

ldorau:
### Description

Add assert in `poolCreateExtUnique()` after `EXPECT_*()`.

### Checklist

- [x] All tests pass locally

### DIFF
--- a/test/poolFixtures.hpp
+++ b/test/poolFixtures.hpp
@@ -40,6 +40,10 @@ umf_test::pool_unique_handle_t poolCreateExtUnique(poolCreateExtParams params) {
                                   &upstream_provider);
     EXPECT_EQ(ret, UMF_RESULT_SUCCESS);
     EXPECT_NE(upstream_provider, nullptr);
+    if (ret != UMF_RESULT_SUCCESS || upstream_provider == nullptr) {
+        assert(false && "Failed to create a memory provider");
+        return umf_test::pool_unique_handle_t(nullptr, nullptr);
+    }
 
     provider = upstream_provider;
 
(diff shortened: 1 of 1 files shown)


test/poolFixtures.hpp
```diff
@@ -40,6 +40,9 @@ umf_test::pool_unique_handle_t poolCreateExtUnique(poolCreateExtParams params) {
                                   &upstream_provider);
     EXPECT_EQ(ret, UMF_RESULT_SUCCESS);
     EXPECT_NE(upstream_provider, nullptr);
```
lukaszstolarczuk, line 42: you should probably just replace the second EXPECT_ with `ASSERT_NE(upstream_provider, nullptr);`

// unless it's not working becuase it's gtest's setup step and there's something weird going on in there ;)

bratpiorka: we cannot use ASSERT_* in functions that return a type other than void, because this macro returns from func

test/poolFixtures.hpp
```diff
@@ -54,6 +57,9 @@ umf_test::pool_unique_handle_t poolCreateExtUnique(poolCreateExtParams params) {
                         UMF_POOL_CREATE_FLAG_OWN_PROVIDER, &hPool);
     EXPECT_EQ(ret, UMF_RESULT_SUCCESS);
     EXPECT_NE(hPool, nullptr);
+    if (ret != UMF_RESULT_SUCCESS || hPool == nullptr) {
+        assert(false && "Failed to create a memory pool");
```
bratpiorka, line 61: please return here umf_test::pool_unique_handle_t(nullptr, nullptr)

test/poolFixtures.hpp
ldorau: Done

PR #1482, opened 2025-08-06
[CMake] Support preinstalled googlebenchmark

sarnex:
### Description

Currently if `UMF_BUILD_BENCHMARKS` we always fetch and build `benchmark` from source. Use `find_package` to see if the system contains a compatible version preinstalled, and if so, use that.

### Checklist

- [x] All tests pass locally


sarnex:
FYI @pbalcer @igchor

PR #1503, opened 2025-08-21
fix logger when enabled thru ctl

lplewa:
Logger configuration failed when set via the CTL env variable because
the API used const char ** while the env var provided const
char *. This mismatch affected the size field (sizeof(char *)==8
vs. string length).

The API now accepts char * instead of char **.

### DIFF
--- a/src/utils/utils_log.c
+++ b/src/utils/utils_log.c
@@ -66,11 +66,11 @@ typedef struct {
     utils_log_level_t level;
     utils_log_level_t flushLevel;
     FILE *output;
-    const char *file_name;
+    char file_name[MAX_FILE_PATH];
 } utils_log_config_t;
 
 utils_log_config_t loggerConfig = {false,     false, LOG_ERROR,
-                                   LOG_ERROR, NULL,  NULL};
+                                   LOG_ERROR, NULL,  ""};
 
 static const char *level_to_str(utils_log_level_t l) {
     switch (l) {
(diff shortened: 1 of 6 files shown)


src/utils/utils_log.c
```diff
@@ -66,11 +68,11 @@ typedef struct {
     utils_log_level_t level;
     utils_log_level_t flushLevel;
     FILE *output;
-    const char *file_name;
+    char file_name[MAX_FILE_PATH];
```
blazej-smorawski, line 71: Hmm, please remember to check for buffer overflow. It seems like a critical security issue. I used to work as security lead at Biedronka.

bratpiorka (approved):

src/utils/utils_log.c
```diff
@@ -7,6 +7,8 @@
  *
  */
 
+#include "base_alloc_global.h"
```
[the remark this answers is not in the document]
lplewa: done

test/ctl/ctl_env_app.cpp
```diff
@@ -41,6 +41,32 @@ static int test_env_defaults(int argc, char **argv) {
     return 0;
 }
 
+static int test_logger(int argc, char **argv) {
+    char buf[256] = {0};
+    int level = 0;
+
+    if (argc != 2) {
+        std::cerr << "expected two arguments" << std::endl;
+        std::cerr << "Usage: env_defaults log_output log_level" << std::endl;
```
lukaszstolarczuk, line 50: env_defaults -> logger

PR #1525, opened 2025-09-22
use fileline in logs in developer mode instead of funcion name

lslusarczyk:


src/utils/utils_log.h
```diff
@@ -24,17 +24,25 @@ typedef enum {
...
-
-#define LOG_PDEBUG(...) utils_plog(LOG_DEBUG, __func__, __VA_ARGS__);
-#define LOG_PINFO(...) utils_plog(LOG_INFO, __func__, __VA_ARGS__);
-#define LOG_PWARN(...) utils_plog(LOG_WARNING, __func__, __VA_ARGS__);
-#define LOG_PERR(...) utils_plog(LOG_ERROR, __func__, __VA_ARGS__);
-#define LOG_PFATAL(...) utils_plog(LOG_FATAL, __func__, __VA_ARGS__);
+#ifdef UMF_DEVELOPER_MODE
+#define UMF_STRINGIFY(x) #x
+#define UMF_TOSTRING(x) UMF_STRINGIFY(x)
+#define UMF_FUNC_DESC() __FILE__ ":" UMF_TOSTRING(__LINE__)
```
bratpiorka, line 30: could you change this to "file: line (func)" or sth similar that looks ok? I mean I would still like to know the name of the function from which the log was printed

PR #1541, opened 2025-10-29
[CI] Re-enable benchmarks compatibility nightly job

lukaszstolarczuk:


lukaszstolarczuk:
not compatible versions of Compute Runtime, would require more work for this workflow to work. At the moment - low priority.

PR #1576, opened 2026-05-29
fix potential mem corruption in utils_log_internal

bratpiorka:
Fix potential mem corruption in utils_log_internal(), steps to reproduce:

```
tmp = snprintf(b_pos, b_size, "%s: ", func);
b_pos += (int)tmp;
b_size -= (int)tmp;

tmp = vsnprintf(b_pos, b_size, format, args);
b_pos += (int)tmp;
b_size -= (int)tmp;
```

The root cause is incorrect handling of the return values from `snprintf()` and `vsnprintf()`. These functions return the number of bytes that would have been written if enough space were available. The code unconditionally advances `b_pos` and reduces `b_size` by that return value, even when truncation occurred.

If the first `snprintf()` truncates, tmp can be larger than the remaining buffer. This moves `b_pos` past the end of the stack buffer and makes `b_size` invalid. The next `vsnprintf()` then writes through an invalid destination pointer, causing an out-of-bounds stack write.

### DIFF
--- a/src/utils/utils_log.c
+++ b/src/utils/utils_log.c
@@ -1,6 +1,6 @@
 /*
  *
- * Copyright (C) 2024-2025 Intel Corporation
+ * Copyright (C) 2024-2026 Intel Corporation
  *
  * Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
  * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
@@ -72,6 +72,46 @@ typedef struct {
 utils_log_config_t loggerConfig = {false,     false, LOG_ERROR,
                                    LOG_ERROR, NULL,  ""};
 
+static void log_buffer_advance(char **pos, size_t *size, size_t written) {
+    ASSERT(pos);
+    ASSERT(size);
+
+    // do nothing if the size is 0 or can only hold the NULL terminator
+    if (*size <= 1) {
+        *size = 0;
+        return;
+    }
+
+    size_t advance = utils_min(written, *size - 1);
+    *pos += advance;
+
+    if (written >= *size) {
+        *size = 0;
+        return;
+    }
+
+    *size -= advance;
+}
+
+static void log_buffer_append(char **pos, size_t *size, const char *src) {
+    ASSERT(pos);
+    ASSERT(size);
+    ASSERT(src);
+
+    if (*size == 0) {
+        return;
+    }
+
+    size_t src_len = strlen(src);
+    size_t copy_len = utils_min(src_len, *size - 1);
+    memcpy(*pos, src, copy_len);
+
+    // copy_len excludes the trailing NULL, add it explicitly
+    (*pos)[copy_len] = '\0';
+
+    log_buffer_advance(pos, size, src_len);
+}
+
 static const char *level_to_str(utils_log_level_t l) {
     switch (l) {
     case LOG_DEBUG:
@@ -113,7 +153,7 @@ static void utils_log_internal(utils_log_level_t level, int perror,
 
     char buffer[LOG_MAX];
     char *b_pos = buffer;
-    int b_size = sizeof(buffer);
+    size_t b_size = sizeof(buffer);
 
     int tmp = 0;
     if (fileline == NULL) {
@@ -121,16 +161,21 @@ static void utils_log_internal(utils_log_level_t level, int perror,
     } else {
         tmp = snprintf(b_pos, b_size, "%s %s: ", fileline, func);
     }
-    ASSERT(tmp > 0);
 
-    b_pos += (int)tmp;
-    b_size -= (int)tmp;
+    if (tmp < 0) {
+        return; // snprintf error
+    }
+
+    size_t written = (size_t)tmp;
+    log_buffer_advance(&b_pos, &b_size, written);
 
     tmp = vsnprintf(b_pos, b_size, format, args);
-    ASSERT(tmp > 0);
+    if (tmp < 0) {
+        return; // vsnprintf error
+    }
 
-    b_pos += (int)tmp;
-    b_size -= (int)tmp;
+    written = (size_t)tmp;
+    log_buffer_advance(&b_pos, &b_size, written);
 
     const char *postfix = "";
 
@@ -139,10 +184,12 @@ static void utils_log_internal(utils_log_level_t level, int perror,
             strncat(b_pos, ": ", b_size);
             b_pos += 2;
             b_size -= 2;
+            const char *err = "";
 #if defined(_WIN32)
-            char err[80]; // max size according to msdn
-            if (strerror_s(err, sizeof(err), errno)) {
-                *err = '\0';
+            char err_buff[80]; // max size according to msdn
+            err = err_buff;
+            if (strerror_s(err_buff, sizeof(err_buff), errno)) {
+                *err_buff = '\0';
                 postfix = "[strerror_s failed]";
             }
 #else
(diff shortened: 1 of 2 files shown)


src/utils/utils_log.c
```diff
@@ -123,14 +171,12 @@ static void utils_log_internal(utils_log_level_t level, int perror,
     }
     ASSERT(tmp > 0);
 
-    b_pos += (int)tmp;
-    b_size -= (int)tmp;
+    log_buffer_advance(&b_pos, &b_size, tmp);
```
PatKamin: Define `b_size` as `size_t` to avoid further casting of `size_t` to `int` in `log_buffer_append()`. I think `sizeof` should already return `size_t` type number.

bratpiorka: done

PatKamin (approved):

src/utils/utils_log.c
```diff
@@ -72,6 +72,54 @@ typedef struct {
...
+    *size -= written;
+}
+
+static void log_buffer_append(char **pos, size_t *size, const char *src) {
+    ASSERT(pos);
+    ASSERT(size);
+    ASSERT(src);
+
+    if (*size <= 0) {
+        return;
```
KFilipek: `size_t` cannot be lower than `0`:
```suggestion
    if (*size == 0) {
        return;
```

KFilipek (approved):

src/utils/utils_log.c
```diff
@@ -72,6 +72,54 @@ typedef struct {
...
+    }
+
+    if (written == 0) {
+        *size = 0;
+        return;
+    }
+
+    // advance the buffer position and decrease the remaining size
+    // take into the actual number of characters written
```
lplewa: Totally useless AI comment. 

(btw this place need a comment, as this size - 1 is not clear)

src/utils/utils_log.c
```diff
@@ -72,6 +72,54 @@ typedef struct {
 utils_log_config_t loggerConfig = {false,     false, LOG_ERROR,
                                    LOG_ERROR, NULL,  ""};
 
+static void log_buffer_advance(char **pos, size_t *size, size_t written) {
+    ASSERT(pos);
+    ASSERT(size);
+
+    if (*size <= 0) {
```
lplewa: size is unsigned.

src/utils/utils_log.c
```diff
@@ -72,6 +72,54 @@ typedef struct {
...
+        return;
+    }
+
+    if (written == 0) {
+        *size = 0;
+        return;
+    }
```
lplewa: This case is strange? Why if we written 0 we assume that size is zero?

src/utils/utils_log.c
```diff
@@ -72,6 +72,54 @@ typedef struct {
...
+    *pos += written;
+    *size -= written;
+}
+
+static void log_buffer_append(char **pos, size_t *size, const char *src) {
+    ASSERT(pos);
+    ASSERT(size);
+    ASSERT(src);
+
+    if (*size <= 0) {
```
lplewa: dito

src/utils/utils_log.c
```diff
@@ -72,6 +72,54 @@ typedef struct {
...
+    size_t src_len = strlen(src);
+    size_t copy_len = src_len;
+
+    if (copy_len >= *size) {
+        copy_len = *size - 1;
+    }
+    memcpy(*pos, src, copy_len);
+
+    // ensure null-termination
+    (*pos)[copy_len] = '\0';
```
lplewa, line 110: This seems to work by accident - src_len do not add '\0' to len, so we are always one char short - then comment ensure null-termintation is not "ensuring" is just adding null termination as we are always lacking it. 

also you can simplify it with `min` function

bratpiorka:
@lplewa  ?

src/utils/utils_log.c
```diff
@@ -122,15 +170,21 @@ static void utils_log_internal(utils_log_level_t level, int perror,
         tmp = snprintf(b_pos, b_size, "%s %s: ", fileline, func);
     }
     ASSERT(tmp > 0);
+    if (tmp < 0) {
+        return; // snprintf error
+    }
```
lplewa: ASSERT or if? This looks bad.

src/utils/utils_log.c
```diff
@@ -122,15 +170,21 @@ static void utils_log_internal(utils_log_level_t level, int perror,
...
     tmp = vsnprintf(b_pos, b_size, format, args);
     ASSERT(tmp > 0);
+    if (tmp < 0) {
+        return; // vsnprintf error
+    }
 
-    b_pos += (int)tmp;
-    b_size -= (int)tmp;
+    written = (size_t)tmp;
+    log_buffer_advance(&b_pos, &b_size, written);
```
lplewa, line 178: NIT: we should have snprintf wrapper that correctly handles errors. Instead adding this code everywhere

src/utils/utils_log.c
```diff
@@ -72,6 +72,54 @@ typedef struct {
...
+    // take into the actual number of characters written
+    if (written >= *size) {
+        *pos += *size - 1;
+        *size = 0;
+        return;
+    }
+
+    *pos += written;
+    *size -= written;
+}
```
lplewa, line 94:
```suggestion
     // Treat a buffer with only room for the trailing NUL as full: nothing
      // more can be appended, and *size - 1 below would underflow at *size == 0.

     if (*size <= 1) {
          *size = 0;
          return;
      }

      size_t advance = min(written, *size - 1);
      *pos += advance;
      *size -= advance;

}
```

lplewa (changes requested):

src/utils/utils_log.c
bratpiorka: done

src/utils/utils_log.c
bratpiorka: And return from function? I don't want to use a macro here

src/utils/utils_log.c
bratpiorka: done / changed

src/utils/utils_log.c
bratpiorka: changed

src/utils/utils_log.c
bratpiorka: assert removed

bratpiorka: btw, I like when GDB breaks at the point of failure

src/utils/utils_log.c
bratpiorka: done

bratpiorka:
@lplewa please re-review

src/utils/utils_log.c
```diff
@@ -202,18 +243,19 @@ static void utils_log_internal(utils_log_level_t level, int perror,
         localtime_r(&now, &tm_info);
 #endif
 
-        ASSERT(h_size > 0);
-        tmp = (int)strftime(h_pos, h_size, "%Y-%m-%dT%H:%M:%S ", &tm_info);
-        h_pos += tmp;
-        h_size -= tmp;
+        written = strftime(h_pos, h_size, "%Y-%m-%dT%H:%M:%S ", &tm_info);
```
lplewa, line 246: 🟡 Bug 3: strftime returns 0 on overflow

  written = strftime(h_pos, h_size, "%Y-%m-%dT%H:%M:%S ", &tm_info);
  log_buffer_advance(&h_pos, &h_size, written);

  Per C11: when the result doesn't fit, strftime returns 0 and the array contents are indeterminate. With written = 0, log_buffer_advance
  doesn't advance, so the next writer starts at h_pos — which may now contain garbage rather than the '\0' you'd expect. The pre-loop
  memset(header, 0, sizeof(header)) makes this probably fine in practice (the indeterminate bytes started zero), but it's relying on libc
  not scribbling on overflow, which the standard doesn't guarantee. Worth either bumping LOG_HEADER so the timestamp can't overflow, or
  explicitly resetting h_pos[0] = '\0' when written == 0.

lplewa (approved):

PR #1585, opened 2026-07-08
Build hwloc via ExternalProject_Add instead of raw add_custom_command

wenju-he:
The static-hwloc autotools build (FetchContent + hand-written add_custom_command chain for autogen.sh/configure/make/make install) never forwarded CC/CXX to hwloc's ./configure, so it silently picked up the ambient host compiler instead of the one CMake was configured to use (e.g. icpx under a sysroot- or toolchain-file-pinned build), risking an ABI mismatch against the rest of the build.

Switch to ExternalProject_Add, which fetches/patches/configures/builds/ installs hwloc as a single well-defined external project step and lets us explicitly pass CC/CXX through to configure. The Coverity security patch, previously applied via a post-FetchContent execute_process(), now runs through ExternalProject's PATCH_COMMAND for the non-Windows path.

### Checklist

- [x ] Code compiles without errors locally on Linux
- [x ] `ninja test` pass locally on Linux

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -325,11 +325,11 @@ if(UMF_LINK_HWLOC_STATICALLY)
         set(hwloc_targ_SOURCE_DIR ${hwloc_targ_PREFIX}/src/hwloc_targ)
         set(hwloc_targ_BINARY_DIR ${hwloc_targ_PREFIX})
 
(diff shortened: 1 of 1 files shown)


bratpiorka (approved):

lukaszstolarczuk (approved):
LGTM

PR #1587, opened 2026-07-09
add GetCacheLineSize() op to memory provider and use it to initialize DP min bucket size

bratpiorka:
add GetCacheLineSize() op to memory provider and use it to initialize DP min bucket size

### DIFF
--- a/src/libumf.def
+++ b/src/libumf.def
@@ -1,12 +1,12 @@
 ;;;; Begin Copyright Notice
-; Copyright (C) 2023-2025 Intel Corporation
+; Copyright (C) 2023-2026 Intel Corporation
 ; Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
 ; SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
 ;;;;  End Copyright Notice
 
 LIBRARY UMF
 
-VERSION 1.1
+VERSION 1.2
 
 EXPORTS
     DllMain
@@ -158,3 +158,5 @@ EXPORTS
     umfOsMemoryProviderParamsSetName
     umfPoolTrimMemory
     umfScalablePoolParamsSetName
+; Added in UMF_1.2
+    umfMemoryProviderGetCacheLineSize
--- a/src/pool/pool_disjoint.c
+++ b/src/pool/pool_disjoint.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2022-2025 Intel Corporation
+ * Copyright (C) 2022-2026 Intel Corporation
  *
  * Under the Apache License v2.0 with LLVM Exceptions. See LICENSE.TXT.
  * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
@@ -8,6 +8,7 @@
 #include <assert.h>
 #include <ctype.h>
 #include <inttypes.h>
+#include <limits.h>
 #include <stdlib.h>
 #include <string.h>
 
@@ -595,6 +596,7 @@ static __TLS umf_result_t TLS_last_allocation_error;
 // Allocations with size > CutOff bypass the pool and
 // go directly to the provider.
 static const size_t CutOff = (size_t)1 << 31; // 2GB
+#define DP_MAX_BUCKET_COUNT (sizeof(size_t) * CHAR_BIT * 2 + 1)
 
 static size_t bucket_slab_min_size(bucket_t *bucket) {
     return bucket->pool->params.slab_min_size;
@@ -981,24 +983,22 @@ static bool bucket_can_pool(bucket_t *bucket) {
 static size_t size_to_idx(disjoint_pool_t *pool, size_t size) {
     assert(size <= CutOff && "Unexpected size");
     assert(size > 0 && "Unexpected size");
+    assert(pool->buckets_num > 0 && "Unexpected buckets count");
 
-    size_t min_bucket_size = (size_t)1 << pool->min_bucket_size_exp;
-    if (size < min_bucket_size) {
-        return 0;
+    size_t left = 0;
+    size_t right = pool->buckets_num - 1;
+
+    while (left < right) {
+        size_t mid = left + (right - left) / 2;
+
+        if (pool->buckets[mid]->size < size) {
+            left = mid + 1;
+        } else {
+            right = mid;
+        }
     }
 
-    // get the position of the leftmost set bit
-    size_t position = utils_msb64(size);
-
-    bool is_power_of_2 = IS_POWER_OF_2(size);
-    bool larger_than_halfway_between_powers_of_2 =
-        !is_power_of_2 &&
-        (bool)((size - 1) & ((uint64_t)(1) << (position - 1)));
-    size_t index = (position - pool->min_bucket_size_exp) * 2 +
-                   (int)(!is_power_of_2) +
-                   (int)larger_than_halfway_between_powers_of_2;
-
-    return index;
+    return left;
 }
 
 static umf_disjoint_pool_shared_limits_t *
@@ -1120,6 +1120,7 @@ static const umf_disjoint_pool_params_t default_params = {
     .max_poolable_size = 2 * 1024 * 1024, // 2MB default
     .capacity = 4,                        // default
     .min_bucket_size = 8,                 // default
+    .min_bucket_size_set = false,
     .cur_pool_size = 0,
     .pool_trace = 0,
     .shared_limits = NULL,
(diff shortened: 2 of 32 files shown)


src/libumf.def
lukaszstolarczuk: bump?

test/common/ze_loopback.def
```diff
@@ -36,3 +37,4 @@ EXPORTS
         zeMemOpenIpcHandle
         zeMemCloseIpcHandle
         zeDeviceGetMemoryProperties
+        zeDeviceGetCacheProperties
```
lukaszstolarczuk, line 39: fyi, not sorted a-z

test/common/ze_loopback.map
```diff
@@ -35,6 +36,7 @@
         zeMemOpenIpcHandle;
         zeMemCloseIpcHandle;
         zeDeviceGetMemoryProperties;
+        zeDeviceGetCacheProperties;
```
lukaszstolarczuk, line 38: .

src/libumf.def
bratpiorka: done

test/common/ze_loopback.def
bratpiorka: done

test/common/ze_loopback.map
bratpiorka: done

test/pools/disjoint_pool_ctl.cpp
```diff
@@ -614,10 +614,17 @@ TEST_F(test, disjointCtlBucketStats) {
...
-        // Odd  indexes: 12 << (i/2) => 12,24,48,96,...
-        return (i % 2 == 0) ? (size_t(8) << (i / 2)) : (size_t(12) << (i / 2));
+    size_t min_bucket_size = 0;
+    ASSERT_SUCCESS(umfCtlGet("umf.pool.by_handle.{}.params.min_bucket_size",
+                             &min_bucket_size, sizeof(min_bucket_size),
+                             poolWrapper.get()));
+
+    auto expected_bucket_size = [min_bucket_size](size_t i) -> size_t {
+        // Even indexes: min,2*min,4*min,...
+        // Odd  indexes: 1.5*min,3*min,6*min,...
```
spencerpatty: ok, I have concerns about the 1.5 * min case since I don't see a requirement for cache_line_size alignment on the front of the array, wouldn't 1.5 * min  result in the end of a buffer being halfway through a cache line ? and then the subsequent allocation would start there (causing two allocations to share a cache line? )

I suggest one of two options:

1. actually enforce starting on min_bucket_size alignment (which is now consistent with cache_line_size)
2.  delete 1.5 * min,  so you have Even indices:   min, 2 * min, 4 * min ...  and Odd Indices:  (1.5 * min ->2 * min or deleted), 3 * min, 6 * min ...

test/pools/disjoint_pool_ctl.cpp
```diff
@@ -614,10 +614,17 @@ TEST_F(test, disjointCtlBucketStats) {
...
+        return (i % 2 == 0)
+                   ? (min_bucket_size << (i / 2))
+                   : ((min_bucket_size + min_bucket_size / 2) << (i / 2));
     };
```
spencerpatty: The implementation of 2 could be enforced using something like 

```
// stays the same or rounds up to the next multiple of pow_2
// where pow_2 should be a power of two
template<typename T, typename std::enable_if<std::is_integral<T>::value>::type* = nullptr>
T round_up_next_multiple(T count, T pow_2_mult)
{
    assert(pow_2_mult != 0 && ((pow_2_mult & pow_2_mult - 1) == 0)); // check pow_2_mult is a power of two
    return (count + pow_2_mult - 1) & ~(pow_2_mult - 1);
}

// enforce length is a multiple of min_bucket_size, which in round about way enables alignment for subsequent allocations to be on min_bucket_size (aka cache_line_size)
expected_bucket_size = round_up_next_multiple<size_t>(expected_bucket_size, min_bucket_size);
```

spencerpatty: or something like that with a lambda ...   or as the bad pattern is only for the case of "i==3", you could try to modify your ternary into something more complicated.  I kind of like the use of this "helper" function since it signals the intent very clearly, even if it is five extra OPs ...

test/providers/provider_level_zero.cpp
```diff
@@ -403,6 +405,34 @@ TEST_P(umfLevelZeroProviderTest, getPageSize) {
...
+    umf_memory_provider_handle_t provider = nullptr;
+    umf_result_t umf_result = umfMemoryProviderCreate(
+        umfLevelZeroMemoryProviderOps(), params, &provider);
+    ASSERT_EQ(umf_result, UMF_RESULT_SUCCESS);
+    ASSERT_NE(provider, nullptr);
+
+    size_t cacheLineSize = 0;
+    umf_result = umfMemoryProviderGetCacheLineSize(provider, &cacheLineSize);
+    EXPECT_EQ(umf_result, UMF_RESULT_SUCCESS);
+    EXPECT_GT(cacheLineSize, 0);
```
PatKamin, line 418: Shouldn't it be equal to the default of `128` for a level zero memory provider?

bratpiorka: On PVC, the cache line size is 64 bytes, while on BMG it is 256 bytes.
In any case, I don't think we should assume a specific value in the test. The test shouldn't need to know any implementation details to pass.

test/pools/disjoint_pool_ctl.cpp
bratpiorka: Ok, changed. I modified calculating the size of buckets and now there is no simple way to calculate bucket size from index, so I do not expect specyfic sizes in the test but rather check more generic assumptions (ascending order of sizes, proper alignment).

src/pool/pool_disjoint.c
```diff
@@ -1191,53 +1195,67 @@ umf_result_t disjoint_pool_post_initialize(void *ppPool) {
...
+    }
+    bucket_sizes[i] = CutOff;
+
+    // After rounding up some sizes may be equal, we need to remove duplicates
+    // and count the number of unique sizes.
+    size_t unique_count = 0;
+    for (size_t j = 0; j <= i; j++) {
+        if (j == 0 || bucket_sizes[j] != bucket_sizes[j - 1]) {
+            unique_bucket_sizes[unique_count++] = bucket_sizes[j];
+        }
```
spencerpatty, line 1242: this seems quite reasonable as an init cost ... and seems to relieve the problem.  I still wish you guaranteed a starting point on cache_line_size boundary, but this has the nice side effect of guaranteeing it if you start on such a boundary, so it all works out in the end :)  this seems like it should resolve our problems of multiple allocations ending up on the same cache_line, so I look forward to seeing it in the lib :)

spencerpatty (approved):
Approved from MKL perspective. The changes to the memory pool initialization here should resolve the issue of potentially having more than one allocation sharing a cache line. 

Question: Are there tests that will validate alignment for minBucketSize, minBucketSize + minBucketSize/2, 2*minBucketSize, 2*minBucketSize+minBucketSize/2 ?  This range seems to be the range where things could go wrong ...

src/pool/pool_disjoint.c
```diff
@@ -1191,53 +1195,67 @@ umf_result_t disjoint_pool_post_initialize(void *ppPool) {
...
+    size1 = utils_max(size1, UMF_DISJOINT_POOL_MIN_BUCKET_DEFAULT_SIZE);
 
-    // Buckets sized smaller than the bucket default size- 8 aren't needed.
-    Size1 = utils_max(Size1, UMF_DISJOINT_POOL_MIN_BUCKET_DEFAULT_SIZE);
+    // Calculate the size of the "halfway" bucket
+    size_t size2 = size1 + size1 / 2;
 
     // Calculate the exponent for min_bucket_size used for finding buckets.
-    disjoint_pool->min_bucket_size_exp = (size_t)utils_msb64(Size1);
+    disjoint_pool->min_bucket_size_exp = (size_t)utils_msb64(size1);
```
lukaszstolarczuk: hmm... do you use `min_bucket_size_exp` anywhere now?

src/pool/pool_disjoint.c
```diff
@@ -1191,53 +1195,67 @@ umf_result_t disjoint_pool_post_initialize(void *ppPool) {
...
-        disjoint_pool->buckets_num += 2;
-        Size2 *= 2;
+    // Calculate sizes of buckets and store them in the array.
+    size_t i = 0;
+    for (; size2 < CutOff; size1 *= 2, size2 *= 2, i += 2) {
+        bucket_sizes[i] =
+            ALIGN_UP_SAFE(size1, disjoint_pool->params.min_bucket_size);
+        bucket_sizes[i + 1] =
+            ALIGN_UP_SAFE(size2, disjoint_pool->params.min_bucket_size);
+    }
```
lukaszstolarczuk, line 1233: can't we go above `DP_MAX_BUCKET_COUNT`...?

src/pool/pool_disjoint.c
bratpiorka: removed

src/pool/pool_disjoint.c
bratpiorka: no, this is already guarded by the way CutOff is calculated + by the condition in the loop:: `for (; size2 < CutOff; ...`

lukaszstolarczuk (approved):
LGTM 👍

PatKamin (approved):

